Issue #21, opened 2019-07-27
consistent ui design / layout / spacing spec


tomholub:
There are lot of broken layouts around the app, especially when viewed on different size devices.

No thought was given to responsive design so far, and this needs to be fixed.

Here's an example of a broken layout, but there are many, and many of them are subtle.

As a part of this issue:
 - [x] fix any obvious brokenness like the one on the screenshots below
 - [x] fix any more subtle broken ui, and document what was fixed
 - [x] establish common padding and margins around text and buttons, and consistently apply it throughout the app
 - [x] document here your choices for common element padding sizes, distance between buttons, line item sizes, text sizes, etc
 - [x] test and fix design for devices starting with iPod touch all the way up to iPhone X
 - [x] test the app on at least one iPad and fix any obviously broken parts
 - [x] test the app in landscape iPhone view and fix where necessary
 - [x] provide screenshots of what was fixed and how the result looks
 - [x] fix or propose ui / design improvements
 - [x] fix xcode layout warnings
 - [x] when all done, please provide a screenshots of what the app looks on several screens on small iPhone, large iPhone and any tablet.

tomholub:
Eg here the gray box was supposed to be just a thin line (bottom border under the pass phrase)

![image](https://user-images.githubusercontent.com/6306961/61993520-6c9c9300-b09f-11e9-92f0-1448d91a5402.png)

Or here it's way too broken:

![image](https://user-images.githubusercontent.com/6306961/61993524-850cad80-b09f-11e9-99fc-134659fce50b.png)

Or here the title text does not even fit on screen

![image](https://user-images.githubusercontent.com/6306961/61993528-9950aa80-b09f-11e9-87ad-740e24f17fac.png)

etc etc

tomholub:
Unexpected gray block in compose, was probably supposed to be a line just like between recipient and subject.

![image](https://user-images.githubusercontent.com/6306961/62066318-7b5e8380-b220-11e9-8859-924d64b4138a.png)

RomanKovalchuk2019:
Fixed in pr #32

tomholub:
@RomanKovalchuk2019 did you also look at these three?

* [ ]  document here your choices for common element padding sizes, distance between buttons, line item sizes, text sizes, etc
* [ ]  test the app on at least one iPad and fix any obviously broken parts
* [ ]  fix xcode layout warnings

Let me know.

RomanKovalchuk2019:
[...]

Issue #13, opened 2019-07-27
top bar improvements + icon spacing


tomholub:
fix icon spacing in compose - should lean towards the right side, with equal spacing between them, leaving enough space for the title (eg `Compose`)

Hide `trash` icon when the message is already in Trash, relevant code (to fix):
```Swift
        if self.path == "[Gmail]/Trash" {
            // self.btnTrash.isHidden = true; // todo - hide the trash btn
        }
```
and to remove:
```Swift
        guard self.path != "[Gmail]/Trash" else { // todo - does not need to be here once trash btn is hidden
            self.showToast("Message is already in trash")
            return
        }
```
Ensure spacing of other icons does not break without the trash icon.

tomholub:
Here broken spacing
![image](https://user-images.githubusercontent.com/6306961/61993327-c780bb00-b09c-11e9-80d6-245c75a67097.png)

Here also broken, should be closer together:
![image](https://user-images.githubusercontent.com/6306961/61993330-d7000400-b09c-11e9-8704-2e9c0cd68ad1.png)

RomanKovalchuk2019:
Fixed in pr #32

Issue #12, opened 2019-07-27
compose: body text improvements


tomholub:
- [x]  Remove msg placeholder when focused
- [x] make typed text show in black
- [x]  use standard size for composed text: 17
- [x] make compose field scrollable (currently cannot see text if I write too many lines)

tomholub:
here example of too much text / cannot scroll:

![image](https://user-images.githubusercontent.com/6306961/61993338-f9921d00-b09c-11e9-86ae-3d7ab110ca07.png)

RomanKovalchuk2019:
Fixed in pr #32

Issue #7, opened 2019-07-25
fix decrypted message scrolling


tomholub:

tomholub:
doesn't actually scroll fully yet

tomholub:
The issue is visible with messages that span more than one screen. It looks like it would scroll, but it really doesn't.

RomanKovalchuk2019:
Fixed in pr #32

PR #32, opened 2019-08-17
Remake compose text view

RomanKovalchuk2019:
Fixed navigation bar layout, fixed message scrolling, added compose message scrolling. 
I also remove title in compose page, because I think navigation bar would look better without title. Please, take a look on that, and if you don't agree with me, I will change it.


tomholub:
#21 #13 #12 #7

FlowCrypt/Controllers/Inbox/InboxViewController.swift
```diff
@@ -20,20 +25,44 @@ class InboxViewController: BaseViewController, UITableViewDelegate, UITableViewD
...
+    }
+
+    override func viewWillDisappear(_ animated: Bool) {
+        self.sideMenuController()?.sideMenu?.allowslideGesture = false
+        self.tableView.reloadData()
+    }
+
+    func movedOrUpdated(objMessage: MCOIMAPMessage) {
+        let index = self.messages.firstIndex(of: objMessage)
+        self.messages.remove(at: index!)
```
tomholub: Getting NPE here after archiving email

Screenshot 2019-08-18 at 2 07 35 AM

Issue #33, opened 2019-08-19
delete/trash stopped working properly after PR


tomholub:
@RomanKovalchuk2019 

After #32 landed, the "trash" icon will no longer move mail to trash folder. Maybe it does something? But I'm not sure what.

When I click it on a particular message, I expect it will say "Moved to trash" and show inbox, without that message. Instead, it stays in message view. If I click trash again, nothing happens again. I can go back to inbox, refresh it and still see the message. Then I can open it again, click trash again, and nothing happens.

If I kill the app and retry, the trash icon on that message still does nothing.

This used to work.

PR #42, opened 2019-08-20
Changed target of trash button

RomanKovalchuk2019:
issue #33


FlowCrypt.xcodeproj/project.pbxproj
```diff
@@ -831,7 +831,7 @@
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				COPY_PHASE_STRIP = NO;
 				DEBUG_INFORMATION_FORMAT = dwarf;
-				ENABLE_BITCODE = YES;
+				ENABLE_BITCODE = NO;
```
tomholub: What is the reason to disable bitcode in development?

FlowCrypt.xcodeproj/project.pbxproj
```diff
@@ -892,7 +892,7 @@
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				COPY_PHASE_STRIP = NO;
 				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
-				ENABLE_BITCODE = YES;
+				ENABLE_BITCODE = NO;
```
tomholub: Releases should definitely have bitcode enabled, unless you have reason to disable them - in which case please let me know.

FlowCrypt.xcodeproj/project.pbxproj
RomanKovalchuk2019: I changed it, because I can not run tests. I have an error that says bitcode and bundle could not be used together.

tomholub: Interesting. I will revert this change for now (bitcode is a good thing to have for debugging), and later we can take a look at what causes the issues with tests, in another issue.

Issue #39, opened 2019-08-20
do not use force unwraping unnecessarily


tomholub:
> do not use force unwraping(let userInfo: NSDictionary = notification.userInfo! as NSDictionary) and so on

PR #44, opened 2019-08-20
Replace force unwraping

Kharchevskyi:
Resolve Issues

- https://github.com/FlowCrypt/flowcrypt-ios/issues/39
- https://github.com/FlowCrypt/flowcrypt-ios/issues/36

_(2 hours)_


FlowCrypt/Extensions/URLSessionExtension.swift
```diff
@@ -18,19 +18,27 @@ struct HttpErr: Error {
...
+                let status = res?.statusCode ?? Constants.Global.generalError
+
+                print("URLSession.call status:\(status) ms:\(start.millisecondsSince()) \(urlRequest.url?.absoluteString ?? "??")")
+
+                let validStatusCode = 200...299
+                let isCodeVaild = validStatusCode ~= status || tolerateStatus?.contains(status)
+                let isValidResonse = error == nil && isCodeVaild
+
+                if let data = data, isValidResonse {
+                    resolve(HttpRes(status: status, data: data!))
```
tomholub: My xcode is showing many different errors across the codebase now, for example:
image

Should I update some settings, etc?

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -148,7 +148,7 @@ class ComposeViewController: BaseViewController {
             : txtSubject.text ?? "(no subject)"
 
         let from = GoogleApi.instance.getEmail()
-        let replyToMimeMsg = replyToMime.map { String(data: $0, encoding: .utf8) }
+        let replyToMimeMsg = replyToMime.flatMap { String(data: $0, encoding: .utf8) }
```
tomholub: Curious why the flatMap here?

Kharchevskyi: because `String(data: $0, encoding: .utf8)` will return you optional `Data?` in case it would be replyToMime.map {   }, your replyToMimeMsg will be: `String??` 
flatMap will flattened failable init for you

tomholub: Oh so flatMap gets rid of optionals?

PR #49, opened 2019-08-21
refactored order of class params

RomanKovalchuk2019:
All folder names is capitalized now


tomholub:
I would actually prefer to not capitalize them, for consistency with our Android app

RomanKovalchuk2019:
Ok. I will remove capitalization then, and please close issue 29 after that

tomholub:
It looks like master meanwhile already received exactly the same change (so after merging changes from master, there is no change). So I will go ahead and close this.

Issue #28, opened 2019-07-29
inbox folder not selected in menu right after start


tomholub:
Correctly it should be showing like this, and it does when I click on folder to select it manually in the menu:

![image](https://user-images.githubusercontent.com/6306961/62050403-b0f37480-b200-11e9-86a0-ef5fd7484ae3.png)

But when I open the menu right after app start, the folder is not highlighted:

![image](https://user-images.githubusercontent.com/6306961/62050468-cc5e7f80-b200-11e9-9d7f-cde37d8405d5.png)

PR #50, opened 2019-08-21
Added selection for first item in menu after start

RomanKovalchuk2019:
I removed bounce on scroll, because I think this looks better. Please check it, if you do not agree with me let me now, I will change it


tomholub:
Issue #28

tomholub:
Thanks for disabling the bounce, that's much better now.

Issue #40, opened 2019-08-20
retain cycle - memory leak?


tomholub:
> there are few places where you have retain cycle.
> Your viewcontrollers would never dealocated

Issue #38, opened 2019-08-20
refactor view controller setup to follow best practices


tomholub:
- [x] fix menu closing bug

Issue #37, opened 2019-08-20
use private func and private var


tomholub:
where appropriate

PR #57, opened 2019-08-21
Project refactoring

Kharchevskyi:
- refactored MyMenuTableViewController
- refactored Message view controller 
   - replaced memory leaks
   - added error handling
   - added db service
   - refactored ui setup 
- refactored InboxViewController
   - refactored ui setup 
   - replaced memory leaks 
-refactored SignInViewController
   - replaced memory leak
- use `final` in classes
- use private/internal access levels
- added Then extension 
- minor updates
- added extensions for step by step replacing BaseViewController

https://github.com/FlowCrypt/flowcrypt-ios/issues/40
https://github.com/FlowCrypt/flowcrypt-ios/issues/38
https://github.com/FlowCrypt/flowcrypt-ios/issues/37

_9 hours_


tomholub:
#38

FlowCrypt/Controllers/SignIn/SignInViewController.swift
```diff
@@ -19,19 +21,22 @@ final class SignInViewController: BaseViewController {
...
     }
 
     // MARK: - Events
     @IBAction func signInWithGmailButtonPressed(_ sender: Any) {
-        async({ try await(GoogleApi.instance.signIn(viewController: self)) }, then: { [weak self] user in
-            self?.performSegue(withIdentifier: "RecoverSegue", sender: nil)
-        })
+        showSpinner()
+        googleAPI.signIn(viewController: self)
+            .then(on: .main) { [weak self] _ in
```
tomholub: when `googleAPI.signin()` returns a failed promise, how would that be handled?

tomholub: From the code, it seems it would get stuck with a spinner running forever.

Kharchevskyi: It should be somehow handled. But it wasn't handled before, so it's up to you. 
I will just remove spinner for now(see next commit, it still `wip`)

tomholub: It was handled before, every time any promise was used.

```Swift
        async({ try await(GoogleApi.instance.signIn(viewController: self)) }, then: { [weak self] user in
            self?.performSegue(withIdentifier: "RecoverSegue", sender: nil)
        })
```
This will call a method that has a default `fail` to hide spinner and show error alert, I think.

tomholub: If you don't like the `async` convenience thing for any reason, you'll have to make sure to manually decide what to do every time there's an error. Most of the time, it's appropriate to tell the user - that's why I had a method for it that does that.

Kharchevskyi: Yes but in this code you have a lot of memory leaks. 
self was captured in at least one time. 
Also in case of self is already nil async operation will cause app to crash

tomholub: Sure, I'm happy for all of that to get cleaned up.

In the previous code, all promise errors were handled, and this is something to be careful to preserve after the refactor, too.

Here's the async method it was calling: 

```Swift
    func async<T>(_ work: @escaping () throws -> T, then: @escaping (T) throws -> Void, fail alertMsg: String = "Action failed") {
        self.async(work, then: then, fail: { error in self.showErrAlert("\(alertMsg)\n\n \(error)") })
    }
```
Notice the default error handler. `showErrAlert` will also hide spinner, so this will be very convenient in 99% of error situations.

Everywhere you do a `.then`, you _have_  to do `.catch`, else you're silently dropping errors. 

So it would look something like:
```Swift
        googleAPI.signIn(viewController: self)
            .then(on: .main) { [weak self] result in ... }
            .catch(on: .main) { [weak self] error in self.showErrAlert("\(alertMsg)\n\n \(error)") }
```
as an example.

Kharchevskyi: This `async` func has side effects which better to avoid. 
Also better to use composition over inheritance, so try not to use BaseViewControllers and so on
Same with retain cycles/memory leaks. It will cause your app to crashes

tomholub: I agree with all of that. Please do it in a way that no promise rejections are silently dropped :+1:

Kharchevskyi:
Also there are a lot of memory leaks inside `Imap` and `GoogleApi` services. Better to refactor them as well

tomholub:
Do you want to also do that in this PR or another one?

FlowCrypt/Common/Constants.swift
```diff
@@ -18,6 +18,12 @@ enum Constants {
         }
     }
 
+    static let inboxDateFormatter: DateFormatter = {
```
tomholub: I think this one is unused & should be removed (unless I'm wrong). Now we use `convertDate` below.

tomholub:
And a test failure:

image

FlowCrypt/Controllers/SignIn/SignInViewController.swift
```diff
@@ -19,19 +21,25 @@ class SignInViewController: BaseViewController {
...
+                self?.hideSpinner()
+                self?.performSegue(withIdentifier: "RecoverSegue", sender: nil)
+            }
+            .catch(on: .main) { [weak self] _ in
+                self?.hideSpinner()
```
tomholub: Here it should also show alert to the user like before.

FlowCrypt/Core/Core Services/DataBaseService.swift
```diff
@@ -0,0 +1,42 @@
...
+
+protocol DataBaseService {
+    func save(message data: Data, isEmail: Bool) -> Promise<CoreRes.ParseDecryptMsg>
+}
+
+struct RealmDataBaseService: DataBaseService {
+    static let shared = RealmDataBaseService()
+
+    func save(message data: Data, isEmail: Bool) -> Promise<CoreRes.ParseDecryptMsg> {
+        return Promise { resolve, reject in
```
tomholub: Same comment as above, plus:
1) This does not need to return a Promise. There's no async functionality here. Promises are useful in situations where we need to handle a callback, with an escaping closure, and want to avoid callback hell. No callbacks here -> no need for Promises.
2) Don't hide decryption in a function that has nothing to do with it (save)
3) Don't call it `save` if there is nothing being saved

I recommend to delete this file completely, and put:
```Swift
            let realm = try Realm()
            let keys = PrvKeyInfo.from(realm: realm.objects(KeyInfo.self))
            let decrypted = try Core.parseDecryptMsg(
              encrypted: data,
              keys: keys,
              msgPwd: nil,
              isEmail: isEmail
           )
```
Into `MsgViewController` where it belongs.

There should not be a need for do/catch either, because you can `try` things inside a promise `.then` (in MsgViewController), and any thrown errors should then surface to `.catch`.

FlowCrypt/Controllers/Msg/MsgViewController.swift
```diff
@@ -7,179 +7,223 @@ import MBProgressHUD
...
+// MARK: - Message
+extension MsgViewController {
+    private func fetchMessage() {
+        guard let input = input else { return }
+        showSpinner(Language.loading, isUserInteractionEnabled: true)
+        imap.fetchMsg(message: input.objMessage, folder: input.path)
+            .then { [weak self] data -> Promise<CoreRes.ParseDecryptMsg> in
+                guard let self = self else { return Promise(ImapError.general) }
+                self.input?.bodyMessage = data
+                return self.db.save(message: data, isEmail: true)
```
tomholub: This looks wrong:
 - there is no need to save the message anywhere (yet)
 - the `db.save` method is not saving anything anyway, it is only loading private keys from db and decrypting
 - it should be VERY obvious that we are decrypting a message, `self.db.save` is a complete mismatch in terms of naming

Just decrypt it right here, as before.

FlowCrypt/Core/Core Services/MailDestinations.swift
```diff
@@ -0,0 +1,21 @@
...
+
+import Foundation
+
+enum MailDestination {
+    enum Gmail {
+        case trash
+
+        var path: String {
+            switch self {
+            case .trash: return "[Gmail]/Trash"
```
tomholub: For now OK, but in the future we will support more different email providers, and every email provider may call it something else :) ..

tomholub:
- [ ] mark as read doesn't work

tomholub:
- [ ] decrypt err showing different info than before

tomholub:
+2 hrs

tomholub:
@Kharchevskyi Later please take a look at my commits above and let me know if you have any comments

PR #61, opened 2019-08-22
issue 15 17 inbox refresh - squashed

tomholub:
#15 #17 
I squashed the commits from #59 into another branch in an attempt to be able to compare the code diff comfortably.

Issue #17, opened 2019-07-27
pull down inbox to refresh


tomholub:
Eg when I know I'm expecting a message, I intuitively keep pulling down

Issue #15, opened 2019-07-27
increase (double?) amount of loaded messages in inbox


tomholub:
currently can only see the very last few messages

PR #62, opened 2019-08-22
issue #15 #17 inbox refresh - squashed (#61)

tomholub:
#15 #17 

Code contributed by RomanKovalchuk2019 

* Changed folder naming

* All folder names is capitalized now

* Added toast for btn attach tap

* Removed capitalization

* Changed converter

* Added pull to refresh

* Changed number of loads

* Changed max number of messages

* Fixed unread status change on inbox

* Code refactoring

### DIFF
--- a/FlowCrypt/Common/Constants.swift
+++ b/FlowCrypt/Common/Constants.swift
@@ -11,7 +11,6 @@ enum Constants {
     // TODO: Anton - Use Themes instead of Constants
     static let NUMBER_OF_MESSAGES_TO_LOAD = 20
 
-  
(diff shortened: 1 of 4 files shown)


RomanKovalchuk2019:
Thank you, about refactoring my code

tomholub:
The change itself works well. The only issue was that the user would not see if there was error - I think it's better to surface errors to user, so that they know what's going on (eg no internet).

I've added an alert on error here: https://github.com/FlowCrypt/flowcrypt-ios/pull/62/commits/538e572b608e4c175e4e2085502bc6cd3a8f9b72

Issue #67, opened 2019-08-23
update core to latest version


tomholub:
the latest one uses `isFullyDecrypted` / `isFullyEncrypted` instead of `isDecrypted`.

![image](https://user-images.githubusercontent.com/6306961/63620854-ff1e4c80-c5e1-11e9-93ff-834c7cec5083.png)

after #57 lands

PR #70, opened 2019-08-25
issue 67 update core + check that imported key isFullyEncrypted

tomholub:
#67 
1 hour


tomholub:
I'll merge it today without your review, but please take a look and let me know if there are problems, in which case we can fix them in another PR.

PR #74, opened 2019-08-25
issue 30 retry imap connection drop

### LINKED ISSUE #30: stable connection could not be established
eg when opening a message

currently it just alerts

there should be a common handling mechanism for these errors, probably a retry/cancel button

tomholub:
fixes #30 
30 minutes

### DIFF
--- a/FlowCrypt/Common/Imap.swift
+++ b/FlowCrypt/Common/Imap.swift
@@ -23,7 +23,7 @@ final class Imap {
     struct EmptyError: Error {}
     private typealias ReqKind = MCOIMAPMessagesRequestKind
     private typealias Err = MCOErrorCode
-    private var lastErr: [String: Err] = []
+    private var lastErr: [String: Err] = [:]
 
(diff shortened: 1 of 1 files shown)


FlowCrypt/Common/Imap.swift
```diff
@@ -22,7 +22,8 @@ final class Imap {
 
     struct EmptyError: Error {}
     private typealias ReqKind = MCOIMAPMessagesRequestKind
-    private var lastErr = [String: MCOErrorCode]();
+    private typealias Err = MCOErrorCode
+    private var lastErr = [String: Err]();
```
Kharchevskyi: minor, but complier better. and remove `;` 
`private var lastErr: [String: Err] = []`

tomholub: Ah, ok!

tomholub: It seems like it wants `[:]`

PR #80, opened 2019-08-26
issue 77 show toast when msg not found

### LINKED ISSUE #77: Imap err code=19 when fetching certain messages
I'll take a look at this one

image

tomholub:
fixes #77 
1 hour

### DIFF
--- a/FlowCrypt/Controllers/Msg/MsgViewController.swift
+++ b/FlowCrypt/Controllers/Msg/MsgViewController.swift
@@ -132,7 +132,15 @@ extension MsgViewController {
             }
             .catch(on: .main) { [weak self] error in
                 self?.hideSpinner()
-                self?.showAlert(error: error, message: Language.could_not_open_message)
+                if let e = error as NSError?, e.code == Imap.Err.fetch.rawValue {
+                    // todo - the missing msg should be removed from the list in inbox view
+                    // reproduce: 1) load inbox 2) move msg to trash on another email client 3) open trashed message in inbox
+                    self?.showToast("Message not found in folder: \(input.path)")
+                } else {
+                    // todo - this should be a retry / cancel alert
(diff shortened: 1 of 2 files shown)


FlowCrypt/Controllers/Msg/MsgViewController.swift
```diff
@@ -132,7 +132,15 @@ extension MsgViewController {
...
-                self?.showAlert(error: error, message: Language.could_not_open_message)
+                if let e = error as NSError?, e.code == Imap.Err.fetch.rawValue {
+                    // todo - the missing msg should be removed from the list in inbox view
+                    // reproduce: 1) load inbox 2) move msg to trash on another email client 3) open trashed message in inbox
+                    self?.showToast("Message not found in folder: \(input.path)")
+                } else {
+                    // todo - this should be a retry / cancel alert
+                    self?.showAlert(error: error, message: Language.could_not_open_message + "\n\n\(error)")
+                }
+                self?.navigationController?.popViewController(animated: true)
```
Kharchevskyi: if you have a lot of actions on some block better to separate it to new private func. So it could be 
```
private func handleMessageSend(error: Error) {
  hideSpinner()
  if let e = error as NSError?, e.code == Imap.Err.fetch.rawValue {
  ...
  ...
  (p.s. without self.)
}
```

and inside catch it would be much cleaner 
```
.catch(on: .main) { [weak self] error in 
  self?.handleMessageSend(error: error)
}
```

and do you need to pop viewcontroller in case there is no error? It will show alert and immediately close controller

tomholub: So far, any exception we receive from that call is considered an error. Of course, maybe in the future there may be an exception that is not really an error where we don't want to pop the controller.. then we'd adjust it later.

PR #82, opened 2019-08-29
issue 78 message fetching

Kharchevskyi:
- added AppUrlHandler to handle app urls
- added common app error FCError
- replaced google api with user service. More testable. can be added new ways of authentication in future
- added DataManager to store app needed data. Can be refactored to use Keychain instead of UserDefaults for safety
- added app assemblies 
- refactored Imap service.


FlowCrypt/Common/Imap.swift
```diff
@@ -368,3 +314,19 @@ enum MailDestination {
         }
     }
 }
+
+
+enum FCError: Error {
```
tomholub: Why FC Error in IMAP file? I suggest somewhere more general, like original error file.

FlowCrypt/Core/DataManager.swift
```diff
@@ -0,0 +1,58 @@
+//
+//  DataManager.swift
```
tomholub: Please put all files in FlowCrypt/Core as before. 

Core folder should only contain code related to our JavaScript core library that we refuse across apps. Nothing else. 

You can use FlowCrypt/Common for the rest.

Kharchevskyi: Din't know about it. Will fix

tomholub:
There are also many test failures, please take a look @Kharchevskyi

tomholub:
Another issue - force cast / unwrap fail when starting the app

image

tomholub:
I have also fixed a fairly serious problem @Kharchevskyi where a nonsense string was being saved instead of pass phrase, causing problems. This was merged in a previous PR. You have to properly check that the app works after making changes.

image

Issue #76, opened 2019-08-26
icons spacing fix


tomholub:
This issue applies to icon spacing in all toolbars.

Currently looks like this:

![image](https://user-images.githubusercontent.com/6306961/63707771-3713d280-c822-11e9-9fa1-7343c105fbb2.png)

------------------------

 - [ ] there should be more space between icons (eg I added 20px more on the screenshot on the bottom)
 - [ ] there should be less space on the back/menu icon, to match amount of space on the right
 - [ ] the trash icon should be vertically squashed to fit well vertically with the other icons

![image](https://user-images.githubusercontent.com/6306961/63707840-662a4400-c822-11e9-9aca-a82347d3cfbb.png)

------------------------

The fixed result, very roughly:

![image](https://user-images.githubusercontent.com/6306961/63707926-a5589500-c822-11e9-85e5-9364cb0d3873.png)

PR #89, opened 2019-09-13
updated layout for navigation bar

Kharchevskyi:
https://github.com/FlowCrypt/flowcrypt-ios/issues/76
Screen Shot 2019-09-13 at 23 26 31


Kharchevskyi:
Fixed spacing as well as it can be with this icons. If it possible - provide with proper icons to achieve better result.

PR #97, opened 2019-09-22
issue #93 fix sender without pgp + refactor compose with Promise/await

### LINKED ISSUE #93: sending to someone who doesn't have pubkey
Regression

Should be showing a proper error, currently showing "Could not compose message"

tomholub:
close #93 / 3 hours

### DIFF
--- a/FlowCrypt/Common/AttesterApi.swift
+++ b/FlowCrypt/Common/AttesterApi.swift
@@ -13,24 +13,25 @@ final class AttesterApi {
 
     static let shared: AttesterApi = AttesterApi()
 
-    private let url = "https://flowcrypt.com/attester/"
+    private static let url = "https://flowcrypt.com/attester/"
 
     private init() { }
 
     func lookupEmail(email: String) -> Promise<PubkeySearchResult> {
-        return Promise<PubkeySearchResult> { [weak self] (resolve, reject) in
-            guard let self = self else { reject(CoreError.undefined); return }
-            let lookupUrl = self.url + "pub/\(self.normalize(email))"
+        return Promise { () -> PubkeySearchResult in
+            let lookupUrl = AttesterApi.url + "pub/\(AttesterApi.normalize(email))"
             let res = try await(URLSession.shared.call(lookupUrl, tolerateStatus: [404]))
-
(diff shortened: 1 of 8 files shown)


FlowCrypt/Common/AttesterApi.swift
```diff
@@ -13,24 +13,25 @@ final class AttesterApi {
 
     static let shared: AttesterApi = AttesterApi()
 
-    private let url = "https://flowcrypt.com/attester/"
+    private static let url = "https://flowcrypt.com/attester/"
```
tomholub: I added static so that `lookupEmail` does not have to do the `[weak self]` dance.

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -147,23 +148,7 @@ extension ComposeViewController {
     }
 
     @objc private func handleSendTap() {
-        sendMessage()
-    }
-
-    private func isInputValid() -> Bool {
```
tomholub: I moved this to where it's actually used.

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -186,68 +172,57 @@ extension ComposeViewController {
...
         showSpinner(Constants.sending)
 
-        attesterApi.lookupEmail(email: email)
-            .then(on: .main) { [weak self] recipientPub in
-                self?.handleLookupEmail(result: recipientPub, message: text, email: email)
-            }
-            .catch(on: .main) { [weak self] error in
-                self?.handleSend(error: error)
-            }
+        Promise<Void> { [weak self] in
```
tomholub: Here a security reviewer can get an easy overview, in a few lines.

FlowCrypt/Core/Core.swift
```diff
@@ -7,7 +7,6 @@ import JavaScriptCore
 enum CoreError: Error {
     case exception(String)
     case notReady(String)
-    case undefined
```
tomholub: Not sure what `undefined` error would be for, the `Core` library does not need it.

Issue #109, opened 2019-09-24
add support for generating keys


tomholub:
That way new users have an option to follow when iOS is the first FlowCrypt app they have ever used. Currently, they would be stuck on the recovery screen.

PR #110, opened 2019-09-24
issue #109 generate keys support

tomholub:
#109


FlowCrypt/Core/CoreTypes.swift
```diff
@@ -76,6 +102,12 @@ struct UserId: Encodable {
     let name: String;
 }
 
+extension UserId {
+    func toMime() -> String {
+        return "\(self.name) <\(self.email)>"
```
tomholub: There should be proper escaping in place (currently there is not), maybe the imap library has a method for this.

tomholub:
SemaphoreCI seems to have a some problems now, will try in the evening

Issue #58, opened 2019-08-21
migrate to Texture, declare ui in code


tomholub:
My biggest worry is that if we do all UI in code, there will be too much code, or code readability will suffer by mixing UI and logic.

Hopefully you can find a way to minimize that.

Kharchevskyi:
Can share you some code examples with texture. 
Basically you declare all your ui in views and all your business logic contains in services to handle this. 

Also all calculations to draw ui are calculated not on main thread so we can reach about 60fps with Texture. 

Will add Texture in separate branch.

tomholub:
Yes I have researched Texture. Actually the thing I like the most is that you can access your UI elements from any thread, so you don't have to always go to main thread to operate on UI. Is this correct? If yes, that could already clean up some code.

This issue is marked in the "**Later**" milestone, that means we will wait until issues in First and Second priority milestones are done.

PR #114, opened 2019-09-26
Refactoring InboxView controller

Kharchevskyi:
https://github.com/FlowCrypt/flowcrypt-ios/issues/58

- Using Promises instead of RXSwift
- Added batch fetching
- Implemented Texture UI for Inbox


FlowCrypt/Common/Taptic.swift
```diff
@@ -0,0 +1,47 @@
...
+//  Taptic.swift
+//  FlowCrypt
+//
+//  Created by Anton Kharchevskyi on 24.09.2019.
+//  Copyright © 2019 FlowCrypt Limited. All rights reserved.
+//
+
+import UIKit
+
+enum TapTicFeedback {
```
tomholub: What does TapTic mean and what is it used for? Thanks!

Kharchevskyi: for iPhones started I think from 8 - they edit taptic feedback when you do some actions. For example on. iPhone X when you tap on button you will have small feedback.

Kharchevskyi:
Do you have step to reproduce for black screen?

tomholub:
Just delete the app, reinstall it, then set it up by recovering from
backup. It seemed like the transition from backup screen to inbox was
broken.

PR #115, opened 2019-10-01
InboxViewController refactoring + Texture 

Kharchevskyi:
https://github.com/FlowCrypt/flowcrypt-ios/pull/114 + Xcode11


FlowCrypt/Controllers/Setup/SetupViewController.swift
```diff
@@ -174,12 +174,16 @@ extension SetupViewController {
...
         }.then(on: .main) { [weak self] in
-            self?.performSegue(withIdentifier: "InboxSegue", sender: nil)
+            self?.moveToMainFlow()
         }.catch(on: .main) { [weak self] error in
             self?.showAlert(error: error, message: "Could not finish setup, please try again")
         }
     }
 
+    private func moveToMainFlow() {
+        GlobalRouter().proceedAfterLogOut()
```
tomholub: `proceedAfterLogout` is not the best name considering how we use it. Maybe `GlobalRouter().reloadApp()` ?

Issue #127, opened 2019-10-22
tests should be re-worked to run in parallel


tomholub:
It works now. But tests should be re-worked to run in parallel, so that we don't waste time waiting on tests. I will make another issue.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/125#issuecomment-544949486_

PR #128, opened 2019-10-22
Executing tests in paralel

Kharchevskyi:
https://github.com/FlowCrypt/flowcrypt-ios/issues/127


tomholub:
Additionally, we should take advantage of semaphore parallelization. I'll take a look

Issue #131, opened 2019-11-02
minor style and functionality fixes


tomholub:
- [x] There should be light gray horizontal lines separating items in inbox
- [x] when I open side menu and click the green top, it turns grey, and it's not clear why
- [x] in the side menu green top, the email address text is too small, hard to read. No text in the app should be this small.
- [x] side menu items text is too small, should be the default size text like rest of the app (eg the size of text in inbox or in view title, just not bold)
- [x] should disable side menu bounce when scrolling up (the top green part will bounce down and show some white on the top)
- [x] I no longer see archive icon on top when I open message in inbox

PR #134, opened 2019-11-13
UI Updates

Kharchevskyi:
close https://github.com/FlowCrypt/flowcrypt-ios/issues/131

Simulator Screen Shot - iPhone 11 Pro Max - 2019-11-13 at 16 04 34
Simulator Screen Shot - iPhone 11 Pro Max - 2019-11-13 at 16 04 26
Simulator Screen Shot - iPhone 11 Pro Max - 2019-11-13 at 16 04 20
Simulator Screen Shot - iPhone 11 Pro Max - 2019-11-13 at 16 04 22

### DIFF
--- a/FlowCrypt/Controllers/Msg/MsgViewController.swift
+++ b/FlowCrypt/Controllers/Msg/MsgViewController.swift
@@ -57,8 +57,12 @@ final class MsgViewController: ASViewController<ASTableNode> {
         let mailInput = NavigationBarItemsView.Input(image: UIImage(named: "mail"), action: (self, #selector(handleMailTap)))
         let buttons: [NavigationBarItemsView.Input]
         switch input?.path {
-        case MailDestination.Gmail.trash.path: buttons = [infoInput, trashInput]
-        default: buttons = [infoInput, archiveInput, trashInput, mailInput]
+        case MailDestination.Gmail.trash.path:
+            buttons = [infoInput, trashInput]
+        case MailDestination.Gmail.inbox.path:
+            buttons = [infoInput, archiveInput, trashInput, mailInput]
+        default:
+            buttons = [infoInput, trashInput, mailInput]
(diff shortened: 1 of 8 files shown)


FlowCrypt/Controllers/Msg/MsgViewController.swift
```diff
@@ -58,8 +58,7 @@ final class MsgViewController: ASViewController<ASTableNode> {
         let buttons: [NavigationBarItemsView.Input]
         switch input?.path {
         case MailDestination.Gmail.trash.path: buttons = [infoInput, trashInput]
-        case MailDestination.Gmail.inbox.path: buttons = [infoInput, archiveInput, trashInput, mailInput]
-        default: buttons = [infoInput, trashInput, mailInput]
+        default: buttons = [infoInput, archiveInput, trashInput, mailInput]
```
tomholub: Because "archiveInput" will move move mail from inbox to "All Mail" (or more precisely, delete the inbox copy), it does not make sense to have the archive button on any mail outside of inbox.

Please look into it in more detail, to make sure the buttons in inbox differ from buttons in other folders, the way this switch case was originally written, and find out why it didn't work the way it was intended.

tomholub: You can look at what the Gmail app does, which is a good approximation of what we're trying to achieve when it comes to action buttons.

Kharchevskyi: You were right 
Google returns folders path as 

INBOX
[Gmail]/All Mail
[Gmail]/Drafts
[Gmail]/Important
[Gmail]/Sent Mail
[Gmail]/Spam
[Gmail]/Starred
[Gmail]/Trash

It leads to a bug with an archive button. Fixed in last commit

Issue #153, opened 2019-11-22
encrypt Realm?


tomholub:
> I meant maybe to separate storing some data to keychain storage or other please instead of Realm and UserDefaults. It's not really safe to save sensetive data in UserDefaults.

Definitely. We should be using whichever is the most secure method of storage.

Doesn't Realm support encryption? We could just store the Realm encryption password in keychain.

Kharchevskyi:
https://github.com/FlowCrypt/flowcrypt-ios/pull/156

PR #156, opened 2019-11-25
issue 153 Encrypt realm

Kharchevskyi:
close https://github.com/FlowCrypt/flowcrypt-ios/issues/153


FlowCrypt/Common/Constants.swift
```diff
@@ -13,11 +13,11 @@ enum Constants {
 
     enum EmailConstant {
         static let recoverAccountSearchSubject = [
-            "Your FlowCrypt Backup",
-            "Your CryptUp Backup",
-            "Your CryptUP Backup",
-            "CryptUP Account Backup",
-            "All you need to know about CryptUP (contains a backup)",
+            "1Your FlowCrypt Backup",
```
tomholub: This should not be checked into the repo. If you need to test situation without a backup, you should log into another account.

Kharchevskyi: this is still WIP

tomholub: I'm mentioning it because last time it was also there when PR was approved for review. It could have slipped into production code if I didn't notice it. It's safer to just use more accounts. I can share some accounts to use.

Kharchevskyi: Safer for sure. 
Would be nice to have some other accounts for testing.

FlowCrypt/Common/Services/DataManager.swift
```diff
@@ -9,59 +9,71 @@
...
+    func publicKey() -> String? {
+        encryptedStorage.publicKey()
     }
 
-    func currentUser() -> User? {
-        guard let data = userDefaults.object(forKey: Constants.userKey) as? Data else { return nil }
-        return try? PropertyListDecoder().decode(User.self, from: data)
+    func startForNew(user: User?, with token: String?) {
+        localStorage.saveCurrent(user: user)
+        encryptedStorage.ecnryptFor(email: user?.email)
```
tomholub: Is it necessary to encrypt encrypted storage separately for each email? If I remember correctly we settled on not needing email to initialize encrypted storage.

FlowCrypt/Common/Services/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -0,0 +1,109 @@
...
+    private let accessCheck: () -> (Bool)
+
+    init(keychainHelper: KeyChainServiceType = KeyChainService(), accessCheck: @escaping () -> (Bool)) {
+        self.keychainHelper = KeyChainService()
+        self.accessCheck = accessCheck
+    }
+
+    private var encryptedConfiguration: Realm.Configuration? {
+        guard canHaveAccessToStorage else { return nil }
+        let configuration = Realm.Configuration(encryptionKey: self.keychainHelper.getEncryptedKey())
```
tomholub: rename `getStorageEncryptionKey`

FlowCrypt/Common/Services/Encrypted Storage/KeyChainService.swift
```diff
@@ -0,0 +1,71 @@
...
+    private let tag = "flowcrypt-realm-encryption-key"
+    private let keyGenerator: KeychainKeyGeneratorType
+
+    init(
+        keyGenerator: KeychainKeyGeneratorType = KeychainKeyGenerator()
+    ) {
+        self.keyGenerator = keyGenerator
+    }
+
+    func saveEncryptedKey() -> KeyChainStatus {
```
tomholub: `generateAndSaveStorageEncryptionKey`

FlowCrypt/Common/Services/Encrypted Storage/KeychainKeyGenerator.swift
```diff
@@ -0,0 +1,23 @@
...
+//  Copyright © 2019 FlowCrypt Limited. All rights reserved.
+//
+
+import Foundation
+
+protocol KeychainKeyGeneratorType {
+    func generateKeyData() -> Data
+}
+
+struct KeychainKeyGenerator: KeychainKeyGeneratorType {
```
tomholub: Instead of this you can use `CoreHost.getSecureRandomByteNumberArray(64)`, and convert that to `Data` if needed.

FlowCrypt/Common/Services/Local Storage/LocalStorageTest.swift
```diff
@@ -0,0 +1,36 @@
...
+//  LocalStorageTest.swift
+//  FlowCryptTests
+//
+//  Created by Anton Kharchevskyi on 25.11.2019.
+//  Copyright © 2019 FlowCrypt Limited. All rights reserved.
+//
+
+import XCTest
+
+class LocalStorageTest: XCTestCase {
```
tomholub: Thanks for adding tests!

FlowCrypt/Model/Token.swift
```diff
@@ -0,0 +1,19 @@
...
+//  FlowCrypt
+//
+//  Created by Anton Kharchevskyi on 25.11.2019.
+//  Copyright © 2019 FlowCrypt Limited. All rights reserved.
+//
+
+import Foundation
+import RealmSwift
+
+final class Token: Object {
```
tomholub: If this is access token, then we should probably call it `EmailAccessToken`.

Maybe easier to keep access token as a column in User table? Making a separate table may be an overkill.

tomholub:
Also the auto-merge broke it :/ @Kharchevskyi please take a look

FlowCrypt/Common/Services/DataManager.swift
Kharchevskyi: Nice catch. I just need to logOut() here. Will fix in next commit

FlowCrypt/Common/Services/Encrypted Storage/KeychainKeyGenerator.swift
Kharchevskyi: Fixed

FlowCrypt/Model/Token.swift
Kharchevskyi: It's easier to keep them separate. And I don't whant to have whole user in places where I just need a token. For example some "ServerManager" don't need to know about User itself it just need a token
Renamed with `EmailAccessToken`

tomholub: Ok, got it. Lets do it your way then, and keep them separate.

Issue #123, opened 2019-10-16
handle app opened when offline


tomholub:
1) set up the app to see inbox
2) kill it
3) go offline
4) open app to show inbox

Various errors will show up, and the loader will not stop loading. Instead, loader should go away and where there are normally list of messages, it should say: `Cannot load inbox: connection error (offline)`.

tomholub:
Same for list of folders in the menu, alternatively we could start caching these.

PR #157, opened 2019-11-27
Handling connection error inside inbox and folders

Kharchevskyi:
https://github.com/FlowCrypt/flowcrypt-ios/issues/123


Kharchevskyi:
IMG_3188

PR #164, opened 2019-12-19
Search inbox

Kharchevskyi:
close #164


Kharchevskyi:
https://github.com/FlowCrypt/flowcrypt-ios/pull/164

tomholub:
image

FlowCrypt/Common/Imap/ImapHelper.swift
```diff
@@ -0,0 +1,63 @@
...
+        total: Int,
+        from: Int
+    ) -> MCOIndexSet
+    
+    func createSearchExpressions(
+        from possibleExpressions: [MCOIMAPSearchExpression]
+    ) -> MCOIMAPSearchExpression?
+}
+
+struct ImapHelper: ImapHelperType {
```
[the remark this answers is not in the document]
Kharchevskyi: It's a separate unit of code, just easier to cover with tests. If I move it to Imap in this case I need to move the whole Imap class to Tests target including Mailcore lib.

For naming - also don't like this name but have no idea how to rename it(

FlowCrypt/Controllers/SearchViewController.swift
```diff
@@ -0,0 +1,321 @@
...
+ 
+    @objc private func handleInfoTap() {
+        #warning("ToDo")
+        showToast("Email us at human@flowcrypt.com")
+    }
+}
+
+// MARK: - MessageHandlerViewConroller
+
+extension SearchViewController: MessageHandlerViewConroller {
```
tomholub: How is this used? The search screen does not have any delete or markAsRead buttons. Isn't this duplicating code from MsgView?

Kharchevskyi: I will fix this bug in the video. 
No, it isn't code duplicate. MessageHandlerViewConroller handles read and remove actions after you open the message. 
For example you have searched for google -> received 10 results -> open first message -> delete it. In this case, it will remove this message from the list. Same with reading.

tomholub: `extension InboxViewController: MessageHandlerViewConroller {` contains nearly identical code. Why extend the same functionality two times? Why not just inherit it? I do see this as code duplication.

FlowCrypt/Common/Imap/ImapHelper.swift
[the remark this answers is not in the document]
tomholub: I don't think we have to test these two functions separately with unit tests.

Instead we should test the app as a whole (we discussed it in other issues too).

Just move ImapHelper methods into Imap, and write tests that test user behavior, not a single method.

FlowCrypt/Controllers/SearchViewController.swift
Kharchevskyi: Why it's extneded twice? There is a protocol MessageHandlerViewConroller which have 2 functions 
```
func openMessageIfPossible(with message: MCOIMAPMessage, path: String)
func handleMessage(operation: MsgViewController.MessageAction, message: MCOIMAPMessage)
```

And I made a protocol with a default implementation for `openMessageIfPossible` function
```
extension MessageHandlerViewConroller where Self: UIViewController {
    func openMessageIfPossible(with message: MCOIMAPMessage, path: String) {
  
    }
}
```
but because datasource is different in every viewcontroller - there is no way to make default implementation for

`func handleMessage(operation: MsgViewController.MessageAction, message: MCOIMAPMessage)`

tomholub: I see it a little better now. But `handleMessage` seems to be the same for both implementations. Let me take a look.

tomholub:
Ok we'll leave the ImapHelper but the name could really improve.. not sure what to call it, because the content is too diverse. Will leave it as is for now.

PR #171, opened 2020-01-08
Issue 167 key settings provider

### LINKED ISSUE #167: re-consider KeySettingsProvider
1) keys in our own storage are expected to be parseable. If any key from our own storage cannot be parsed, that should be a fatal error.
2) We don't need to differentiate types of errors here, unnecessary complication
3) after doing the above, the code will simplify to the extent that having a separate class seems an overkill

tomholub:
close #167


tomholub:
I have refactored a small part of the app to demonstrate my displeasure with several patterns I'm seeing across the codebase, and would like to discuss them with you.

My main concern is: **unwanted complexity**. In more detail:

1) `KeySettingsProvider` featured dependency injection, error handling, error types, a prototype etc when the actual functionality could have been a simple method in `KeySettingsViewController` instead (unwanted bloat)
2) Although it returned several error types, none of the errors were actually handled (empty code)
3) `KeySettingsItem` was defined even though it contains more or less the same data as `KeyDetails`. When existing data type suffices, don't create new data types. (duplicate code)
4) We should focus our efforts on writing tests that mimic user behavior, not tests that verify we can concatenate a string: `XCTAssert(keySettingsItem?.users == "ilon@tesla.com donald@trump ")` (unhelpful test)

tomholub:
@Kharchevskyi haven't received any feedback from you here, so I'll merge it

Issue #185, opened 2020-01-31
app loading initial checks


tomholub:
> Tom I think I need to rework a little bit to make things more clear
> I will create some kind of step approach for app loading and initializing and bootstrapingWhere we can for sure know that we
> • keychain is accessible
> • storage is encrypted
> • Imap session can be started
> so we know for sure that we are ready to work with app to avoid this problem in a future.

Issue #183, opened 2020-01-31
remove canHaveAccessToStorage and accessCheck from EncryptedStorage


tomholub:
remove `canHaveAccessToStorage` and `accessCheck` from `EncryptedStorage` , and check these things externally, and don't call or use or initiate `EncryptedStorage`  if not true. That means returning realm config will not be Optional: `private var encryptedConfiguration: Realm.Configuration {`

That means I think overall signature will change to return fewer optionals:
```swift
protocol EncryptedStorageType {
    func addKeys(keyDetails: [KeyDetails], passPhrase: String, source: KeySource)
    func saveToken(with string: String?)
    func currentToken() -> String?
    func publicKey() -> String
    func keys() -> Results<KeyInfo>
}
```

PR #186, opened 2020-02-11
App loading checks. Realm migration. 

Kharchevskyi:
Add step approach for app loading: close https://github.com/FlowCrypt/flowcrypt-ios/issues/185
* Setup Core of the project
* Setup database and perform migration if needed
* Setup services required for user notification (Google SDK)
* Setup a session. Create or renew Imap session
* Check if user has encrypted storage
* Presenting flow for user

Remove encrypted storage access checks: close https://github.com/FlowCrypt/flowcrypt-ios/issues/183
Realm Migration: close https://github.com/FlowCrypt/flowcrypt-ios/issues/190
Minor fix for renewing access token


FlowCrypt/Common/Services/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -84,24 +110,109 @@ extension EncryptedStorage: LogOutHandler {
...
+        guard let documentDirectory = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true).first else {
+            assertionFailure("No path direction for documentDirectory")
+            return
+        }
+
+        let unencryptedRealmPath = documentDirectory + "/default.realm"
+        let encryptedPath = documentDirectory + "/" + Constants.encryptedDBName
+
+        let isUnencryptedRealmExsist = fileManager.fileExists(atPath: unencryptedRealmPath)
+        let isEncryptedRealmExsist = fileManager.fileExists(atPath: encryptedPath)
```
tomholub: The right language use here is `does`, not `is`

FlowCrypt/Common/Services/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -84,24 +110,109 @@ extension EncryptedStorage: LogOutHandler {
...
+
+extension EncryptedStorage {
+    func performMigrationIfNeeded(_ completion: @escaping () -> Void) {
+        guard let documentDirectory = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true).first else {
+            assertionFailure("No path direction for documentDirectory")
+            return
+        }
+
+        let unencryptedRealmPath = documentDirectory + "/default.realm"
+        let encryptedPath = documentDirectory + "/" + Constants.encryptedDBName
```
tomholub: call it `encryptedRealmPath` for consistency

FlowCrypt/Common/Services/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -18,23 +24,43 @@ protocol EncryptedStorageType {
...
+        static let schemaVersion: UInt64 = 1
+        static let encryptedDBName = "encrypted.realm"
+    }
 
     let keychainService: KeyChainServiceType
-    private var canHaveAccessToStorage: Bool { accessCheck() }
-    private let accessCheck: () -> (Bool)
+    private let fileManager: FileManager
+
+    var isEncrypted: Bool {
```
tomholub: Should there ever be situation when `EncryptedStorage.isEncrypted` should return false?

FlowCrypt/Common/Services/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -84,24 +110,109 @@ extension EncryptedStorage: LogOutHandler {
...
+            .forEach {
+                destroyStorage(at: $0)
             }
+    }
+
+    private func destroyStorage(at url: URL) {
+        do {
+            try fileManager.removeItem(at: url)
         } catch CocoaError.fileNoSuchFile {
+            debugPrint("Realm at url \(url) not existed")
```
tomholub: `Realm at url \(url) did not exist`

FlowCrypt/Common/Services/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -84,24 +110,109 @@ extension EncryptedStorage: LogOutHandler {
...
+        let isEncryptedRealmExsist = fileManager.fileExists(atPath: encryptedPath)
+
+        guard isUnencryptedRealmExsist && !isEncryptedRealmExsist else {
+            debugPrint("Migration not needed")
+            completion()
+            return
+        }
+
+        if !isEncryptedRealmExsist && !isUnencryptedRealmExsist {
+            debugPrint("Realm was not exist")
```
tomholub: "did not exist"

FlowCrypt/Common/Services/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -84,24 +110,109 @@ extension EncryptedStorage: LogOutHandler {
...
+        let isUnencryptedRealmExsist = fileManager.fileExists(atPath: unencryptedRealmPath)
+        let isEncryptedRealmExsist = fileManager.fileExists(atPath: encryptedPath)
+
+        guard isUnencryptedRealmExsist && !isEncryptedRealmExsist else {
+            debugPrint("Migration not needed")
+            completion()
+            return
+        }
+
+        if !isEncryptedRealmExsist && !isUnencryptedRealmExsist {
```
tomholub: Typos "Exsist"

Is there any xcode plugin that looks for language typos?

FlowCrypt/Common/Services/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -84,24 +110,109 @@ extension EncryptedStorage: LogOutHandler {
...
+            return
+        }
+
+        let unencryptedRealmPath = documentDirectory + "/default.realm"
+        let encryptedPath = documentDirectory + "/" + Constants.encryptedDBName
+
+        let isUnencryptedRealmExsist = fileManager.fileExists(atPath: unencryptedRealmPath)
+        let isEncryptedRealmExsist = fileManager.fileExists(atPath: encryptedPath)
+
+        guard isUnencryptedRealmExsist && !isEncryptedRealmExsist else {
```
tomholub: Typos "Exsist"

FlowCrypt/Common/Services/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -84,24 +110,109 @@ extension EncryptedStorage: LogOutHandler {
...
+        if let key = keychainService.getOldKeychainKey() {
+            if let previouslyEncrypted = try? Realm(configuration: Realm.Configuration(encryptionKey: key)) {
+                debugPrint("use previouslyEncrypted config \(previouslyEncrypted) for migration")
+                oldRealm = previouslyEncrypted
             }
-        } catch CocoaError.fileNoSuchFile {
+        }
+
+         guard let realm = oldRealm else {
+            debugPrint("Relam was not exist. Migration not needed")
```
tomholub: "did not exist"

FlowCrypt/Common/Services/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -84,24 +110,109 @@ extension EncryptedStorage: LogOutHandler {
...
+
+        // write copy of realm db
+        do {
+            try realm.writeCopy(
+                toFile: URL(fileURLWithPath: encryptedPath),
+                encryptionKey: realmKey
+            )
         } catch let error {
-            fatalError("Could not delete encryptedConfiguration: \(error)")
+            debugPrint(error)
```
tomholub: should this not fail with a fatal error?

FlowCrypt/Common/Services/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -84,24 +110,109 @@ extension EncryptedStorage: LogOutHandler {
...
+                try? self?.fileManager.removeItem(atPath: unencryptedRealmPath)
+                completion()
+            }
+        ) 
+
+        _ = try! Realm(configuration: configuration)
     }
-}
 
+    private func isEncryptedRealmExsist() -> Bool {
```
tomholub: `does` and `Exist`

FlowCrypt/Common/Services/Encrypted Storage/KeyChainService.swift
```diff
@@ -82,4 +78,39 @@ struct KeyChainService: KeyChainServiceType {
...
         return validKey
     }
+
+    func generateNewKey() {
+        UserDefaults.standard.set(nil, forKey: Constants.indexSecureKeychainPrefix.rawValue)
+        genertateAndSaveKey()
+    }
+
+    #warning("use only for migration purposes") 
+    func getOldKeychainKey() -> Data? {
```
tomholub: The last public version does not use Realm encryption at all. And future versions will always use the same key going forward. Does that not mean that we should not need this method?

FlowCrypt/Common/Services/Local Storage/LocalStorage.swift
```diff
@@ -16,7 +16,7 @@ protocol LocalStorageType {
 struct LocalStorage: LocalStorageType {
 
     private enum Constants: String, CaseIterable {
-        case indexCurrentUser = "indexCurrentUser"
+        case indexCurrentUser = "keyCurrentUser"
```
tomholub: Some other day we may want to migrate this too, to `indexCurrentUser`. Because our app does encryption, "key" is often used for private information then for storage access keys (indexes). Therefore it will be better to call these "index" to remove ambiguity.

But for this time, we'll leave it as is to keep things simple, as we agreed.

FlowCrypt/LaunchFlow/LaunchFlowController.swift
```diff
@@ -0,0 +1,31 @@
...
+    let flowController: FlowController
+
+    init(flowController: FlowController) {
+        self.flowController = flowController
+    }
+
+    @discardableResult
+    func startFlow(with launchContext: LaunchContext) -> Bool {
+        return flowController.execute(with: launchContext) { completion in
+            log("Luanch execution finished \(completion)")
```
tomholub: It's a pity to not give it start time, since we're logging it, might as well know how long it took.

FlowCrypt/LaunchFlow/FlowController.swift
```diff
@@ -0,0 +1,77 @@
...
+//
+//  Created by Anton Kharchevskyi on 07/02/2020.
+//  Copyright © 2020 FlowCrypt Limited. All rights reserved.
+//
+
+import Foundation
+
+protocol FlowController: FlowStepHandler { }
+
+final class DefaultFlowController: FlowController {
```
tomholub: Better call all of these `DefaultLaunchFlowController`, `LaunchFlowController` and similar. Add the `Launch`, else it's unclear what it does.

FlowCrypt/Common/Services/Logger.swift
```diff
@@ -8,7 +8,7 @@
 
 import Foundation
 
-func log(_ message: String, error: Error?, res: Any?, start: DispatchTime) {
+func log(_ message: String, error: Error? = nil, res: Any? = nil, start: DispatchTime = .now()) {
```
tomholub: As noted elsewhere, the start should be passed explicitly to know how long it took, and therefore there it does not need a default.

FlowCrypt/Common/Services/Encrypted Storage/EncryptedStorage.swift
Kharchevskyi: It's just an additional check to be sure and doesn't proceed farther if the implementation of KeychainService or EncryptedStorage will be changed in the future.

FlowCrypt/Common/Services/Local Storage/LocalStorage.swift
Kharchevskyi: Better to create a ticket for that migration. If we just change a key here it will log out all users.

tomholub: I understand, will make a ticket

FlowCrypt/LaunchFlow/FlowController.swift
Kharchevskyi: DefaultFlowController is responsible for executing LaunchFlow steps.
Renamed just to `FlowController` 

`LaunchFlowController` starts flow with some launch context.

FlowCrypt/Common/Services/Encrypted Storage/EncryptedStorage.swift
tomholub: I think we should not be creating checks for future code changes that have not yet occurred.  We may be building for future that never arrives, and creates code bloat/confusion. I'd rather see it removed, `EncryptedStorage` should always be encrypted.

tomholub:
But the tests are failing for some reason @Kharchevskyi

tomholub:
todo: 
 - [ ] test migration
 - [ ] remove local storage token after migration done
 - [ ] test AppStartup error rendering
 - [ ] fix failing tests
 - [ ] run UITests

tomholub:
Cannot run UI tests, stuck on #191

Kharchevskyi:
This solution is not testable. And not scalable. 
This complexity was added for the purpose. 

Every of this launch steps ( CoreLaunchStep, AuthLaunchStep,DataBaseLaunchStep,SessionLaunchStep,MainLaunchStep) can be tested separately or mocked for tests. All of them has it's own responsibility. 

LaunchFlow and LaunchFlowStepFactory create steps, base on a context. In case we receive push notification or open app from deeplink it will create Launchsteps which are responsible for handling events. 

I don't agree about over-engineering

This was scalable, testable and reusable. Now it's only one god object class.

tomholub:
To me, simplicity is more important. I need our code to be easily readable and reviewable, and I need it to be of minimal complexity.

> Every of this launch steps ( CoreLaunchStep, AuthLaunchStep,DataBaseLaunchStep,SessionLaunchStep,MainLaunchStep) can be tested separately or mocked for tests. All of them has it's own responsibility.

Factory/step approach:
 - 381 lines over 12 classes, 11 files
 - difficult to follow business logic
 - each piece testable individually
 - difficult to do security review

Single class approach:
 - 60 lines, 1 class
 - easy to follow business logic, chronological
 - hard to test individually
 - trivial security review

For our usecase, the tradeoffs are different from a regular app. We need **simplicity, readability above everything else**.

> LaunchFlow and LaunchFlowStepFactory create steps, base on a context. In case we receive push notification or open app from deeplink it will create Launchsteps which are responsible for handling events.

To handle deeplinks, edit `chooseView`. Add an argument containing the possible deeplink + 1 if.

------------------

The added complexity comes with a tradeoff, and the tradeoff is not worth it for us.

tomholub:
As for testing, consider this:

- CoreLaunchStep - we already test core extensively
- AuthLaunchStep - just fills a few constants, nothing to test
- DataBaseLaunchStep - if this fails, no UITests will work, already tested. Plus we aren't going to test migrations, anyway.
- SessionLaunchStep - if this doesn't work, it will show in other tests
- MainLaunchStep - This can be easily tested with UITests

I prefer tests that are not tightly coupled to implementation details. Any tests you write that would be testing the factory/step design directly would necessarily be tightly tied to implementation. That is not the kind of tests I want to focus on.

PR #209, opened 2020-02-19
Feature/issue 200 recipient text field

### LINKED ISSUE #200: a way to enter several recipients
Currently, the recipients field is just a simple text field. But instead we should have a way to enter recipient emails.

Have a look at FlowCrypt Android App or Browser Extension. 

**Within scope for this PR:** ability to enter several recipients, and these recipients should show as distinct "chips". Ability to remove each recipient (by putting carrot to the right of them and hitting backspace? take a look at other email apps, particularly Gmail, and do something similar).

**Out of scope:** evaluating recipients and coloring them (which will be in issue #201 separately). But please choose UI elements that will be easy to color, select, copy, paste and otherwise manipulate, similar to what Android does (or any other email app).

Kharchevskyi:
close #200 
* Add FlowCryptUI and FlowCryptCommon Targets 
* Add single view application target for testing nodes
  * Easier to debug than Playgrounds because of console and view hierarchy. 
  * Add possibility to compile FlowCryptUI target every time test app build it (for playground you need to do it manually)
  * Playground doesn't show crashes or exceptions
  * Add Readme file inside test app
* Add multiple recipient input field


FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -145,27 +151,57 @@ extension ComposeViewController {
...
-                subject: subject,
-                message: text,
-                email: email
-            )
-            try await(self.imap.sendMail(mime: encrypted.mimeEncoded))
+
+            let encryptedMessages = lookupRes.compactMap { (searchResult: PubkeySearchResult) -> CoreRes.ComposeEmail? in
+                guard let armored = searchResult.armored else { assertionFailure(); return nil }
+                return self.encryptMsg(
+                    pubkeys: [myPubkey, armored],
```
tomholub: This is wrong, there should be a single email message that is encrypted for all pubkeys, and mine

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -144,27 +195,53 @@ extension ComposeViewController {
...
                 return false
             }
+
             let encrypted = self.encryptMsg(
-                pubkeys: [myPubkey, recipientPubkey],
+                pubkeys: allRecipientPubs + [myPubKey],
                 subject: subject,
                 message: text,
-                email: email
+                email: recipients.map { $0.email }
```
tomholub: here `email:` could be renamed to `to:` since it's now an array. Also, in the future we may add `cc:` and `bcc:` so it makes sense to distinguish it now.

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -175,12 +252,12 @@ extension ComposeViewController {
         navigationController?.popViewController(animated: true)
     }
 
-    private func encryptMsg(pubkeys: [String], subject: String, message: String, email: String) -> CoreRes.ComposeEmail {
+    private func encryptMsg(pubkeys: [String], subject: String, message: String, email: [String]) -> CoreRes.ComposeEmail {
```
tomholub: that means here also change to `to: [String]`

FlowCryptCommon/UIVIewExtensions.swift
```diff
@@ -0,0 +1,107 @@
+//
+// © 2017-2019 FlowCrypt Limited. All rights reserved.
+//
+
+import UIKit
+
+public extension UIView {
+    @discardableResult
+    func bordered(color: UIColor, width: CGFloat) -> Self {
```
tomholub: I like these kind of extensions, they increase code clarity, remove repetitive code, improve readability. Well done.

Issue #204, opened 2020-02-14
Use google contacts api to search for recipients


tomholub:
This means we will have to add another oauth permission when logging in with google. If the permission is not available in current access token (users who installed on previous version), you could just show a toast `To enable searching contacts, please Log out and set up FlowCrypt again`.

Then when I type `tom` in the recipient input field, I should see search results of people to tap on. See what Browser Extension does.

This will use the Google Contacts api. Here is an example of it being in use: https://github.com/FlowCrypt/flowcrypt-browser/blob/master/extension/js/common/api/google.ts#L44 the `Google.contactsGet` method.

Issue #165, opened 2020-01-03
describe code naming conventions


tomholub:
We must put together code documentation that will make it easier for me to do code reviews, and also make it easier to orientate security reviewers:

- what is a decorator, what should go in it, where should it be
- what is a helper, what should go in it, where should it be
- what is a service, what should go in it, where should it be
- what is a view controller, what should go in it, where should it be
- what is node, what should go in it, where should it be
- what is provider, what should go in it, where should it be
- what is manager, what should go in it, where should it be

I feel like there is some overlap between some of the chosen names, so it would be good to define them and write it down.

PR #211, opened 2020-02-28
Search Google contacts

Kharchevskyi:
close https://github.com/FlowCrypt/flowcrypt-ios/issues/204

- Main changes for this issue are located in `ComposeViewController, GoogleService, GoogleContactsResponse
`
- Ordered and simplified folder structure for  https://github.com/FlowCrypt/flowcrypt-ios/issues/165
- Marked `TODO` for nodes to move them to the common target
- Get rid of redundant "ViewController + setup" and "helper" folders


FlowCrypt/Extensions/URLSessionExtension.swift
```diff
@@ -24,7 +24,7 @@ extension URLSession {
             self.dataTask(with: urlRequest) { data, response, error in
                 let res = response as? HTTPURLResponse
                 let status = res?.statusCode ?? GeneralConstants.Global.generalError
-                print("URLSession.call status:\(status) ms:\(start.millisecondsSince) \(urlRequest.httpMethod ?? "GET") \(urlRequest.url?.absoluteString ?? "??")")
+//                debugPrint("URLSession.call status:\(status) ms:\(start.millisecondsSince) \(urlRequest.httpMethod ?? "GET") \(urlRequest.url?.absoluteString ?? "??")")
```
tomholub: I would still print these, it's useful to see calls / results, unless you have some other ways to inspect http requests?

tomholub:
Code look, functionality good but login screen gives a privacy error. I'll need to check Google.

FlowCrypt/Extensions/URLSessionExtension.swift
Kharchevskyi: It will spam a lot. 
Maybe there is an idea to have different levels for logging? 
Like UI logs, Imap logs, session logs and so on.
And also be able to set a list of logs to inspect.

Kharchevskyi:
I had this privacy error on first time I logged in. I think this error for not publicly distributed app

tomholub:
It started showing again after adding search. Most likely not your fault,
it has to do with settings on google, I will check it.

You can test it by revoking the permissions on google account settings,
then trying again. Maybe I didn't verify the app for use of contacts.
--
Tom James Holub <http://holub.me/>

tomholub:
I went ahead and merged it, but will need to make sure to look into the google permissions later.

Issue #146, opened 2019-11-21
quote previous reply


tomholub:
when replying to message previous reply should be quoted, like this

```
reply from iOS

On 2019-11-21 at 12:11, mindoutofframe@gmail.com wrote:
> This was the original message
> Now it is quoted
```

I'll add more details

PR #214, opened 2020-03-06
Quote previous reply

Kharchevskyi:
close https://github.com/FlowCrypt/flowcrypt-ios/issues/146

### DIFF
--- a/FlowCrypt/Controllers/Compose/ComposeDecorator.swift
+++ b/FlowCrypt/Controllers/Compose/ComposeDecorator.swift
@@ -74,15 +74,11 @@ struct ComposeDecorator: ComposeDecoratorType {
         dateFormatter.timeStyle = .short
         let time = dateFormatter.string(from: info.sentDate)
 
-        let from = info.recipient?.displayName
-            ?? info.recipient?.mailbox
-            ?? "no subject"
+        let from = info.recipient?.mailbox ?? "unknown sender"
 
         let text: String = "\n\n"
-            + "compose_reply_from_title".localized
-            + "\n"
             + "compose_reply_from".localizeWithArguments(date, time, from)
-            + "\n\n"
+            + "\n"
 
         let message = " > " + info.message.replacingOccurrences(of: "\n", with: "\n > ")
 
--- a/FlowCrypt/Controllers/Compose/ComposeDecorator.swift
+++ b/FlowCrypt/Controllers/Compose/ComposeDecorator.swift
@@ -13,7 +13,8 @@ protocol ComposeDecoratorType {
     func styledTextViewInput(with height: CGFloat) -> TextViewCellNode.Input
     func styledTextFieldInput(with text: String) -> TextFieldCellNode.Input
     func styledRecipientInfo(with email: String) -> InfoCellNode.Input
-    func styledTitle(with text:String?) -> NSAttributedString?
+    func styledTitle(with text: String?) -> NSAttributedString?
+    func styledReplyQuote(with input: ComposeViewController.Input) -> NSAttributedString
 }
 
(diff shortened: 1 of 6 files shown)


FlowCrypt/Controllers/Compose/ComposeDecorator.swift
```diff
@@ -59,6 +60,34 @@ struct ComposeDecorator: ComposeDecoratorType {
...
+
+        let date = dateFormatter.string(from: info.sentDate)
+
+        dateFormatter.dateStyle = .none
+        dateFormatter.timeStyle = .short
+        let time = dateFormatter.string(from: info.sentDate)
+
+        let from = info.recipient?.displayName
+            ?? info.recipient?.mailbox
+            ?? "no subject"
```
tomholub: should say `"unknown sender"`

FlowCrypt/Controllers/Compose/ComposeDecorator.swift
```diff
@@ -59,6 +60,34 @@ struct ComposeDecorator: ComposeDecoratorType {
...
+        dateFormatter.dateStyle = .short
+        dateFormatter.timeStyle = .none
+
+        let date = dateFormatter.string(from: info.sentDate)
+
+        dateFormatter.dateStyle = .none
+        dateFormatter.timeStyle = .short
+        let time = dateFormatter.string(from: info.sentDate)
+
+        let from = info.recipient?.displayName
```
tomholub: No need to use display name, just go with email

FlowCrypt/Resources/en.lproj/Localizable.strings
```diff
@@ -35,6 +35,8 @@
 "compose_no_pub_sender" = "Missing sender public key. Is FlowCrypt iOS app well set up?";
 "compose_error" = "Could not compose message";
 "compose_reply_successfull" = "Reply successfully sent";
+"compose_reply_from_title" = "Reply from iOS";
+"compose_reply_from" = "On %@ at %@ %@ wrote:"; // Date, time, subject
```
tomholub: Date, time, **sender**

FlowCrypt/Controllers/Compose/ComposeDecorator.swift
```diff
@@ -59,6 +60,34 @@ struct ComposeDecorator: ComposeDecoratorType {
...
+            ?? info.recipient?.mailbox
+            ?? "no subject"
+
+        let text: String = "\n\n"
+            + "compose_reply_from_title".localized
```
tomholub: Don't add this title, it's not needed

FlowCrypt/Controllers/Compose/ComposeDecorator.swift
```diff
@@ -59,6 +60,34 @@ struct ComposeDecorator: ComposeDecoratorType {
...
+            + "\n"
+            + "compose_reply_from".localizeWithArguments(date, time, from)
+            + "\n\n"
```
tomholub: I think it should be just one newline, not two

FlowCrypt/Controllers/Compose/ComposeDecorator.swift
Kharchevskyi: `Reply from iOS` is not needed? It was in task description. Removed

tomholub: That was supposed to be example of a message reply that the user is sending.

If I receive a message `This was the original message\nNow it is quoted`, and I type type `reply from iOS`, then the sent message should be exactly as defined in the issue:
```
reply from iOS

On 2019-11-21 at 12:11, mindoutofframe@gmail.com wrote:
> This was the original message
> Now it is quoted
```

Issue #202, opened 2020-02-14
error: requested folder does not exist (probably when creating Realm?)


tomholub:
> Hi there, Have installed the FlowCrypt app for Ios but am having problems getting started. After selecting my Gmail account as the account I want to use, I then get invited to enter my pass phrase (although the screen also shows a loading icon). When I enter my pass phrase I eventually get the error screen below. 

![image](https://user-images.githubusercontent.com/6306961/74566735-1ea19b00-4faf-11ea-8e2a-b620175506a3.png)

tomholub:
This was on 0.1.6

tomholub:
iOS 13.3.1 on an iPhone XR.

tomholub:
There is one more user with this exact issue, so far 2 users reported it total.

apbristow:
Still happening with 0.1.7.  Possible contributing factors:  I have a Multiword passphrase?  I have uploaded more than one key in desktop version?  

Will keep trying. 

Andy

tomholub:
@apbristow None of the things you mention should cause issues, but thanks for mentioning it! It will probably be some combination of device / system version or other system settings.

tomholub:
@Kharchevskyi this is an important issue, please take a look soon

PR #215, opened 2020-03-08
Attempt to identify and specify an error on backup request

Kharchevskyi:
https://github.com/FlowCrypt/flowcrypt-ios/issues/202


Kharchevskyi:
Double-checked it now. Yes, it fallback to alert with error message where you can choose between retrying or choosing another account

PR #229, opened 2020-04-22
Evaluate recipients

### LINKED ISSUE #201: evaluate and color recipients right after entered
Right now, we ignore if recipients have a public key or not until the message send button is tapped.

Instead I would like to evaluate recipients right after they are entered, like Android and Web Extensions do.

This will cause each recipient to be colored as:
 - no background color, when loading (initial)
 - green for pgp key found
 - gray for not found
 - red for a network error (with a way to retry)
 - red for other error (not a valid email)

Kharchevskyi:
close #201


tomholub:
- [ ] test no pgp
 - [ ] test loading state
 - [ ] test has pgp
 - [ ] test network error

tomholub:
While testing, I encountered an error. The backend was very slow to respond (60 seconds) and then returned 500 in one case and 504 in other case when searching `tom@flowcrypt.com`.

In these cases it should show in red, which one time it did but one time it showed gray as if tom@flowcrypt.com didn't have any public key. There may be an error somewhere in handling this situation, it should never show email as having no pgp key if there was an error loading.

It's not too bad, so I will make a separate issue and meanwhile merge this as is.

PR #231, opened 2020-05-15
Fix for fonts in labels

### LINKED ISSUE #222: Font with paragraph style - sans serif
When adding a paragraph style to the attributed string it changes the font style for thin weight fonts

```
struct SignInViewDecorator: SignInViewDecoratorType {
    var description: NSAttributedString {
        "sign_in_description"
            .localized
            .attributed(
                .thin(17),
                color: UIColor.colorFor(darkStyle: .mainTextColor, lightStyle: .textColor),
                alignment: .center //  **change font style**
        )
    }
```

Kharchevskyi:
close #222 
IMG_4294


tomholub:
Can you please have a look at the tests? They are failing lately, but master looks ok

Kharchevskyi:
> Can you please have a look at the tests? They are failing lately, but master looks ok
It looks like Semaphore use Xcode 11.4 
https://docs.semaphoreci.com/examples/ios-continuous-integration-with-xcode/

and probably the problem with `libgmp-6.1.2-iPhoneOS-arm64`
https://forums.developer.apple.com/thread/117346#364519

Can we replace this library with any other? For what do we use it?

Kharchevskyi:
This problem can be solved by using another library instead of `libgmp-6.1.2-iPhoneOS-arm64`.
Or I can try to solve it with carthage to add it dynamically which will copy framework to the application during build.
https://forums.developer.apple.com/thread/128797
Or as a temp solution we can try to switch to 11.3.1 on CI

PR #237, opened 2020-06-04
Replace Bignum. Run build on Xcode 11.4

### LINKED ISSUE #235: Build on Xcode 11.4
> Can you please have a look at the tests? They are failing lately, but master looks ok

It looks like Semaphore use Xcode 11.4 
https://docs.semaphoreci.com/examples/ios-continuous-integration-with-xcode/

and probably the problem with `libgmp-6.1.2-iPhoneOS-arm64`
https://forums.developer.apple.com/thread/117346#364519

Can we replace this library with any other? For what do we use it?

_Originally posted by @Kharchevskyi in https://github.com/FlowCrypt/flowcrypt-ios/pull/231#issuecomment-629824622_

Kharchevskyi:
resolve #235


FlowCrypt/Core/CoreHost.swift
```diff
@@ -54,14 +55,15 @@ final class CoreHost: NSObject, CoreHostExports {
         return Cryptor(operation: .decrypt, algorithm: .aes, mode: .CFB, padding: .NoPadding, key: key, iv: iv).update(byteArray: ct)!.final()!
     }
 
-    // rsa verify is used by OpenPGP.js during decryption as well to figure out our own key preferences
-    // this slows down decryption the first time a private key is used in a session because bn.js is slow
```
tomholub, line 58: Please don't remove these two lines, they are still true

FlowCrypt/Core/CoreHost.swift
```diff
@@ -54,14 +55,15 @@ final class CoreHost: NSObject, CoreHostExports {
...
-    // Using GMP C library modular exponentiation reduces rsa4096 verify time from 800ms to 40ms
```
tomholub, line 59: Please measure how long does it take to decrypt a message (first time, and second time) when this new library is in place, and compare that to speed of decrypting a message when you return an empty string here (in which case bn.js will have to compute it, likely slower).

Measure each thing twice because there is some caching going on.

Then please update this comment.

FlowCrypt/Core/CoreHost.swift
```diff
@@ -54,14 +55,15 @@ final class CoreHost: NSObject, CoreHostExports {
...
     func verifyRsaModPow(_ base: String, _ exponent: String, _ modulo: String) -> String {
-        // only supported on arm64 because was not able to build it for other platforms yet
-        // in fact, I'm no longer able to build any working library other than the one I've built before
-        // when not supported, the function returns empty string, and JS falls back on bn.js
```
tomholub, line 63: This comment is still important (last line), but it could start with "If there is an error parsing provided numbers"

FlowCrypt/Core/CoreHost.swift
```diff
@@ -147,3 +149,29 @@ extension SecPadding {
     // https://developer.apple.com/documentation/security/secpadding/ksecpaddingnone
     public static let NONE = SecPadding(rawValue: 0)
 }
+
+
+private func modPow<T: BinaryInteger>(n: T, e: T, m: T) -> T {
```
tomholub, line 154: Where is this from?

Kharchevskyi: https://rosettacode.org/wiki/Modular_exponentiation#Swift
Will also add a comment there

tomholub: I'm very hesitant to use a code grabbed from a wiki for modular exponentiation. It does look correct, but I'm not a cryptographer. There's a chance of side channel leaks, like timing leaks. On the other hand, in the context of our app, that should not be a problem.

Given that this is a beta, I'll merge it. We'll have to have a cryptographer have a deeper look at this before we move to prod.

Kharchevskyi: Should I leave some extra comments there to not to forget this before moving to prod?

tomholub: For sure. Plus please make an issue, and add the issue link in the code comment. Also, please move it into a separate file. Thanks!

Kharchevskyi: https://github.com/FlowCrypt/flowcrypt-ios/issues/238

FlowCrypt/Core/CoreHost.swift
Kharchevskyi: Decrypting measurements: 
**First message**
1st run - 783 ms
2nd run - 251 ms

_with empty string_ 
1st run - 314 ms
2nd run - 138 ms

**Second message**
1st run - 510 ms
2nd run - 73 ms

_with empty string_ 
1st run - 318 ms
2nd run - 297 ms

tomholub:
Also there's this in tests:

image

Kharchevskyi:
Is it always like that? can it be some migration issues between different branches? 
what OS you run on iPOD?

Kharchevskyi:
@tomholub did some investigation, and it looks like iPOD is 32 bit device. This could be related to this issue

tomholub:
I think I have a 64 bit one, I'll check.

This issue only happens on the recent branches, I can build and run master just fine.

tomholub:
Correction, I can build and run code from last released version just fine. I can try master.

PR #242, opened 2020-07-07
Fix for evaluation of recipients

### LINKED ISSUE #241: double check + add tests for situation when attester returns 500 and 504
While testing, I encountered an error. The backend was very slow to respond (60 seconds) and then returned 500 in one case and 504 in other case when searching `tom@flowcrypt.com`.

In these cases it should show in red, which one time it did but one time it showed gray as if tom@flowcrypt.com didn't have any public key. There may be an error somewhere in handling this situation, it should never show email as having no pgp key if there was an error loading.

It's not too bad, so I will make a separate issue and meanwhile merge this as is.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/229#issuecomment-654308311_

Kharchevskyi:
close #241 

I was able to reproduce #241 once, and as far as I found the possible problem was in case where recipient was already in a list of recipients. This pr should fix this issue. Recipient which was already in a list will not be added once more, but instead will be checked once more.


FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -653,6 +666,7 @@ extension ComposeViewController {
         attesterApi.lookupEmail(email: recipient.email)
             .then(on: .main) { [weak self] result in
                 guard let self = self else { return }
+                print("\(result.armored == nil)")
```
tomholub, line 669: pls remove the print

PR #243, opened 2020-07-08
Fix for deleting messages in outlook account

### LINKED ISSUE #240: outlook: deleting emails doesn't work
1) new account
2) create a new backup during setup
3) on ios open the backup email -> move to trash
4) log out
5) log in, it will say "1 backup found". This is not wanted, the app should not search for backups in Trash folder.

Kharchevskyi:
close #240


FlowCrypt/Controllers/Msg/MessageViewController.swift
```diff
@@ -256,13 +256,13 @@ extension MessageViewController {
 
     private func permanentlyDelete() {
         guard let input = input else { return hideSpinner() }
-        input.objMessage.flags = MCOMessageFlag(rawValue: input.objMessage.flags.rawValue | MCOMessageFlag.deleted.rawValue)
+        input.objMessage.flags = MCOMessageFlag(rawValue: MCOMessageFlag.deleted.rawValue)
```
tomholub, line 259: We're doing this two times - here and just below - that seams wrong. I would delete this line, what do you think?

Kharchevskyi: I removed this line because it's quite strange to have 
`input.objMessage.flags.rawValue | MCOMessageFlag.deleted.rawValue` so it means it will have original flag **OR** deleted flag, but leave in previous implementation to not break anything. 
I can take a look in this functionality, but would be nice to have some ACs for this. Can you please provide some "expected behaviour" for this?

tomholub: I think this is a binary (bitwise) OR. That's different than a logical OR. It's working on a bit by bit basis.

For example, if you have a bitmask for a one byte value:

```
first bit: delete
second bit: important
third bit: ..
```
and then you have an existing binary value (one byte), with various flags:
```
0100101
```
And the delete flag is
```
1000000
```
Then ORing them together produces
```
1100101
```
Meaning it changed the first bit representing the delete. This is just an example, look up the docs and also how bitwise or works.

That itself looks correct. What I meant is, that this line is here two times. Why change it two times?

image

What's why I said the first changed line is probably there by error in the first place.

tomholub:
This looks good to merge except it's missing a test. If you need the secrets file updated let me know.

Kharchevskyi:
I messed a bit with this secrets file. It should be in a root directory but I changed it to another one, so CI couldn't find it. Reverted my changes for it back
BTW I think we are not running UI tests on CI yet.

tomholub:
> I messed a bit with this secrets file. It should be in a root directory but I changed it to another one, so CI couldn't find it. Reverted my changes for it back
> BTW I think we are not running UI tests on CI yet.

Got it, we'll look into that separately. I have a fix in mind for the frequently failing Gmail login issues, once I have the fix ready, I'll share and we should be able to enable ui tests on the CI. Also, meanwhile, it would be good to already enable UI tests of IMAP/SMTP where this is not a problem. Unless there was some other issue when running UI tests?

Kharchevskyi:
I think I just need a bit to update them, but when I was working on IMAP/SMTP everything works fine, except I have some warnings that outlook can identify it as unusual activity which should be confirmed

_

> We detected something unusual about a recent sign-in for the Microsoft account fl*****@outlook.com. For example, you might be signing in from a new location, device, or app. To help keep you safe, we've blocked access to your inbox, contacts list, and calendar for that sign-in. Please review your recent activity and we'll help you secure your account. To regain access, you'll need to confirm that the recent activity was yours.

_

would be nice to find not so strict email provider which allows IMAP/SMTP connection and use this one in tests, but I will try it with Outlook account on CI as well

tomholub:
> would be nice to find not so strict email provider which allows IMAP/SMTP connection and use this one in tests, but I will try it with Outlook account on CI as well

Agree, better use some other email provider.

Best would be set up our own fake SMTP/IMAP server during CI. That's what we will be doing for Android soon, then we could copy the solution here. Until than, use some unrestrictive email provider.

PR #244, opened 2020-07-08
Submit keys to attester

### LINKED ISSUE #224: submit keys to attester with id_token
When submitting keys to attester, and logged in with Gmail, use a different method to submit public key: change PUT to POST and add a header: `Authorization: Bearer eyXXXXXXXXX` where `eyXXXXXXXXX` is an OpenID Token that you receive during Google authentication. 

Still keep the previous submit PUT method available (without id_token), and use it or IMAP/SMTP because that flow will not allow you to grab an id token.

By using id_token, you prevent this:

image

Because this will allow the backend to verify to request and replace the public key even if there is a conflicting pubkey available.

Kharchevskyi:
close #224


Kharchevskyi:
Simulator Screen Shot - iPhone SE (2nd generation) - 2020-07-08 at 22 25 38
Still have this error when submitting key with outlook account. 
With Google account - works without any errors

tomholub:
If you do https://flowcrypt.com/attester/update-public-keys?intention=delete for that outlook email and then go through setup again, does it work without error?

If yes then all is good.

Kharchevskyi:
just tried for `flowcrypt.compatibility.ios.nokey@outlook.com` - works without an error

PR #245, opened 2020-07-08
Import key from settings

### LINKED ISSUE #198: add a way to import more private keys
a (+) button in `FlowCrypt Settings -> Keys` screen that lets me import from file or from clipboard

Kharchevskyi:
close #198


FlowCrypt/Functionality/Services/ErrorHandler.swift
```diff
@@ -0,0 +1,22 @@
...
+//  FlowCrypt
+//
+//  Created by Anton Kharchevskyi on 20/07/2020.
+//  Copyright © 2020 FlowCrypt Limited. All rights reserved.
+//
+
+import UIKit
+
+extension UIViewController {
+    func handleCommon(error: Error) {
```
tomholub, line 12: Always happy to see explicit error handling, thanks

FlowCrypt/Controllers/ImportKey/Enter Pass Phrase/EnterPassPhraseViewController.swift
```diff
@@ -178,23 +181,43 @@ extension EnterPassPhraseViewController {
...
+        switch keyService.retrieveKeyDetails() {
+        case let .failure(error):
+            handleCommon(error: error)
+        case let .success(existedKeys):
+            checkIfKeyPossibleToAdd(with: existedKeys, and: passPhrase)
+        }
+    }
+
+    private func checkIfKeyPossibleToAdd(with existedKeys: [KeyDetails], and passPhrase: String) {
+        let isKeyAlreadyAdded = Set(fetchedKeys).isSubset(of: existedKeys)
```
tomholub, line 203: Sorry - while reading this source code, I got a different idea.

Instead of showing errors, when you encounter a private key with the same fingerprint, we would **replace** the stored key object instead of skipping it. So we would be doing an upsert, effectively.

So if the user imported 5 keys that were all the same, that would be fine - we'd process one by one and then finally the last one would end up stored in the database in the end.

Finally, we would show an alert.

1) if there were only new keys, no replacement: `Keys imported: <N>`
2) if there were only updates: `Keys imported: 0\nKeys updated: <N>`
3) if there was a mix of new keys and updates: `Keys imported: <N>\nKeys updated: <Y>`

We could show such alert regardless if this is first time setup or if adding keys from settings, either way it's good to know.

All of this will be achieved by comparing first fingerprint in each key, as you did here.

Thanks!

FlowCrypt/Controllers/ImportKey/Enter Pass Phrase/EnterPassPhraseViewController.swift
```diff
@@ -200,25 +203,47 @@ extension EnterPassPhraseViewController {
...
-            return
-        }
+        let keysToUpdate = Array(Set(existedKeys).intersection(fetchedKeys))
+        let newKeysToAdd = Array(Set(fetchedKeys).subtracting(existedKeys))
 
         keysDataService.addKeys(
-            keyDetails: fetchedKeys,
+            keyDetails: newKeysToAdd,
             passPhrase: passPhrase,
             source: .generated
```
tomholub, line 212: This is wrong - here the source of the key is imported, not generated

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -169,6 +170,25 @@ extension EncryptedStorage {
...
     }
 
+    func updateKeys(keyDetails: [KeyDetails], passPhrase: String, source: KeySource) {
+        // KeyInfo doesn't have primaty key, to avoid migration we need to delete keys and then save them
+
+        // delete keys
+        keyDetails.forEach { keyDetail in
+            try? storage.write {
+                storage.delete(storage.objects(KeyInfo.self)
+                    .filter("longid=%@", keyDetail.ids[0].longid))
```
tomholub, line 180: We should start saving fingerprints in the storage (in addition to longids), and compare by fingerprints wherever possible instead of longids. It's safer - collisions less likely. I'll make another issue for this.

FlowCrypt/Controllers/ImportKey/Enter Pass Phrase/EnterPassPhraseViewController.swift
Kharchevskyi: Do I need to introduce a new case for it? 
Now we have 
```swift
enum KeySource {
    case backup
    case generated

    var path: String {
        switch self {
        case .backup: return "backup"
        case .generated: return "generated"
        }
    }
}
```

tomholub: I think that's a good idea. If the key was auto-loaded from backups, source should be "backup". If it was manually imported, then "imported" and if it was generated then "generated".

tomholub:
Functionality is good, I can merge this once conflicts are resolved.

PR #252, opened 2020-08-21
Local public key database

### LINKED ISSUE #199: Contacts: local public key database
Currently, when evaluating a recipient, we fetch public key from Attester each time we email a recipient. But we should instead have a mechanism to track public keys locally as Contacts.

The workflow is as follows:
1) when a recipient is evaluated (see issue #201 ), first search local contacts for public key
2) if found, use the local contact
3) if not found, search on attester
4) if pubkey found on attester, save the public key in local contacts (TOFU - Trust on First Use)

Either way, later when I hit the send button, the public key will be fetched locally from Contacts. If there are none, it will show alert that recipient doesn't use pgp. 

When saving contacts, the following info needs to be saved, as a Realm object:
```js
type Contact = {
  email: string; // lowercased, trimmed email address
  name: string | null; // name if known, else null
  searchIndex: string; // this will be used to search contacts, and will be: `\(email) \(name)` lowercased. You can skip this if Realm can do a composed query such as `WHERE email CONTAINS "tom" OR name CONTAINS "tom"`.
  pubkey: string | null; // armored public key, or null if we don't know pubkey
  pubkeyLastSig: Date  | null; // todo for tom - I'll give you a way to get this
  pubkeyLastChecked: number | null; // the date when pubkey was retrieved from Attester, or null if none
  pubkeyExpiresOn: number | null; // pubkey expiration date
  longid: string | null; // first pubkey longid
  longids: string | null; // all pubkey longids, comma-separated
  lastUsed: Date | null; // last time an email was sent to this contact, update when email is sent (will be used to order contact results)
};
```
For now, we only support sending emails to users who DO have a public key. Therefore, we'll only store a contact if we already know their public key.

Kharchevskyi:
close #199


tomholub:
Screenshots look good. I'll have a look. Thanks!

tomholub (approved):

tomholub (approved):

tomholub (approved):

PR #270, opened 2021-02-16
Switching accounts

Kharchevskyi:


tomholub:
Thank you for the screenshots

Kharchevskyi:
Simulator Screen Shot - iPhone 12 mini - 2021-02-22 at 17 34 34

Kharchevskyi:
Simulator Screen Shot - iPhone 12 mini - 2021-02-22 at 17 34 34

Kharchevskyi:
@tomholub Looks like it's not possible to switch between accounts with `GIDSignIn`.
Probably need to add one more dependency https://github.com/google/GTMAppAuth.

tomholub:
Is it because the original library is trying to bee too clever, and assumes just one account? Does it do it's own token storage - it's not possible to plug our own storage, which would give a different account when user switches?

Kharchevskyi:
Looks like GoogleSignIn is a shared instance and `signIn`, `renewSession` and `signOut` could be done only for one user. 
So my idea is to use https://github.com/google/GTMAppAuth to renew token when user switch account and store this token in our DB. I will investigate a bit more and will let you know

tomholub:
ok

tomholub:
Are you still running into the same issue?

Kharchevskyi:
I spent some time last week to investigate it, but  I couldn't find a way how to use GoogleSignIn with multiple accounts. As far as I see it doesn't supports multiple accounts. To be able to use multiple accounts and to switch between them we need to use https://github.com/google/GTMAppAuth

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -654,7 +624,7 @@ extension ComposeViewController {
 // MARK: - Action Handling
 extension ComposeViewController {
     private func searchEmail(with query: String) {
-        googleService.searchContacts(query: query)
+        contactsProvider.searchContacts(query: query)
```
tomholub, line 627: Is this searching local contacts or external / google contacts? We may want to distinguish for clarity, something like:

`LocalContactsProvider` vs `CloudContactsProvider` ?

FlowCrypt/Controllers/ImportKey/Enter Pass Phrase/EnterPassPhraseViewController.swift
```diff
@@ -251,6 +251,7 @@ extension EnterPassPhraseViewController {
     }
 
     private func moveToMainFlow() {
+        // TODO: - ANTON - enter pass phrase
```
tomholub, line 254: Is this still to be done?

FlowCrypt/Controllers/Msg/MessageViewController.swift
```diff
@@ -119,7 +119,9 @@ final class MessageViewController: TableNodeViewController {
             // in case we are in trash folder ([Gmail]/Trash or Deleted for Outlook, etc)
             // we need to have only help and trash buttons
             items = [helpButton, trashButton]
-        case MailDestination.Gmail.inbox.path.lowercased():
+
+        // TODO: - ANTON - Check if this should be fixed
```
tomholub, line 123: Just highlighting

FlowCrypt/Controllers/Setup/SetupViewController.swift
```diff
@@ -396,17 +396,12 @@ extension SetupViewController {
...
     private func handleOtherAccount() {
-        userService.signOut()
-            .then(on: .main) { [weak self] _ in
-                self?.router.proceed()
-            }
-            .catch(on: .main) { [weak self] error in
-                self?.showAlert(error: error, message: "Could not switch accounts")
-            }
+        // TODO: - ANTON - Setup - handleOtherAccount
+        router.signOut()
```
tomholub, line 400: Highlighting

FlowCrypt/Controllers/SideMenu/Menu/MyMenuViewController.swift
```diff
@@ -115,6 +185,7 @@ final class MyMenuViewController: ASDKViewController<ASDisplayNode> {
         tableNode.reloadData()
     }
 
+    // TODO: - ANTON - Handle error
```
tomholub, line 188: Various todos in this file - thank you for marking them clearly :+1:

FlowCrypt/Controllers/SignIn Other/EmailProviderViewController.swift
```diff
@@ -489,23 +492,19 @@ extension EmailProviderViewController {
...
     private func handleConnection(error: Error) {
-        imap.disconnect()
-        dataService.logOutAndDestroyStorage()
-        DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
-            let message = (error as? AppErr)?.userMessage ?? error.localizedDescription
-            self.hideSpinner()
-            self.showToast(message)
-        }
+        globalRouter.signOut()
+        // TODO: - ANTON email provider - logout
```
tomholub, line 496: +1

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -74,8 +82,23 @@ final class EncryptedStorage: EncryptedStorageType {
...
+        let users = storage.objects(UserObject.self)
+
+        // in case there is only one user - just delete storage
+        if users.count == 1, users.first?.email == email {
+            destroyEncryptedStorage()
+        } else {
+            // remove user and keys for this user
+            let userToDelete = users.filter { $0.email == email }
+            let keys = storage.objects(KeyInfo.self).filter { $0.account == email }
+            let folders = storage.objects(FolderObject.self)
```
tomholub, line 95: Are folder rows also differentiated by users? Or are they getting shared among users?

FlowCrypt/Functionality/Migration/DBMigrationService.swift
```diff
@@ -0,0 +1,109 @@
...
+
+import Foundation
+import Promises
+import RealmSwift
+
+protocol DBMigration {
+    func performMigrationIfNeeded() -> Promise<Void>
+}
+
+struct DBMigrationService {
```
tomholub, line 17: We could probably remove all previous migrations - after we release 0.2.0 (it's been a while, and not many users since it was in beta and then expired for a while). I'll make another issue for later.

tomholub (commented):
Doing a checkpoint review so that later I only have to review the diff. Thanks!

FlowCrypt/Controllers/ImportKey/Enter Pass Phrase/EnterPassPhraseViewController.swift
Kharchevskyi: This all are just markers for me to no to forget to resolve before PR will be ready

FlowCrypt/Controllers/SideMenu/Menu/MyMenuViewController.swift
Kharchevskyi: will make them more readable next time)

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
Kharchevskyi: this is folders for an account, like inbox, snoozed, somePersonalFolderName

FlowCrypt/Functionality/Migration/DBMigrationService.swift
Kharchevskyi: great point 👍

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
tomholub: How do you prevent folders from one account from mixing with folders from another account?

For example, for keys you do `.filter { $0.account == email }` but for folders you pull all. Or is there also an `account` column on the folders table?

Kharchevskyi: For now we do not associate folders with a user. 
We fetch and update folders in storage on every time user opens a side menu. 
I still need to associate this list of folders to a user.

tomholub: Understood

Kharchevskyi: Fixed in latest commit

Kharchevskyi:
@tomholub will mark this PR as "Ready for review", meanwhile will do some small ui updates and will start testing.
is it possible for you to test this as well before merge? 
Please remove old app and install this build (didn't perform all DB migration due to https://github.com/FlowCrypt/flowcrypt-ios/issues/284)

tomholub (commented):
Looks good - haven't tried running the app yet

tomholub (commented):
I ran into an issue when testing this:

1) wipe app, install new one
2) log in with flowcrypt.compatibility account
3) send an email, open some emails, delete some emails
4) click the chevron next to account email 
5) clicked add another account
6) clicked `Continue with Gmail`
7) got prompt to sign in with email (the "alert" type of pop up)
8) clicked continue and the alert disappeared, but stayed on the same screen
9) (I think) I tried again and got another alert, and clicked continue again, stuck on the same sreen, app crashed

image

Restarting the app, then going straight to step 5) to 9) can still reproduce it.

Kharchevskyi:
Thanks @tomholub will check this issue ASAP

Kharchevskyi:
what iOS version do you have? 
can you please enable exception breakpoint and try to reproduce this crash once more?
Screenshot 2021-04-21 at 22 02 26

tomholub:
Here, although it looks about the same. Did I do it right? I'm on an ipod running 12.5.2 (13 not available)

image

image

tomholub:
Another thing I noticed (but can be fixed in another PR) - clicking on the chevron to show more accounts the first time is quite slow (even a few seconds). Causes me to click again, at which point the "add account" options shows and disappears quickly. Followup toggles are fast. Is that a network request?

tomholub:
Just tried on an iphone 6, same iOS version - also reproducible (but I could click `continue with gmail` and `continue` 3 or 4 times before it crashed. Maybe I clicked `Continue with Outlook` too)

tomholub:
Also ran into: https://github.com/FlowCrypt/flowcrypt-ios/issues/293 (iphone6, this time my own account)

Kharchevskyi:
Yes, you did it correctly, but unfortunately there is no stack trace for this exception. It can be smth related to OS version. Will try to reproduce this issue on simulator with lower iOS version

Kharchevskyi:
Reproduced on simulator running on iOS 12. Will continue investigating this

> clicking on the chevron to show more accounts the first time is quite slow (even a few seconds)

I think also related to iOS version. 
Will try to find some iPhone with old iOS version to test on a real device

tomholub:
Let's drop iOS 12. I'll get an iOS 13 device to test on. I'll go ahead and merge this as is, and then once I've tested on iOS 13, I'll file issues if I discover any.

Issue #193, opened 2020-02-14
ability to view list of attachments


tomholub:
part of  #275

When I open a message, currently it doesn't show a list of attachments on that message.

Please list attachments in message view, and their sizes. Downloading or decrypting doesn't need to be implemented yet (clicking each attachment will produce a toast "Downloading attachments not implemented yet".

 The `Core.parseDecryptMsg` returns an array of MsgBlock of various types, and one textual representation of the message body. Currently we only look at textual representation of the message body `text` but not actual individual blocks.

You'll want to look at `decrypted.blocks` to find the ones of the following types:  `plainAtt`, `encryptedAtt`, `decryptedAtt` and render them below. Use our Android app for rough visual guidance. Attachments that are encrypted (encryptedAtt, decryptedAtt) should be styled in green similar to how encrypted text is green. `plainAtt` should be styled in a neutral color.

When opening a (draft) PR please add a screenshot to the description as well.

The functionality should be accompanied with meaningful tests - for guidance you can ask me (general approach) and @Kharchevskyi (specific code questions).

qsdemir:
@tomholub @Kharchevskyi 

`Core.parseDecryptMsg` returns an array of MsgBlock, but that array always consists of only one element.  

`decrypted.blocks` always has one block, and its type is `plainHtml`, which means that blocks of `plainAtt`, `encryptedAtt`,  `decryptedAtt` types don't exist. This might be an issue with js code, most likely `parseDecryptMsg`.

Can you please take a look into this? I didn't want to fiddle around js, and Core too much, because I didn't want to break something else unknowingly.

tomholub:
Hello, that should be the case if source message contains no attachments. But if there are any attachments, they should show up as `MsgBlock`s. 

I'll share a test account with you (over email) where you will find a few sample messages in various formats to test with to confirm.

tomholub:
@qsdemir I've sent you email with credentials to access a test account.

There, please test the following messages:
 - `file preview - RICH TEXT PGP/MIME - open txt file, image, pdf, unknown`
 - `file preview - open txt file, image, pdf, unknown`
 - `[enigmail] signed message inline + attachment`
 - `[enigmail] signed message + attachment + PGP/MIME`

There are various types of attachments represented there:
1) plain
2) encrypted separately from message ("inline")
3) encrypted together with message ("pgp/mime")
4) inside pgp/mime signed message

[...]

PR #295, opened 2021-04-28
issue #193 add attachment nodes

qsoftdevelopment:
This is a draft PR to check whether or not we're on a good path. Please note that most issues come with using Texture, since we're unfamiliar with it, so here you can find two attachments hardcoded and shown in every email just to check if nodes and this approach are correct.


tomholub:
@Kharchevskyi please have a look at the rendering parts. Thanks!

FlowCrypt/Controllers/Msg/MessageViewController.swift
```diff
@@ -377,6 +384,8 @@ extension MessageViewController: ASTableDelegate, ASTableDataSource {
                 return MessageSubjectNode(subject, time: time)
             case .text:
                 return MessageTextSubjectNode(self.message)
+            case .attachment:
+                return AttachmentsNode.init(input: [input, input2])
```
Kharchevskyi, line 388: please omit `.init` here to be consistent with all code base

FlowCryptUI/Nodes/AttachmentNode.swift
```diff
@@ -0,0 +1,99 @@
...
+//  FlowCryptUI
+//
+//  Created by QSD BiH on 16. 4. 2021..
+//  Copyright © 2021 FlowCrypt Limited. All rights reserved.
+//
+
+import AsyncDisplayKit
+
+public struct Attachment {
+    var name, size: NSAttributedString
```
Kharchevskyi, line 12: in case this struct is public you will need to create public initialiser

FlowCryptUI/Nodes/AttachmentNode.swift
```diff
@@ -0,0 +1,99 @@
...
+    private var attachmentNodes: [AttachmentNode] = []
+
+    public init(input: [Input]) {
+        super.init()
+        input.forEach { input in
+            attachmentNodes.append(
+                AttachmentNode(
+                    input: AttachmentNode.Input(name: input.name, size: input.size)
+                )
+            )
```
Kharchevskyi, line 38: instead of using `forEach` you can use `map`
`attachmentNodes = input.map { AttachmentNode.Input(name: $0, size: $0.size) }`

or you can have private extended init for AttachmentNode.Input which can be inited with `Input`
In this case you will have 
`attachmentNodes = input.map(AttachmentNode.Input.init)`

FlowCryptUI/Nodes/AttachmentNode.swift
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    public override func layoutSpecThatFits(_: ASSizeRange) -> ASLayoutSpec {
+        return ASInsetLayoutSpec(
+            insets: UIEdgeInsets(top: 8, left: 0, bottom: 8, right: 0),
```
Kharchevskyi, line 44: should right inset be 0? 
can you please add a screenshot of how this should look like

FlowCryptUI/Nodes/AttachmentNode.swift
```diff
@@ -0,0 +1,99 @@
...
+
+    private let titleNode = ASTextNode()
+    private let subtitleNode = ASTextNode2()
+    private let imageNode = ASImageNode()
+    private let buttonNode = ASButtonNode()
+    private let separatorNode = ASDisplayNode()
+
+    public init(input: Input) {
+        super.init()
+        self.borderWidth = 0.5
```
Kharchevskyi, line 67: are you sure about 0.5 border width?

FlowCryptUI/Nodes/AttachmentNode.swift
```diff
@@ -0,0 +1,99 @@
...
+        self.borderColor = UIColor(named: "red")?.cgColor
```
Kharchevskyi, line 68: UIColor.red

FlowCryptUI/Nodes/AttachmentNode.swift
```diff
@@ -0,0 +1,99 @@
...
+
+        imageNode.image = UIImage(named: "paperclip")
+        buttonNode.setImage(UIImage(named: "paperclip"), for: .normal)
+        titleNode.attributedText = NSAttributedString.text(from: input.name, style: .medium(16))
+        subtitleNode.attributedText = NSAttributedString.text(from: input.size, style: .medium(12))
```
Kharchevskyi, line 73: 👍

FlowCrypt/Core/CoreTypes.swift
```diff
@@ -19,7 +19,7 @@ struct CoreRes {
         let text: String
     }
 
-    struct ParseDecryptMsg {
+    struct ParseDecryptMsg: Decodable {
```
Kharchevskyi, line 22: why this should be Decodable?

FlowCrypt/Controllers/Msg/MessageViewController.swift
```diff
@@ -14,7 +14,7 @@ final class MessageViewController: TableNodeViewController {
     }
 
     enum Parts: Int, CaseIterable {
-        case sender, subject, text
+        case sender, subject, text, attachment
```
Kharchevskyi, line 17: are all attachments goes under the text by design? 

can it be the case that text will go after an attachment?
text
attachment
text

?

tomholub: Right now, we'll not worry about messages that have attachments in other positions - we'll simply render them always at the end. Later (~ year 2023) we can revisit that.

Kharchevskyi:
In general looks good 👍

FlowCryptUI/Nodes/AttachmentNode.swift
qsdemir: @Kharchevskyi Here's what it looks like on Android:
Screen Shot 2021-04-30 at 10 28 16

And currently on iOS:
Screen Shot 2021-04-30 at 10 33 58



I tried to match this as closely as  I could. 

Also, paperclip image is just a sample since I don't have the right ones. @tomholub Can you send them?

tomholub: We don't need to match THAT closely between Android and iOS. It should be roughly aligned in the sense that there are attachments on the bottom, etc, but exact icons and exact borders etc should rather follow platform habits. For example for iOS download button, I'd use some icon from iOS palette that is available for that purpose, rather than Android icon. (sorry for not clarifying earlier)

tomholub: In other words, when I open the android app, it should look like and android app. When I open iOS app, it should look like an iOS app. When I put the two next to each other, they look similar and clearly from the same company, but each their own "flavor" following the platform.

qsdemir: Got it, thanks

Kharchevskyi: Great then 🔥
@qsoftdevelopment can you please make border width to be 1 point(0.5 will highly drop performance on drawing). also please make left and right offset to be 8 points and please make round 8 corner radius to follow App Designn Guidelines. And also use more iOS icons

tomholub:
I should also clarify: as long as the visual doesn't look broken, some ugliness is OK at this stage. Functionality is a lot more important at this stage then looks. There will be a time that we refurbish the app for looks sometime in 2022 but this year we won't have time for that, so a few rough edges in UI are ok. This also applies to PR reviews - the first focus is functionality and correctness (and usability), and visual appeal is secondary.

tomholub:
@Kharchevskyi how do you propose this type of functionality should be tested?

Kharchevskyi:
As I understand there is only UI added in this PR. 
I think it should be tested when "parsing" functionality will be in place

tomholub:
> As I understand there is only UI added in this PR.
> I think it should be tested when "parsing" functionality will be in place

Definitely - once functionality is in place, what should be the way to test this? With you answer, we should probably update readme to document a clear overview of our approach to testing.

Kharchevskyi:
@tomholub Testing is quite specific task by task. For this task I think expected result is that Attachment block is present in messages where they are. They coloured in correct colour. Toast is shown when tapping on attachment. 
I think better to have some "expected result" when describing GH issue.

tomholub:
> @tomholub Testing is quite specific task by task. For this task I think expected result is that Attachment block is present in messages where they are. They coloured in correct colour. Toast is shown when tapping on attachment.
> I think better to have some "expected result" when describing GH issue.

Yes, but more specifically - are we pulling messages from Gmail API to test this? Or do we mock it? Or a unit test? This is what we need to make some standardization and guidance for, else contributors will not know what type of test to write and how.

Kharchevskyi:
Yes, we use Gmail API to test functionality. 
I do remember it was a plan to mock some Mail API, is it already implemented? Can we use it on iOS?

Unit test - good question. We do not have a lot of unit tests now. But I do think we need to work on it and add some scope in each PR for writing tests. Or even cover some previous implementation with tests

tomholub:
We don't have any mock API available for iOS testing yet. On Android, we test though IMAP where we have a mock server. That produces a pretty decent testing ergonomy. While it doesn't test the Gmail API vs IMAP/SMTP layer, it's pretty small difference.

That means for this test, we would test by opening an actual Gmail message, on live Gmail servers like other tests, and testing through the UI - because that's about the only thing we are set up for, for now. correct?

Kharchevskyi:
Is it possible to somehow share this IMAP server?

Yes, for now all tasks like that can be tested only with real Gmail account. Usually I use accounts from `test-ci-secrets.json"

tomholub:
Let's skip tests on this PR, and we'll do manual testing for this one. Then we'll set up tests afterwards with mock IMAP server.

tomholub:
@qsoftdevelopment how is this coming along, could we get this in mergeable state soon?

tomholub:
@qsdemir how is this coming along, could we get this in mergeable state soon?

qsdemir:
@tomholub @Kharchevskyi Hi, can you please take a look into this again?

tomholub:
@qsdemir you can mark the PR ready for review if you think it's good for review / potential merging.

@seisvelas please build and run this to check the functionality

@Kharchevskyi please have a look at the code

FlowCrypt/Controllers/Msg/MessageViewController.swift
```diff
@@ -170,6 +172,10 @@ extension MessageViewController {
                 isEmail: true
             )
             let decryptErrBlocks = decrypted.blocks.filter { $0.decryptErr != nil }
+            let decryptAttBlocks = decrypted.blocks.filter { $0.type == .plainAtt || $0.type == .encryptedAtt || $0.type == .decryptedAtt }
```
Kharchevskyi, line 175: I would suggest to have some computed property for MsgBlock and cover it with tests(not in scope of this PR)
```
extension MsgBlock {
    var isAttributeBlock: Bool {
        type == .plainAtt || type == .encryptedAtt || type == .decryptedAtt
    }
}
```

FlowCrypt/Controllers/Msg/MessageViewController.swift
```diff
@@ -170,6 +172,10 @@ extension MessageViewController {
...
+
+            let attachments = decryptAttBlocks.map { Attachment(name: $0.attMeta?.name ?? "Attachment", size: $0.attMeta?.length ?? 0) }
```
Kharchevskyi, line 177: same for Attachment. You can have
``` 
extension Attachment {
    init(block: MsgBlock) {
    self.name = 
    self.size = 
    }
}
```
and then it can be just
```
let attachments = decrypted.blocks
       .filter(.isAttributeBlock)
       .map(Attachment.init)
```

this solution is more flexible and can be tested easily.

FlowCryptUI/Nodes/AttachmentNode.swift
```diff
@@ -0,0 +1,117 @@
...
+    private var onTap: (() -> Void)?
+
+    public init(attachments: [Attachment], onTap: (() -> Void)?) {
+        super.init()
+        self.onTap = onTap
+        attachmentNodes = attachments.map { AttachmentNode(input: AttachmentNode.Input(name: $0.name, size: $0.size),
+                                                           onTap: {
+                                                            self.onTap?()
+                                                           })
+                                                        }
```
Kharchevskyi, line 47: please change indentation here? hard for understanding 
```
attachmentNodes = attachments.map {
   input: AttachmentNode(
       input:
       onTap:
      )
   )
}
```

Kharchevskyi (changes requested):

Kharchevskyi:
@qsdemir good job 👍
Please fix some small comments and improvements

FlowCrypt/Controllers/Msg/MessageViewController.swift
tomholub: It would be called `isAttachmentBlock`. But this is rather minor as you say

FlowCryptUI/Nodes/AttachmentNode.swift
tomholub: Can xcode be set to indent automatically on save?

Kharchevskyi: We already have a script for that. Will check if it's possible to indent code during build phase. 
Will work on it ASAP to prevent all this "code style" comments

tomholub (approved):
I've tested the functionality, looks good. We can address the style and simplification review points made by Anton in another PR later.

PR #304, opened 2021-05-06
Ignore load account tap while searching for backups

### LINKED ISSUE #293: clicking Load Account while still loading causes crash
My backup screen was showing loader for a long time, enough to enter pass phrase and click `Load Account`. The click crashes the app.

Kharchevskyi:
close #293


FlowCrypt/Controllers/Setup/SetupViewController.swift
```diff
@@ -171,21 +179,7 @@ extension SetupViewController {
             }
     }
 
-    private func fetchEnctyptedKeys(with backupData: Data) {
```
tomholub, line 174: Was this method unused?

Kharchevskyi: this functionality moved to `BackupService` which is now fully responsible to fetch backups from provider(Map or Gmail) and parse keys. 
see `func fetchBackups(for userId: UserId) -> Promise<[KeyDetails]>` inside `BackupService`

just forgot to remove it from view controller

tomholub:
@Kharchevskyi can you please look at the tests?

tomholub:
I debugged the problem(s):

1) there are 5 search expressions - and it seems we make 5 API requests to gmail. This should not be the case, it should be just one search expression and one request with 4 ORs in it.
2) the search expressions are wrong for Gmail search. On my account I have no backups, but it downloads 137 emails! and then 73 attachments with 33mb of data
3) this above seems to take forever, which is why we got to this UI issue in the first place (that you fixed)
4) when it does finish, it was saying "Setup failed" which is a nonsensical error, it should have said "No backups found on this account" (I fixed)

Please address issues 1 and 2 as well

tomholub:
Change - please just fix tests on this PR, and we can address points 1 and 2 some other time, there are more important issues before that.

tomholub:
They seemed to have passed this time - flaky

PR #329, opened 2021-06-01
WKD public keys - for now not integrated

ekievsky:
This PR implements logic for retrieving wkd or attester public keys;

issue #296

**Tests**:
- Tests updated for WKDURLsConstructor
- Can't add tests for WKDURLsApi and AttesterApi because there is no networking entity that injects into these classes and could be spied. 
P.S. It would be great to refactor Networking call `URLSession.shared.call(_ urlRequest: URLRequest, tolerateStatus: [Int]? = nil) -> Promise<HttpRes>` for separate networking entity to test classes that make API calls.

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Services/AttesterApi.swift
+++ b/FlowCrypt/Functionality/Services/AttesterApi.swift
@@ -18,6 +18,11 @@ protocol AttesterApiType {
 }
 
 final class AttesterApi: AttesterApiType {
+
+    private enum Constants {
+        static let lookupEmailRequestTimeout: TimeInterval = 10
+    }
+
     private enum Endpoint {
         static let baseURL = "https://flowcrypt.com/attester/"
     }
@@ -45,6 +50,8 @@ extension AttesterApi {
             // programming error because should never happen
             throw AppErr.unexpected("Status \(res.status) when looking up pubkey for \(email)")
         }
+        .timeout(Constants.lookupEmailRequestTimeout)
+        .recoverFromTimeOut(result: PubkeySearchResult(email: email, armored: nil))
     }
 
     @discardableResult
--- a/FlowCrypt/Functionality/Services/AttesterApi.swift
+++ b/FlowCrypt/Functionality/Services/AttesterApi.swift
@@ -51,7 +51,6 @@ extension AttesterApi {
             throw AppErr.unexpected("Status \(res.status) when looking up pubkey for \(email)")
         }
         .timeout(Constants.lookupEmailRequestTimeout)
-        .recoverFromTimeOut(result: PubkeySearchResult(email: email, armored: nil))
     }
 
     @discardableResult
--- a/FlowCrypt/Functionality/Services/WKDURLsApi.swift
+++ b/FlowCrypt/Functionality/Services/WKDURLsApi.swift
@@ -1,0 +1,129 @@
+//
+//  WKDURLsService.swift
+//  FlowCrypt
+//
+//  Created by Yevhen Kyivskyi on 23.05.2021.
+//  Copyright © 2021 FlowCrypt Limited. All rights reserved.
+//
+
+import CryptoKit
+import Promises
+
+protocol WKDURLsApiType {
+    func lookupEmail(_ email: String) -> Promise<[String]>
+    func rawLookupEmail(_ email: String) -> Promise<CoreRes.ParseKeys?>
+}
+
+class WKDURLsApi: WKDURLsApiType {
+
+    private let wkdURLsConstructor: WKDURLsConstructorType
+    private let core: Core
+
+    init(
+        wkdURLsConstructor: WKDURLsConstructorType = WKDURLsConstructor(),
+        core: Core = Core.shared
+    ) {
+        self.wkdURLsConstructor = wkdURLsConstructor
+        self.core = core
+    }
+
+    func lookupEmail(_ email: String) -> Promise<[String]> {
+        Promise<[String]> { [weak self] resolve, _ in
+            guard let self = self else { return }
+            let response = try awaitPromise(self.rawLookupEmail(email))
+            guard let safeResponse = response else {
+                resolve([])
+                return
+            }
+            let pubKeys = safeResponse.keyDetails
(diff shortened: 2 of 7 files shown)


FlowCrypt/Functionality/Services/WKDURLsApi.swift
```diff
@@ -0,0 +1,129 @@
...
+
+    init(
+        wkdURLsConstructor: WKDURLsConstructorType = WKDURLsConstructor(),
+        core: Core = Core.shared
+    ) {
+        self.wkdURLsConstructor = wkdURLsConstructor
+        self.core = core
+    }
+
+    func lookupEmail(_ email: String) -> Promise<[String]> {
```
tomholub, line 30: Here we are getting the raw bytes, then parsing them, and then armoring them and returning armored result.

Somewhere down the line another code will de-armor, parse the result to actually use the keys.

Best would be to change the signature of returns of both AttesterApi and this class have a public signature only `lookupEmail(_ email: String) -> Promise<[KeyDetails]>` - just one method

FlowCrypt/Functionality/WKDURLs/PubLookup.swift
```diff
@@ -0,0 +1,54 @@
...
+        wkd: WKDURLsApiType = WKDURLsApi(),
+        attesterApi: AttesterApiType = AttesterApi(),
+        core: Core = .shared
+    ) {
+        self.wkd = wkd
+        self.attesterApi = attesterApi
+        self.core = core
+    }
+
+    func lookupEmail(_ email: String) -> Promise<[String]> {
```
tomholub, line 30: And also here - return `Promise<[KeyDetails]>`

tomholub (commented):
Looks good - see comments. Where is the timeout handled for WKD lookup? There should be a timeout of 4 seconds (Attester could have a timeout of 10 seconds if not already present - please check).

tomholub:
@ekievsky please let me know when ready. I understood this is the stage two of #296 - looking forward to get this merged and move onto stage three. Thanks!

ekievsky:
Finishing with timeouts, will be done today.

tomholub:
> Can't add tests for WKDURLsApi and AttesterApi because there is no networking entity that injects into these classes and could be spied.
> P.S. It would be great to refactor Networking call URLSession.shared.call(_ urlRequest: URLRequest, tolerateStatus: [Int]? = nil) -> Promise<HttpRes> for separate networking entity to test classes that make API calls.
> 

Please update these as appropriate to make them more testable.

ekievsky:
@tomholub I've updated PR, resolved your comments and added timeout.

According to networking, I am not sure that it's a good idea to complete it in scope of this task. Because we need to create a separate layer for networking and start using this. I would discuss the implementation with @Kharchevskyi to keep project's code in one design. Please let me know your thought, guys.

tomholub:
Ok. I'll have a look

FlowCrypt/Functionality/Services/WKDURLsApi.swift
```diff
@@ -37,9 +40,10 @@ class WKDURLsApi: WKDURLsApiType {
             }
             let pubKeys = safeResponse.keyDetails
                     .filter { !$0.users.filter { $0.contains(email) }.isEmpty }
-                    .map(\.public)
             resolve(pubKeys)
         }
+        .timeout(Constants.lookupEmailRequestTimeout)
+        .recoverFromTimeOut(result: [])
```
tomholub, line 46: This is correct for WKD - because the recipient may or may not have a WKD server set up.

However, timeout is not the only failure mode here. If their server is not reachable, that should also result in `[]` (when servers are not set up, sometimes it results in a timeout and sometimes it results in "server not reachable" error right away - need to handle both).

tomholub (commented):
Almost there - see comments below.

tomholub:
Also please resolve the conflicts. Thanks!

FlowCrypt/Functionality/Services/WKDURLsApi.swift
ekievsky: it will return [] if any error occurs with request.

FlowCrypt/Functionality/Services/AttesterApi.swift
```diff
@@ -45,6 +50,8 @@ extension AttesterApi {
             // programming error because should never happen
             throw AppErr.unexpected("Status \(res.status) when looking up pubkey for \(email)")
         }
+        .timeout(Constants.lookupEmailRequestTimeout)
+        .recoverFromTimeOut(result: PubkeySearchResult(email: email, armored: nil))
```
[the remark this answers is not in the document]
ekievsky: done, removed recovering from timeout, so it will throw the timeout error in Promise and client that uses this class will be able to catch that error.

FlowCrypt/Functionality/Services/WKDURLsApi.swift
tomholub: Yes and no. 
 - If any error occurred when getting the `/policy` file, then that's a `[]`
 - However, if you are already past the policy stage, and now querying the URL of the public key, then only `HTTP 404` or timeout results in `[]`. Other errors should be escalated as usual.

I'll make an another issue for this.

FlowCrypt/Functionality/WKDURLs/PubLookup.swift
```diff
@@ -0,0 +1,53 @@
...
+            if !wkdResult.isEmpty {
+                resolve(wkdResult)
+            }
+            let attesterResult = try awaitPromise(self.attesterApi.lookupEmail(email: email))
+            guard let armoredData = attesterResult.armored,
+                  let attesterKeys = try? self.core.parseKeys(armoredOrBinary: armoredData),
+                  !attesterKeys.keyDetails.isEmpty else {
+                resolve([])
+                return
+            }
```
tomholub, line 47: Here I recommend to change attesterApi.lookupEmail result to return `[KeyDetails]` too for consistency (in another PR)

tomholub (approved):

Issue #307, opened 2021-05-07
use self-hosted IMAP server for testing


tomholub:
@DenBond7 can you please copy the whole test server setup from Android repo, and copy all IMAP test resources here as well? Then we can start writing test cases against it on iOS.

tomholub:
The goal of this issue is that it will start and run in each Semaphore CI run like on Android, but remain unused.

Then @Kharchevskyi will switch iOS test to start using it in another PR.

tomholub:
![image](https://user-images.githubusercontent.com/6306961/117456616-ec73b480-af37-11eb-8c83-17bad3dc120b.png)

DenBond7:
@Kharchevskyi Please feel free to ask me directly via skype. I will show you what I have. It'll be much faster.

tomholub:
Actually I prefer to keep the communication here. It's very important for future reference. Communication in skype gets lost.

DenBond7:
![image](https://user-images.githubusercontent.com/2863246/117470395-6cf9db80-af5f-11eb-8eaf-97129ae3dcf5.png)


https://github.com/FlowCrypt/flowcrypt-android/blob/4d8f1ded212ccaff5f6f0aef7023038354c37c9d/FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/activity/MessageDetailsActivityTest.kt#L184

tomholub:
@DenBond7 Regarding the source for the docker image that we use for IMAP/SMTP - where exactly is the source now? Is it on Android repo? Could you share a link?

DenBond7:
We store the email server code here
https://github.com/FlowCrypt/flowcrypt-android/tree/master/docker-mailserver

tomholub:
Understood - is it also on dockerhub somewhere? Or you run docker just from that directory, when running tests on Android?

DenBond7:
https://hub.docker.com/r/flowcrypt/flowcrypt-email-server

We've created an organization for that

tomholub:
Great. In that case, for now, you can leave the sources in Android repo. Then on iOS, you can just use the dockerhub directly, without need for android repo - correct?

tomholub:
Then moving  https://github.com/FlowCrypt/flowcrypt-android/tree/master/docker-mailserver to a separate github repo can happen sometime in the future

DenBond7:
[...]

PR #338, opened 2021-06-05
PubLookup integration

### LINKED ISSUE #296: integrate WKD public key search
trust it more than results on Attester (search first on WKD, then on Attester only if WKD doesn't return public key)

1) port `Wkd` class from browser extension https://github.com/FlowCrypt/flowcrypt-browser/blob/83047917fc3ef6853ef5323340a64d4b3703a9d3/extension/js/common/api/key-server/wkd.ts (parts of it already done in #314 )
2) add tests for it independently of the rest of the codebase, merge the class unused. You can use `robot@flowcrypt.com` as a test-case that returns a public key. The public key is returned in binary format (unarmored) - I'm not sure if our `Core` class is able to process these yet. Try and see - let me know if not.
3) another PR once the step 2 is merged - to start using this class during public key lookups. For each new recipient, search on WKD first (with a 3 second timeout). Only if public key not found on WKD, then search on Attester (the currently implemented lookup method).

ekievsky:
This PR integrates PubLookup instead of RemoteContactsProvider

close #296 

**Tests** :
- Difficult to test. We need to add separate layer of Networking to be able to inject it into appropriate classes and "spy" it.

### To be filled by reviewers


tomholub (approved):
Looks good. It troubles me that we can't easily test this, particularly relies on #307

Issue #339, opened 2021-06-05
prepwork for retrieving client configuration (OrgRules) from FES


tomholub:
part of #275
after this, you may do #340
after that can do #276 

FES = FlowCrypt Enterprise Server https://flowcrypt.com/docs/technical/enterprise-server/latest/enterprise-server.html

Flow:
1) parse out user domain: `user@example.com` -> `example.com`
2) skip all following steps if user is on public email domain (@ gmail, @ outlook)
3) call `GET https://fes.$domain/api/` (no authentication) for example `GET https://fes.corp.com/api/`. If it returns a 404 or the server cannot be reached within 4 seconds, skip following steps (similar failure handling to WKD policy call)
4) make sure that the response contains json parameter `service` which must equal `enterprise-server`. If it shows something else or this is not contained, show an error.
5) show a toast message that says `FES at $url not supported on iOS yet`

The class that does steps 1-4 should be `EnterpriseServerApiClient` eg a method `func getActiveFesUrl(for email: String): Promise<String?>` (nil if no FES runs there, else url of the server). 

Do the above during setup immediately after authentication (once you know user email address)

You can test this flow using your Google account at flowcrypt.com domain which I'll send you. Here is the API that you will be calling in that case: https://fes.flowcrypt.com/api/ but of course the domain is dynamic, so you may be calling another URL if you used account on another domain.

tomholub:
> There were 1 email addresses found in the above comment. Please:
> 
>     1. click `three dots` -> `edit` to remove the email addresses
> 
>     2. click `edited` in the comment header, and click on the previous revision of the comment
> 
>     3. when viewing the old revision with an email in it, click `options` -> `delete this revision from history`

@seisvelas it seems to be getting triggered but not sure why. The emails are @ example.com

seisvelas:
>@seisvelas it seems to be getting triggered but not sure why. 

I neglected to update this repo when I updated all the others to have exemptions. Right now I'll review all the public repos in our org to make sure I didn't miss any others. Sorry!

ekievsky:
@tomholub now the API returns `{"v": 4}`, is there a way to get a complete JSON?
what should be passed as a result of `String?`?

P.S. I was using flowcrypt.com domain for test

tomholub:
Are you sure? See this particular URL: https://fes.flowcrypt.com/api/

tomholub:
notice the pattern `GET https://fes.<domain>/api/`

[...]

PR #344, opened 2021-06-06
Prepwork for retrieving client config from FES

ekievsky:
This PR adds service to retrieve FES url configuration

issue #339

**Tests** :
- Difficult to test because we have no separate layer for networking that could be "spied"

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/App/Configuration.swift
+++ b/FlowCrypt/App/Configuration.swift
@@ -9,5 +9,5 @@
 import Foundation
 
 class Configuration {
-    static let excludedDomains = ["gmail.com", "googlemail.com", "outlook.com"]
+    static let publicEmailProviderDomains = ["gmail.com", "googlemail.com", "outlook.com"]
 }
--- a/FlowCrypt/App/Configuration.swift
+++ b/FlowCrypt/App/Configuration.swift
@@ -1,0 +1,13 @@
+//
+//  Configuration.swift
+//  FlowCrypt
+//
+//  Created by Yevhen Kyivskyi on 05.06.2021.
+//  Copyright © 2021 FlowCrypt Limited. All rights reserved.
+//
+
+import Foundation
+
+class Configuration {
+    static let excludedDomains = ["gmail.com", "googlemail.com", "outlook.com"]
+}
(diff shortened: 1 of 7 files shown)


ekievsky:
The next PR will be for integration FES url retrieving after auth and showing toast as a result.

FlowCrypt/App/Configuration.swift
```diff
@@ -0,0 +1,13 @@
...
+//  FlowCrypt
+//
+//  Created by Yevhen Kyivskyi on 05.06.2021.
+//  Copyright © 2021 FlowCrypt Limited. All rights reserved.
+//
+
+import Foundation
+
+class Configuration {
+    static let excludedDomains = ["gmail.com", "googlemail.com", "outlook.com"]
```
tomholub, line 12: Let's call it `publicEmailProviderDomains` - more descriptive of the content

FlowCrypt/Functionality/Services/EnterpriseServerApi.swift
```diff
@@ -0,0 +1,52 @@
...
+                return
+            }
+            let urlString = "https://fes.\(userDomain)/"
+            let request = URLRequest.urlRequest(
+                with: "\(urlString)api/",
+                method: .get,
+                body: nil
+            )
+
+            let response = try awaitPromise(URLSession.shared.call(request, tolerateStatus: [404]))
```
tomholub, line 38: It should tolerate basically all errors and return nil in those situations (since we don't know if there's a server running there or not until we try, for most users, no server will be running there, so we want to return nil graciously)

tomholub (commented):
I'd mark it as a draft since the integration with views is not in place yet

FlowCrypt/App/Configuration.swift
ekievsky: done

FlowCrypt/Functionality/Services/EnterpriseServerApi.swift
ekievsky: done

FlowCrypt/Functionality/Services/EnterpriseServerApi.swift
```diff
@@ -0,0 +1,52 @@
...
+    func getActiveFesUrl(for email: String) -> Promise<String?>
+}
+
+class EnterpriseServerApi: EnterpriseServerApiType {
+
+    private enum Constants {
+        static let getActiveFesTimeout: TimeInterval = 4
+
+        static let serviceKey = "service"
+        static let serviceNeededValue = "enterprise-server"
```
tomholub, line 21: Let's call this `expectedServiceValue`

tomholub:
> The next PR will be for integration FES url retrieving after auth and showing toast as a result.

Got it, then I could merge this

tomholub (approved):

Issue #348, opened 2021-06-07
@testable import instead of adding files to. the test target


Kharchevskyi:
In current implementation in case we want to cover some functionality with unit test, whole file with all related dependencies should be added to test target which lead to compiling each file 2 times, besides that for now not all files are added to both targets which makes testing quite complicated. 
For example `DataService` can't be easily add to test target, so it's hard to use it as default parameter and it should be "hidden" by some protocol to not include whole file.
Investigate is it possible to use `@testable import FlowCrypt` instead of adding files to test target.

Kharchevskyi:
@ekievsky what do you think about it?

ekievsky:
Agree, that's a good approach to avoid compiling the same file twice.

I'll try not to forget about it in new tests 😅

Kharchevskyi:
@ekievsky Not sure we are ready for that) 
I think we need to update Podfile and probably some frameworks can be an issue. 
Do you have any experience with migrating from "adding to test target" to "testable import"?

Kharchevskyi:
new tests - 😍

ekievsky:
I had no experience with migration, but I had worked on a project where we were using @testable and we just set up all the pods for both targets(main and tests).

tomholub:
I see

PR #356, opened 2021-06-18
issue 348 @testable import

Kharchevskyi:
This PR adds FlowCryptAppTest to use `@testable import FlowCrypt`

Part of #348

Pros: 
* Will compile code only once. Now files are compiling twice (Once for FlowCrypt and one for FlowCryptTests)
* Will run tests faster
* No need to add files to test targets.
Screenshot 2021-06-18 at 12 27 45

Before
Screenshot 2021-06-18 at 12 17 21

* Tests will run faster on CI.
* Easier to test. For instance if your need to test some `.swift` file all recursively dependent class/structs should be also added to test target.
If this is ok - will move test one by one to a new target, then old target will be removed.

**Tests**
- Tests added or updated

### To be filled by reviewers


FlowCryptAppTests/GmailSearchExpressionGeneratorTests.swift
```diff
@@ -7,6 +7,7 @@
 //
 
 import XCTest
+@testable import FlowCrypt
```
Kharchevskyi, line 10: `GmailSearchExpressionGenerator` is only in FlowCrypt target

ekievsky:
It would be great to merge this PR by Monday so I could include this feature into my PR and create the tests.

Kharchevskyi:
@tomholub if you ok with this solution I will proceed with this and finish probably today or tomorrow

tomholub (dismissed):
Sure, this looks good.

tomholub (commented):
So far no issues

Kharchevskyi:
- [x] Configured schemes to run tests on CI
- [x] Removed old test target
- [x] Moved all existed tests to a new Target

tomholub (approved):

PR #369, opened 2021-06-22
Refactoring InMemoryPassPhrase expiration

### LINKED ISSUE #363: expiration of in-memory pass phrases should happen inside InMemoryPassPhraseStorage
There should be a better separation of responsibilities. Everything relating to time tracking and expiring old in-memory passphrases should be handled in `InMemoryPassPhraseStorage`. This class that combines the two methods should be more dumb. I'll file a separate issue.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/323#discussion_r654793516_

Kharchevskyi:
This PR refactor logic of inMemoryPassPhrase expiration

close #363

**Tests**: 
- Tests added and updated

### To be filled by reviewers


FlowCrypt/Functionality/Services/Key Services/PassPhraseStorage.swift
```diff
@@ -8,41 +8,63 @@
...
+}
+
+// MARK: - In memory
+protocol InMemoryPassPhraseStorageType {
+    func save(passPhrase: InMemoryPassPhrase)
+    func update(passPhrase: InMemoryPassPhrase)
+    func remove(passPhrase: InMemoryPassPhrase)
+
+    func getPassPhrases() -> [InMemoryPassPhrase]
+}
```
tomholub, line 38: Since these protocols are the same (except `PassPhraseObject` vs `[InMemoryPassPhrase]`, could we unite them? And have only one type of `PassPhrase` returned?

tomholub (commented):
This looks good - see opportunity for potential simplification below.

FlowCrypt/Functionality/Services/Key Services/PassPhraseStorage.swift
tomholub: Also, there could be just one `PassPhraseProviderType` that covers both types of storages. Then I feel we are making actual use of the Type vs implementation distinction. That would be best.

Kharchevskyi: Actually it was done for a purpose.
Don't want to spread Realm data objects all over the place.
App layer communicates only with app layer data objects (PassPhrase) and storages with their internal objects. 
So in case smth changed (for instance we will switch to CoreDarta or any other storage) app will not be dependent on this changes.( _Besides that `InMemmoryPassPhrase` contains additionally date._)
This is common for the whole app, and actually helped us a lot when we was "moving" from "MailCore" to "GmailApi"

I will think, maybe to use `associatedObject` in protocols, to have only one, but I assume it can a bit problematic. Will check this

tomholub: Agree with not spreading realm objects everywhere. We could have a single `PassPhrase` object with a nilable expiration date on it, and then have a single protocol. But I don't insist, it's an implementation detail.

Kharchevskyi: As you suggest left only `PassPhrase` data object with optional `date`

tomholub (dismissed):
Looks good - thanks!

tomholub (approved):

Issue #387, opened 2021-07-09
replace all fatalError with throw


tomholub:

tomholub:
If yes, then we can close this. If no, then we should instead use a way to assert things that will also function in production - I don't want wrong assumptions in code in particular in production.

ekievsky:
`assertionFailure` will only crash on debug builds, it won't crash on production. We can change to `fatalError` so it will crash in production.

tomholub:
Ok, we should use `fatalError` in situations that must never happen, meaning if they do happen, it's our own code fault (like user with missing ClientConfiguration after setup). This will crash the app and then we'll see it in reports and can fix the code.

We will use Promise rejections with an appropriate error for situations that can happen (badly parsed external input, network error, unexpected API input etc) so that these don't crash the app.

ekievsky:
So in scope of this PR I am just going to replace `assertionFailure` with `fatalError`, Promises rejections should be refactored partly, right?

tomholub:
In the scope of this issue, for every `assertionFailure` you have to make a decision. Ask this question:

**Does this error scenario seem "impossible" on a well written, bug-free codebase?** 

- **Yes, it feels impossible unless there's a hidden bug**: Then it's a `fatalError`. Examples: a nil check when we've already nil-checked before, `account?.email` nil check when we are in a flow where the user already must have been logged in, state of the app that should never happen and so on. We want to crash the app, because something is quite wrong.

- **No, this could happen and it's not a code bug:** for example wrong user input, wrong response format from API, server down, server timeout, connection timeout, HTTP 4xx or 5xx and so on. These are errors which would be rendered to user. Then we need to produce an error that will be somehow shown to the user, and stop what we are doing.
    - if we are in a Promise, then this is a promise rejection with an error
    - if we are in view controller code and it makes sense to render error directly, then we render error
    - if we are in other code that does not have promises, I'm not too sure what to do here. Probably throw an error and catch it somewhere where you can render an error to the user (like VC).

tomholub:
You don't have to address every single `assertionFailure` in one PR, you can do one part of the app in one PR and then next one later - but make it clear which parts are there to do and which ones does the PR address.

Thanks!

ekievsky:
got you, thanks for explanation!

tomholub:
Should not be closed yet - I think not every assertionFailure was replaced yet.

ekievsky:
[...]

PR #405, opened 2021-07-21
Replace assertion failure with fatal error - View Controllers

ekievsky:
This PR replaces assertions with fatal errors in ViewControllers and OrgRules in appropriate places;

issue #387

**Tests** :
- Does not need tests

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Settings/KeyLegal/LegalViewControllersProvider.swift
+++ b/FlowCrypt/Controllers/Settings/KeyLegal/LegalViewControllersProvider.swift
@@ -32,8 +32,7 @@ private enum Items: Int, CaseIterable {
 
     var viewController: UIViewController {
         guard let link = self.url else {
-            assertionFailure()
-            return UIViewController()
+            fatalError("Links are hardcoded so they have not to be nil")
         }
         let vc = WebViewController(url: link)
         return vc
--- a/FlowCrypt/Models/OrganisationalRule.swift
+++ b/FlowCrypt/Models/OrganisationalRule.swift
@@ -85,8 +85,7 @@ class OrganisationalRules {
         }
 
         if keyManagerUrlForPrivateKeys == nil {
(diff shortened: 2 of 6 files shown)


FlowCrypt/Controllers/Settings/KeyLegal/LegalViewControllersProvider.swift
```diff
@@ -32,8 +32,7 @@ private enum Items: Int, CaseIterable {
 
     var viewController: UIViewController {
         guard let link = self.url else {
-            assertionFailure()
-            return UIViewController()
+            fatalError("Links are hardcoded so they have not to be nil")
```
tomholub, line 35: Exactly

FlowCrypt/Models/OrganisationalRule.swift
```diff
@@ -85,8 +85,7 @@ class OrganisationalRules {
         }
 
         if keyManagerUrlForPrivateKeys == nil {
-            assertionFailure("Wrong org rules config: using PRV_AUTOIMPORT_OR_AUTOGEN without key_manager_url")
-            return false
+            fatalError("Wrong org rules config: using PRV_AUTOIMPORT_OR_AUTOGEN without key_manager_url")
         }
```
tomholub, line 89: This is a bit complicated - here we work with OrgRules which are external input (from a remote server). They could theoretically be any combination, and it would not be a fault of the app or bug in the app. We could say that we only use `mustAutoImportOrAutogenPrvWithKeyManager` once we checked that `keyManagerUrlForPrivateKeys` exists, which is fair. 

I guess we can keep it how you did it. It is true that if we ever check `mustAutoImportOrAutogenPrvWithKeyManager` without checking `keyManagerUrlForPrivateKeys` or `isUsingKeyManager` first, it suggests wrong coding somewhere.

tomholub (approved):
@Kharchevskyi I'll merge this but please also give it a review, if something looks wrong, then we can fix it in followup PR.

PR #409, opened 2021-07-24
Could not parse org ruels

### LINKED ISSUE #408: couldn't parse data while fetching organizational rules
I have an email on nvimp.com domain. With this, the app will attempt fetching rules from https://fes.nvimp.com/api/ and fail, because there is no FES running there.

Most customers will not run our server software on their domain, therefore this is an expected / normal scenario and the app should continue working.

Follow this logic when calling FES first time to determine if it's running or not in `getActiveFesUrl`:

```
call `https://fes.<domain>/api/`, process the response:
  -> did I receive a response and it's status 200, it's a JSON and has the right service value? -> use FES
  -> did I receive a response and it's status 404? -> don't use FES  
  -> did I receive a response and it's other status 4xx or 5xx? -> show error to user
  -> did I NOT receive a response (server down or timeout) and the internet is otherwise working? -> don't use FES
  -> did I NOT receive a response (server down or timeout) and the internet is not working? -> show error to user
```

That means if I call `https://fes.nvimp.com/api/` and the server is down, but internet is working, then this user does not use FES.

ekievsky:
This PR tolerates not found, timeout and service unavailable statues in EmaiKeyManagerApi and adds handling parsing error in EmailKeyManagerApi

close #408

- Tests will be added later

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Setup/SetupInitialViewController.swift
+++ b/FlowCrypt/Controllers/Setup/SetupInitialViewController.swift
@@ -161,9 +161,11 @@ extension SetupInitialViewController {
                 guard let urlString = self?.emailKeyManagerApi.getPrivateKeysUrlString() else {
                     fatalError("Private keys URL can not be nil at this point")
                 }
-                self?.showToast(
-                    "organisational_rules_ekm_private_keys_message".localizeWithArguments(result.privateKeys.count, urlString)
-                )
+                if let privateKeys = result?.privateKeys {
+                    self?.showToast(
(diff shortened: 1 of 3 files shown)


tomholub (changes requested):
This one tolerates bad response from the EKM which is wrong - I don't want to tolerate that. If we get client configuration from FES and it says there is EKM running, then we must not ignore response from EKM.

The issue was about tolerating situation when FES is not running on `https://fes.<domain>`. Because most customers will not run it there.

I'm closing this as this PR will have to start over, there is no overlap between this PR and what I'd like done.

PR #410, opened 2021-07-25
Improve send message checks

### LINKED ISSUE #399: Improve send message checks
Unsure how was this handled before, looks like we should improve it, but that would be for another PR.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/306#r669354549_

Kharchevskyi:
This PR extracts message validation and sending logic to ComposeMessageService. 
Adds improvements to compose message validation

close #399

**Tests**:
- Tests added

### To be filled by reviewers


FlowCrypt/App/PromiseExtensions.swift
```diff
@@ -24,6 +25,16 @@ extension Promise {
     }
 }
 
+extension Future {
+    static func resolveAfter<T, E: Error>(timeout: TimeInterval = 5, with result: Result<T, E>) -> Future<T, E> {
```
tomholub, line 29: This looks like it belongs in test folder?

Kharchevskyi: good point. Will move to test target

Kharchevskyi (commented):
LGTM

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -0,0 +1,137 @@
...
+
+        let subject = input.subjectReplyTitle
+            ?? contextToSend.subject
+            ?? "(no subject)"
+
+        guard let myPubKey = self.dataService.publicKey() else {
+            return .failure(.validationError(.missedPublicKey))
+        }
+
+        switch getPubKeys(for: recipients) {
```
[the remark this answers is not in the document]
tomholub: Isn't there some mechanism like

```ts
let recipientPublicKeys = await getPubKeys(for: recipients)

let replyToMimeMsg = input.replyToMime
   .flatMap { String(data: $0, encoding: .utf8) }

let msg = SendableMsg(
    text: text,
    to: recipients.map(\.email),
    cc: [],
    bcc: [],
    from: email,
    subject: subject,
    replyToMimeMsg: replyToMimeMsg,
    atts: atts,
    pubKeys: allRecipientPubs + [myPubKey]
)
return .success(msg)
```

That would re-propagate the error from `getPubKeys` automatically?

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -0,0 +1,137 @@
...
+            return .failure(.noPubRecipients(emailsWithoutPubKeys))
+        }
+
+        return .success(pubKeys.compactMap(\.1))
+    }
+
+    // MARK: - Encrypt and Send
+    func encryptAndSend(message: SendableMsg) -> AnyPublisher<Void, ComposeMessageError> {
+        messageGateway.sendMail(mime: encryptMessage(with: message))
+            .mapError { ComposeMessageError.gatewayError($0) }
```
tomholub, line 126: What does `.mapError` with `.eraseToAnyPublisher` do?

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -0,0 +1,137 @@
...
+            .eraseToAnyPublisher()
+    }
+
+    private func encryptMessage(with msg: SendableMsg) -> Data {
+        do {
+            return try core.composeEmail(msg: msg, fmt: MsgFmt.encryptInline, pubKeys: msg.pubKeys).mimeEncoded
+        } catch {
+            fatalError()
+        }
+    }
```
tomholub, line 136: It's possible for this to fail for various reasons. For example, try to run this with an invalid public key. It will produce an error, and this should be handled gratiously.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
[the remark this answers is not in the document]
Kharchevskyi: `getPubKeys` returns `Result<[String], MessageValidationError>`, it can be success or failure cases with some associated values in it. 
I can't use guard or if because I will miss one of 2 associated values
```
guard case .success(let allRecipientPubs) = getPubKeys(for: recipients) else {
           return ...(here I don't have associated error if getPubKeys fails)
}
```

It's possible to use `do catch` here, but I wouldn't recommend this solution, because we need to cast to some particular error in catch 
```
do {
            try getPubKeys(for: recipients).get()
        } catch let error {
            (here it's Error type, not MessageValidationError)
            print(error)
        }
```

best way is to use `Result<SendableMsg, ComposeMessageError>.Publisher` and then in view controller we can chain `validateMessage` with `encryptAndSend(message: sendableMessage)` functions. 
Please check latest commit.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
Kharchevskyi: `func sendMail(mime: Data) -> Future<Void, Error>` will return just Error for us. But we are expecting to have `ComposeMessageError`, so mapError in case of error will transform this to `ComposeMessageError.gatewayError` with associated error. either from Imap or Gmail api. 

`eraseToAnyPublisher` - exposes an instance of AnyPublisher to the downstream subscriber, rather than this publisher’s actual type. 
So without this it will have a return type as `Publishers.MapError<Future<Void, Error>, ComposeMessageError>`. 
Please don't ask why)) Swift - 🤷
Type erasure is not the best in swift nowadays. 
https://developer.apple.com/documentation/combine/just/erasetoanypublisher()

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
Kharchevskyi: Just "replayed" previous app behaviour. 
Should I fix it in scope of this PR?

tomholub: Got it, let's make another PR then. There, we'll check exactly what happens when we give it some bad input.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -83,27 +83,24 @@ final class ComposeMessageService {
...
+                    cc: [],
+                    bcc: [],
+                    from: email,
+                    subject: subject,
+                    replyToMimeMsg: replyToMimeMsg,
+                    atts: atts,
+                    pubKeys: allRecipientPubs + [myPubKey]
+                )
+            }
     }
```
tomholub: It's similarly weird to use `map` if we are talking about a single Future executing one time with a single failure or success. Map implies working with array of potentially many items, which seems a mismatch for what we're doing (even if this is otherwise common in iOS development).

I'll merge this but file a separate issue.

tomholub (approved):

PR #413, opened 2021-07-27
Hangs forever when importing key from a file

### LINKED ISSUE #406: hangs forever when importing key from a file
This file has several keys all encrypted by the same pass phrase.

image

It hangs forever after I click continue.

ekievsky:
This PR ignoring empty keys retrievalю

close #406 

**Tests**:

### To be filled by reviewers


ekievsky:
@Kharchevskyi please have a look if it's not breaking a `failure` state.

tomholub (commented):
While this may address the encountered error in particular, it does not address a deeper problem: failure state returned by this method is getting ignored in the VC.

Let's make sure to always address the deeper problem as well.

tomholub (approved):
Code looks good. Will test

PR #420, opened 2021-07-29
Couldn't parse org rules

### LINKED ISSUE #408: couldn't parse data while fetching organizational rules

ekievsky:
This PR tolerates not found, timeout, service unavailable errors in EnterpriseService;

close #408

**Tests**:
- Does not need tests 

### To be filled by reviewers


FlowCrypt/Functionality/Services/EnterpriseServerApi.swift
```diff
@@ -32,6 +32,9 @@ extension EnterpriseServerApiError: LocalizedError {
 class EnterpriseServerApi: EnterpriseServerApiType {
 
     private enum Constants {
+        /// 404 - Not Found, -1001 - request timed out, 503 - service unavailable
+        static let getToleratedStatuses = [404, -1001, 503]
```
tomholub, line 36: Please strictly follow this decision chart:

```
call https://fes.<domain>/api/, process the response:
-> did I receive a response and it's status 200, it's a JSON and has the right service value? -> use FES to grab OrgRules
-> did I receive a response and it's status 404? -> don't use FES -> empty OrgRules
-> did I receive a response and it's other status 4xx or 5xx? -> show error to user
-> did I NOT receive a response (server down or timeout) and the internet is otherwise working? -> don't use FES, empty OrgRules
-> did I NOT receive a response (server down or timeout) and the internet is not working? -> show error to user
```

503 falls under `did I receive a response and it's other status 4xx or 5xx? -> show error to user`

FlowCrypt/Functionality/Services/EnterpriseServerApi.swift
```diff
@@ -99,12 +114,24 @@ class EnterpriseServerApi: EnterpriseServerApiType {
...
+                )
+            ) else {
+                reject(EnterpriseServerApiError.parse)
+                return
+            }
+
+            if Constants.getToleratedStatuses.contains(safeReponse.status) {
+                resolve(.empty)
+            }
+
```
tomholub, line 130: We should only tolerate the error scenarios when calling `https://fes.<domain>/api/` endpoint specifically, because at that point we don't know if there is any server running there or not.

Don't tolerate any errors when calling `https://fes.\(userDomain)/api/v1/client-configuration` or any other APIs, because once we are calling those, we know that there was supposed to be FES running there (based o successful `https://fes.<domain>/api/` response earlier)

tomholub (changes requested):
> -> did I receive a response and it's other status 4xx or 5xx? -> show error to user

Where was this implemented?

> -> did I NOT receive a response (server down or timeout) and the internet is otherwise working? -> don't use FES, empty OrgRules

Where was this implemented?

> -> did I NOT receive a response (server down or timeout) and the internet is not working? -> show error to user

Where was this implemented?

---------

This PR only addresses 20% of the requirements that were clearly laid out in the issue. If the PR addresses less then all requirements of an issue, you must clearly highlight which parts you addressed, and don't `close` the issue in the PR description.

FlowCrypt/Functionality/Services/EnterpriseServerApi.swift
ekievsky: so how may I understand if `server down`?

tomholub: In Swift, I'm not sure. But generally, it should fail right away, typically with unknown host:

```
$ curl https://unknownnonexistingserver.com
curl: (6) Could not resolve host: unknownnonexistingserver.com
```

tomholub: I'm not sure if one can differentiate `server down` from `timeout`. Most servers that are not there, if I try to curl them, I'm getting a timeout. I think that it would be possible to get a response that says "nothing is running here" but I'm not sure what that would look like.

Anyway, 503 counts as 4xx and 5xx, the only tolerated http status, other than 200, is 404.

To summarize:
 - 503 counts as 5xx
 - for now we can assume that `server down` is the same as timeout
 - need to ensure that "could not resolve host" is also handled

FlowCrypt/Functionality/Services/EnterpriseServerApi.swift
```diff
@@ -32,7 +32,11 @@ extension EnterpriseServerApiError: LocalizedError {
 class EnterpriseServerApi: EnterpriseServerApiType {
 
     private enum Constants {
-        static let getActiveFesTimeout: TimeInterval = 4
+        /// 404 - Not Found
+        static let getToleratedHTTPStatuses = [404]
+        /// -1001 - request timed out, -1003 - сannot resolve host
+        static let getToleratedNSErrorCodes = [-1001, -1003]
+        static let getActiveFesTimeout: TimeInterval = 100
```
tomholub, line 39: Also, what is the number 100 here, 100 seconds? Milliseconds?

ekievsky: for some reason my lates commit wasn't included :/

tomholub (approved):

tomholub (approved):

PR #424, opened 2021-07-31
issue 307 use external IMAP server for CI

tomholub:
This PR attempts to fix IMAP/SMTP mailserver on semaphore for CI tests.

issue #307

Tests:
 - this adds test infrustructure

### To be filled by reviewers


Kharchevskyi:
@tomholub what the purpose of running imap server on CI if we decide to use gmail for ui tests?

tomholub:
We'll do both. 

We can't reliably run Gmail UI tests in CI, because we can't control the login process fully. Gmail tests are still important to ensure login isn't broken, but we'll have to run them manually probably on my own device before shipping an update.

On the other hand, IMAP server login is predictable, so we should be able to write CI tests that are not flaky, and run them on every commit.

So we need both. Everything that can be tested with imap will be tested with IMAP to run on every CI. Then the remaining things that cannot be tested that way will use Gmail and I'll run those tests manually.

tomholub (commented):
lgtm

PR #431, opened 2021-08-05
Fix openning message with expired pass phrase

### LINKED ISSUE #416: app keeps forgetting all keys
reproduce:
1) set up account, choose pass phrase in memory
2) use the app for a bit
3) use it 4 hours later or overnight - open a message

expected: keys are permanently stored and remembered, the app remains usable but asks for pass phrase

actual:

image

65C387EE-FB4A-4B02-A303-C5B53A98CAC3

### LINKED ISSUE #426: decryptedKey response from Core should not be nullable
```swift
    struct DecryptKey: Decodable {
        let decryptedKey: String?
    }
```

should be

```swift
    struct DecryptKey: Decodable {
        let decryptedKey: String
    }
```

Kharchevskyi:
This PR fixes bug when pass phrase expired
For testing purposes timeout can be changed in `InMemoryPassPhraseStorage`

close #416 
 
**Tests** :
- Difficult to test. Time consuming

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Services/Key Services/KeyService.swift
+++ b/FlowCrypt/Functionality/Services/Key Services/KeyService.swift
@@ -9,8 +9,8 @@
 import Foundation
 
 protocol KeyServiceType {
-    func retrieveKeyDetails() -> Result<[KeyDetails], KeyServiceError>
-    func getPrivateKeys(with passPhrase: String?) -> Result<[PrvKeyInfo], KeyServiceError>
+    func getPrvKeyDetails() -> Result<[KeyDetails], KeyServiceError>
+    func getPrvKeyInfo(with passPhrase: String?) -> Result<[PrvKeyInfo], KeyServiceError>
 }
 
 enum KeyServiceError: Error {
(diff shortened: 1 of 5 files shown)


tomholub (commented):
Does this still allow me to view list of keys in Settings -> Keys section when the pass phrase expired?

Kharchevskyi:
> Does this still allow me to view list of keys in Settings -> Keys section when the pass phrase expired?

hm...good catch, I assume - no. Will fix

tomholub:
The design of this method needs to change. It's good that it collects pass phrases, but it should not be required. It's ok to return keys missing pass phrases if they are missing. Let the downstream code deal with that.

Kharchevskyi (commented):
LGTM

FlowCrypt/Functionality/Services/Key Services/KeyService.swift
```diff
@@ -34,26 +34,19 @@ final class KeyService: KeyServiceType {
     }
 
     func retrieveKeyDetails() -> Result<[KeyDetails], KeyServiceError> {
```
tomholub, line 36: The difference between `retrieveKeyDetails` and `getPrivateKeys` is not obvious from the name. Let's rename it:
 - `retrieveKeyDetails` -> `getPrvKeyDetails`
 - `getPrivateKeys` -> `getPrvKeyInfo`

Still not so obvious, but at least it highlights the different formats.

Then, please add a comment to highlight what is each method for, why do we have two similar methods.

tomholub (changes requested):

tomholub (approved):

PR #434, opened 2021-08-09
Passphrase setup for EKM keys

### LINKED ISSUE #279: implement EKM integration to load private keys during setup
EKM = Email Key Manager https://flowcrypt.com/docs/technical/enterprise/email-deployment-overview.html

OrgRule definitions https://flowcrypt.com/docs/business/org-rules.html

As a part of #275 and after #276 and #277 , immediately after authentication when we receive the OIDC and OrgRules, we should check if `orgRules.usesKeyManager() == true`. If yes, we should:

- [x] 7) call `GET <ekm>/v1/keys/private`. Into authorization header please put `Bearer <ID_TOKEN>`. On error, offer retry
- [x] 8) if there are no private keys there (empty array), show an error to the user that there are no private keys configured for them and that they should ask their systems administrator or help desk. Offer retry
- [x] 9) check that all of the received private keys are fully decrypted
- [ ] 10) if there is at least one private key, ask user to provide a passphrase (similar UI to creating a new key), then ask to confirm the pass phrase
- [ ] 11) encrypt the received keys with the pass phrase. Keep pass phrase in memory #197  , store encrypted key in storage
- [ ] 12) finish setup without any further user interaction

The goal is that if user has keys already configured on EKM and appropriate OrgRules are in place, they only need to authenticate and choose a pass phrase, and everything will be done automatically. After authentication and successful automatic setup, they will be sent to their inbox.

In this flow, __do not submit any public key to attester__.

ekievsky:
This PR adds passphrase setup with ekmkes

close #279

**Tests**:
- Tests will be added later 

### To be filled by reviewers


FlowCrypt/Controllers/Setup/SetupCreatePassphraseViewController.swift
```diff
@@ -23,14 +23,19 @@ enum CreateKeyError: Error {
...
+
 /**
  * Controller which is responsible for generating a new key during setup
  * - User is sent here from **SetupInitialViewController** in case there are no backups found
  * - Here user can enter a pass phrase (can be saved in memory or in encrypted storage) and generate a key
  * - After key is generated, user will be redirected to **main flow** (inbox view)
  */
 
-final class SetupGenerateKeyViewController: TableNodeViewController, PassPhraseSaveable {
+final class SetupCreatePassphraseViewController: TableNodeViewController, PassPhraseSaveable {
```
tomholub, line 38: Careful - here you are co-mingling two different setup flows into the same file, which I would like to avoid, for clarity. The way we have designed the setup flows is that each distinct flow gets a distinct VC. (you can still abstract the UI logic since it's similar - in both cases pass phrase needs to be created)

`SetupGenerateKeyViewController` was a situation when consumer-user installs the app, doesn't have any key, and is offered to create one and back it up in inbox. The name describes the intent well here.

I understand that the UI is very similar between that flow and what you are implementing, but the purpose is very different, and I'd like that to be clear from the code.

You could extract UI-specific stuff into a new abstract class `SetupCreatePassphraseAbstractViewController`, for example. Then you could update `SetupGenerateKeyViewController` to use it, but don't add EKM functionality there.

Then add a new class, `SetupWithKeysFromEkmViewController`, which will also inherit from newly created `SetupCreatePassphraseViewController`, and contain this newly added logic.

Then, finally, on the previous view, you forward user to `SetupWithKeysFromEkmViewController` when keys were fetched from EKM.

Then the distinction will be very clear.

FlowCrypt/Controllers/Setup/SetupCreatePassphraseViewController.swift
```diff
@@ -188,6 +195,48 @@ extension SetupGenerateKeyViewController {
         }
     }
 
+    private func setupAccountWithExistingKeys(with passPhrase: String) {
```
tomholub, line 198: This, once moved into `SetupWithKeysFromEkmViewController` should be renamed to `setupAccountWithKeysFetchedFromEkm`

FlowCrypt/Controllers/Setup/SetupCreatePassphraseViewController.swift
```diff
@@ -263,13 +312,17 @@ extension SetupGenerateKeyViewController {
             return
         }
         logger.logInfo("Setup account with \(passPhrase)")
-        setupAccountWithGeneratedKey(with: passPhrase)
+        if keys.isNotEmpty {
+            self.setupAccountWithExistingKeys(with: passPhrase)
+        } else {
+            setupAccountWithGeneratedKey(with: passPhrase)
+        }
```
tomholub, line 319: This decision will happen in the VC that fetched the keys from EKM - in EKM flow it will use the dedicated EKM VC.

FlowCrypt/Controllers/Setup/SetupInitialViewController.swift
```diff
@@ -158,16 +158,10 @@ extension SetupInitialViewController {
...
-                }
-
+            .then(on: .main) { [weak self] result in
                 switch result {
-                case .success(decryptedResponse: let response):
-                    self?.showToast(
-                        "organisational_rules_ekm_private_keys_message".localizeWithArguments(response.privateKeys.count, urlString)
-                    )
+                case .success(keys: let keys):
+                    self?.proceedToSetupPassPhrase(keys: keys)
```
tomholub, line 164: Here you will send to user to EKM VC directly.

FlowCrypt/Controllers/Setup/SetupCreatePassphraseViewController.swift
```diff
@@ -188,6 +195,48 @@ extension SetupGenerateKeyViewController {
...
+                        throw CreatePassphraseWithExistingKeyError.noPrivateKey
+                    }
+                    let encryptedPrv = try self.core.encryptKey(
+                        armoredPrv: privateKey,
+                        passphrase: passPhrase
+                    )
+                    let parsedKey = try self.core.parseKeys(armoredOrBinary: encryptedPrv.encryptedKey.data())
+                    parsedKey.keyDetails.forEach { keyDetails in
+                        let passPhrase = PassPhrase(value: passPhrase, fingerprints: keyDetails.fingerprints)
+                        self.passPhraseService.savePassPhrase(with: passPhrase, inStorage: self.shouldSaveLocally)
```
tomholub, line 217: The pass phrase should only be saved in memory - following the OrgRules. We have already checked that `FORBID_STORING_PASS_PHRASE` is `true` earlier. Now we need to follow it.

FlowCrypt/Controllers/Setup/SetupCreatePassphraseViewController.swift
```diff
@@ -188,6 +195,48 @@ extension SetupGenerateKeyViewController {
...
+                    }
+                    try awaitPromise(self.backupService.backupToInbox(keys: parsedKey.keyDetails, for: self.user))
```
tomholub, line 219: Don't do this in this flow. The keys are already stored on EKM, no need to back them up anywhere else. This would be a severe vulnerability if merged in, easily enough to lose a customer if I didn't catch it in the PR review.

It shows a lack of clarity or awareness in our business use-case. I recommend to take more interest in the business side of things, understand not only what we do but why we do things in a certain way, study the documentation, and use the app itself a lot more, to educate yourself about our products.

Only with full understanding of what our company does and why, and how all of our products fit together, can you be a truly effective contributor.

FlowCrypt/Controllers/Setup/SetupCreatePassphraseViewController.swift
```diff
@@ -188,6 +195,48 @@ extension SetupGenerateKeyViewController {
...
+                    self.keyStorage.addKeys(keyDetails: parsedKey.keyDetails, source: .generated, for: self.user.email)
```
tomholub, line 220: The `source` here is wrong, the key was not generated in the app.

tomholub (changes requested):

FlowCrypt/Controllers/Setup/SetupCreatePassphraseViewController.swift
ekievsky: Should I add a new case for `source`?

tomholub: Yes, let's do that.

FlowCrypt/Controllers/Setup/SetupCreatePassphraseViewController.swift
ekievsky: Should we show a hint to user saying that passphrase will be saved in memory?

tomholub: Good question! Not this time, because we have only one specific customer who has this type of setup, and they will train their stuff to know this. Additionally, in the future, we may allow users to choose temporary vs permanent storage during setup with the EKM flow, similarly how users can choose it in other flows. So I figure it will be easier to add this optionallity if we don't do other adjustments.

Normally, however, this is a good suggestion.

tomholub:
It's progressing well. For comments that you have addressed, please click "Resolve conversation" after you have pushed the appropriate commit, to make it clear which ones were addressed and which not.

FlowCrypt/Controllers/Setup/SetupEKMKeyViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+        super.init(
+            user: user,
+            core: core,
+            router: router,
+            decorator: decorator,
+            storage: storage,
+            keyStorage: keyStorage,
+            passPhraseService: passPhraseService
+        )
+        self.shouldStorePassPhrase = true
```
tomholub, line 55: You have to read issue descriptions more carefully, I've spent a lot of effort to outline the issue in great detail, and it's wasted effort if it's not read and followed.

Have a look at this https://user-images.githubusercontent.com/6306961/124514325-4f64b900-ddcc-11eb-9419-01973f122a37.png
and this https://fes.flowcrypt.com/api/v1/client-configuration?domain=flowcrypt.com
And this https://github.com/FlowCrypt/flowcrypt-ios/issues/279
And this https://flowcrypt.com/docs/business/org-rules.html

You should see a discrepancy very easily.

ekievsky: oh, I did check all the docs and understand the flow, just messed up with name of variable :(

tomholub: Ok. Thank you for responding - it's really important to be careful.

tomholub (approved):
All looks good. Code is clear and well structured. Functionality works now.

In the future, more careful studying of the issue definition and larger context will help you code with more confidence, and reduce my workload when reviewing.

Thank you.

tomholub:
To clarify, it still holds that in cases when the issue is not well defined, asking questions is way better then guessing.

PR #499, opened 2021-09-10
EKM unit tests

### LINKED ISSUE #498: Unit tests for EKM integration

Kharchevskyi:
This PR adds unit tests for ekm integration

close #498  

**Tests**: 
- Tests added or updated

### To be filled by reviewers


FlowCrypt/Models/OrganisationalRule.swift
```diff
@@ -40,7 +40,12 @@ class OrganisationalRules {
 
     /// use when finding out if EKM is in use, to change functionality without actually neededing the EKM
     var isUsingKeyManager: Bool {
-        clientConfiguration.keyManagerUrl != nil
+        // check for empty string
+        guard let urlString = clientConfiguration.keyManagerUrl, urlString.isNotEmpty else {
+            return false
+        }
+        // check is url can be configured
+        return URL(string: urlString) != nil
```
tomholub, line 48: Good initiative to add additional checks - thank you! I'd like the to behave differently. If `keyManagerUrl` is not-null and it's either empty or not a proper URL, the app should show an error and log out.

The reasoning is: some admin somewhere meant to set EKM url to something meaningful, but set it to a wrong string. The intention was that the app speaks to some EKM, so we should not default to null (which would make the app change behavior in ways that may not be secure for companies that want to use EKM). The app should just refuse to do things because the Client Configuration is clearly misconfigured - some admin should then go and fix it.

I prefer to surface errors to the users rather then ignore them, if they may have security implications.

Thanks!

Kharchevskyi: Added in next commit. Please check `organisational_rules_url_not_valid` error message.

FlowCryptAppTests/Functionallity/OrganisationalRulesTests.swift
```diff
@@ -0,0 +1,109 @@
...
+        XCTAssertFalse(sut.isKeyManagerUrlValid)
+
+        clientConfiguration = ClientConfiguration(keyManagerUrl: "not a url string")
+        XCTAssertFalse(sut.isKeyManagerUrlValid)
+    }
+
+    func testMustAutoImportOrAutogenPrvWithKeyManager() {
+        clientConfiguration = ClientConfiguration(
+            flags: [.privateKeyAutoimportOrAutogen],
+            keyManagerUrl: "https://test.ekm.flowcrypt.com"
```
tomholub, line 39: Please use some URL that's obvious it's not a real URL like `https://ekm.example.com` in all of these tests.

Then it's obvious which ones may actually get called during tests and which not.

FlowCryptAppTests/Functionallity/OrganisationalRulesTests.swift
```diff
@@ -0,0 +1,109 @@
...
+//  Created by Anton Kharchevskyi on 17.09.2021.
+//  Copyright © 2021 FlowCrypt Limited. All rights reserved.
+//
+
+import Foundation
+import XCTest
+@testable import FlowCrypt
+
+class OrganisationalRulesTests: XCTestCase {
+    var sut: OrganisationalRules! {
```
tomholub, line 14: What does `sut` here mean?

tomholub: In another PR, we should refactor the naming. It's not clear what's the benefit of having an `OrganisationalRules` object that only wraps `ClientConfiguration` object. We should just have a single `ClientConfiguration` object and that's all. If we do need a wrapper for some justified reason, then we can call that `ClientConfigurationWrapper` and add a comment on why it's needed. If it's to match it to a http response, then we should call it `ClientConfigurationResponse` that contains `ClientConfiguration`.

tomholub (commented):
Thanks - see comments

FlowCryptAppTests/Functionallity/OrganisationalRulesTests.swift
Kharchevskyi: System under test. 
This is pretty common to use this naming in iOS unit tests

https://en.wikipedia.org/wiki/System_under_test

tomholub: got it. that's ok

FlowCryptAppTests/Functionallity/Services/Organisational Rules Service/ClientConfigurationServiceTest.swift
```diff
@@ -0,0 +1,167 @@
...
+    func testCheckDoesNotUseEKM() {
+        // EKM should not be used if keyManagerUrl is nil
+        organisationalRulesService.clientConfiguration = ClientConfiguration(keyManagerUrl: nil)
+        XCTAssert(sut.checkShouldUseEKM() == .doesNotUseEKM)
+
+        // EKM should not be used if keyManagerUrl is nil
+        organisationalRulesService.clientConfiguration = ClientConfiguration(
+            flags: [],
+            keyManagerUrl: nil
+        )
```
tomholub, line 33: I think you wanted to add `XCTAssert(sut.checkShouldUseEKM() == .doesNotUseEKM)` here else this line has no effect

FlowCryptAppTests/Functionallity/Services/Organisational Rules Service/ClientConfigurationServiceTest.swift
```diff
@@ -0,0 +1,167 @@
...
+        organisationalRulesService.clientConfiguration = ClientConfiguration(
+            flags: [
+                .privateKeyAutoimportOrAutogen,
+                .forbidStoringPassphrase
+            ],
+            keyManagerUrl: "https://ekm.example.com"
+        )
+
+        XCTAssert(sut.checkShouldUseEKM() == .usesEKM)
+    }
```
tomholub, line 52: As per this: 

image

taken from here: https://github.com/FlowCrypt/flowcrypt-ios/issues/279

These flags as set up should produce an error, because the flags are missing `NO_PRV_CREATE` which translates to `.forbidCreatingPrivateKey`. I suppose we don't have this flag implemented in the app, so I'll make a separate issue for it.

FlowCryptAppTests/Functionallity/Services/Organisational Rules Service/Mocks/CurrentUserEmailMock.swift
```diff
@@ -0,0 +1,18 @@
+//
```
tomholub, line 1: The folder name `Functionallity` is misspelled, it should be `Functionality`

tomholub (approved):

tomholub (changes requested):
Missing test check as above - else ok

tomholub:
@Kharchevskyi after this PR, please look into the database schema changes. After that, I should be ready to make a testflight release.

tomholub:
@Kharchevskyi correction - first need to do this one: https://github.com/FlowCrypt/flowcrypt-ios/issues/516 and after that the database changes

tomholub:
@Kharchevskyi please have a look to finalize when you have a chance

Kharchevskyi:
Will check asap. 
Fyi: on vacation next week but I think going to work

tomholub:
> Fyi: on vacation next week

Understood - please give me more advanced notice whenever possible, so I can organize appropriately. Thanks!

FlowCryptAppTests/Functionallity/Services/Organisational Rules Service/ClientConfigurationServiceTest.swift
Kharchevskyi: thanks. fixed

tomholub (approved):

Issue #404, opened 2021-07-21
Replace Promise with async/await


tomholub:
continuation of https://github.com/FlowCrypt/flowcrypt-ios/issues/290

similarly how was done in https://github.com/FlowCrypt/flowcrypt-ios/pull/400 and #590

possibly after https://github.com/FlowCrypt/flowcrypt-ios/issues/402

It should not be one big PR. When possible, it would be better to make several PRs to change the app chunk by chunk. That way errors are less likely to slip in during refactoring, and also to prevent merge conflicts.

tomholub:
This may take more than one PR to do comfortably. Just take a chunk of functionality and convert from Promise to Future like you have just done, then do another chunk, etc until the `Promise` library can be dropped from dependencies.

tomholub:
not actionable until big PRs are merged:
 - #854 
 - #657

PR #505, opened 2021-09-17
ability to send encrypted attachments

### LINKED ISSUE #196: ability to send encrypted attachments
part of  #275

ekievsky:
This PR adds the ability to send encrypted attachments

close #196

- Tests will be added later

### To be filled by reviewers


FlowCryptUI/Cell Nodes/TextViewCellNode.swift
```diff
@@ -71,4 +83,9 @@ extension TextViewCellNode: ASEditableTextNodeDelegate {
     public func editableTextNodeDidFinishEditing(_ editableTextNode: ASEditableTextNode) {
         action?(.didEndEditing(editableTextNode.attributedText))
     }
+    
+    public func editableTextNodeDidUpdateText(_ editableTextNode: ASEditableTextNode) {
+        let calculatedHeight = editableTextNode.textView.sizeThatFits(textView.frame.size).height
+        textViewSizeUpdated?(self, calculatedHeight)
```
Kharchevskyi, line 89: is it possible to resize node inside `TextViewCellNode` without passing this to controller? 
It can avoid calling `setHeight` in case text changed?

``` swift 
public func editableTextNodeDidUpdateText(_ editableTextNode: ASEditableTextNode) {
  let calculatedHeight = editableTextNode.textView.sizeThatFits(textView.frame.size).height
  setHeight(calculatedHeight)   
}
```

Kharchevskyi: You can check if `self.supernode` inside `TextViewCellNode` returns node from controller?

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -285,6 +290,11 @@ extension ComposeViewController: ASTableDelegate, ASTableDataSource {
                 case .text: return self.textNode(with: nodeHeight)
                 case .subjectDivider: return DividerCellNode()
                 }
+            case (.main, 2):
```
tomholub, line 293: I wonder if the whole `tableNode` should be in some sort of decorator class? It looks very UI-heavy with not much business logic. What do you think @Kharchevskyi ?

tomholub: If so, then that would apply to the whole `extension ComposeViewController: ASTableDelegate, ASTableDataSource {` section

FlowCrypt/Controllers/Msg/MessageViewController.swift
```diff
@@ -439,11 +442,15 @@ extension MessageViewController: ASTableDelegate, ASTableDataSource {
             onDownloadTap: { [weak self] in
                 guard let self = self else { return }
                 self.filesManager.saveToFilesApp(file: self.processedMessage.attachments[index], from: self)
-                    .catch { error in
-                        self.showToast(
-                            "\("message_attachment_saved_with_error".localized) \(error.localizedDescription)"
-                        )
-                    }
+                    .sinkFuture(
```
tomholub, line 445: Converting to a `Future` is good, thanks

FlowCrypt/Core/Core.swift
```diff
@@ -128,21 +131,30 @@ final class Core: KeyDecrypter, CoreComposeMessageType {
...
-            "replyToMimeMsg": msg.replyToMimeMsg,
-            "atts": msg.atts.map { att in ["name": att.name, "type": att.type, "base64": att.base64] },
-            "format": fmt.rawValue,
-            "pubKeys": pubKeys,
-        ], data: nil)
-        // this call returned no useful json data, only bytes
-        return CoreRes.ComposeEmail(mimeEncoded: r.data)
+    func composeEmail(msg: SendableMsg, fmt: MsgFmt, pubKeys: [String]?) -> Future<CoreRes.ComposeEmail, Error> {
+        Future<CoreRes.ComposeEmail, Error> { [weak self] promise in
+            guard let self = self else { return }
```
tomholub, line 136: Here this call used to by synchronous but you've made it into a `Future`. Is this because the call is too slow? So you don't want it blocking the UI? Or other reason?

FlowCrypt/Functionality/Services/Backup Services/BackupService.swift
```diff
@@ -74,10 +74,11 @@ extension BackupService: BackupServiceType {
...
 
-            self.messageSender
-                .sendMail(mime: backupEmail.mimeEncoded)
+            self.core.composeEmail(msg: message, fmt: .plain, pubKeys: message.pubKeys)
+                .map(\.mimeEncoded)
+                
+                .flatMap(self.messageSender.sendMail)
                 .sink(
                     receiveCompletion: { result in
                         switch result {
```
tomholub, line 83: A bit of a mess here combining `Promise` and `Future` but not your fault, should be cleaned up in #404

tomholub: Maybe we could have a separate PR later to covert this `backupToInbox` to a Future. I'll file it.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageAttachment.swift
```diff
@@ -0,0 +1,59 @@
...
+        let assetResources = PHAssetResource.assetResources(for: asset)
+
+        guard let fileName = assetResources.first?.originalFilename else {
+            return nil
+        }
+
+        self.name = "\(fileName).pgp"
+        self.data = data
+        self.size = data.count
+        self.type = "image/jpg"
```
tomholub, line 35: This is a guess - it's not good to guess. The image may be one of many type of images, not just a jpeg. Real type should be put here to reflect the image. Also above it mentiones `pngData` so there's a mismatch there too.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageAttachment.swift
```diff
@@ -0,0 +1,59 @@
...
+import Photos
+
+struct ComposeMessageAttachment {
+    let name: String
+    let size: Int
+    let data: Data
+    let type: String
+}
+
+extension ComposeMessageAttachment {
```
tomholub, line 19: Maybe I'm misunderstanding this whole class, but is it really necessary to handle it with different methods type by type? There may be 1000 different file types. I know you mentioned something about using a "simpler method for file picking" so I suppose maybe this is the side effect of that.

In the future, will we will have to handle file types individually like this?

As an example, on browser extension, the browser picks up a file and it doesn't care what type of file it is. The name, data, and mime type string is all it cares.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -83,6 +82,15 @@ final class ComposeMessageService {
             return .failure(.validationError(.missedPublicKey))
         }
 
+        let sendableAttachments: [SendableMsg.Attachment] = contextToSend.attachments
```
tomholub, line 85: Please help me ensure that combined size of all attachment is under 10mb. If oversize, return a validation error informing the user (eventually we may implement streaming encryption and upload where the size would be unlimited as far as iOS is concerned, but we're not there yet)

tomholub (commented):
Preliminarly this is is - In particular the encryption/sending looks correct. Adding some comments.

FlowCrypt/Controllers/Compose/ComposeViewController.swift
ekievsky: I agree, we may create DataSource entity for this screen, @Kharchevskyi let's discuss that when it's convenient

tomholub:
Converting to draft while this is being discussed & until adjustments done.

Otherwise this is ready for me to start testing I suppose?

ekievsky:
yes, you may test it, I will answer on comments asap and apply fixes tomorrow. ThanksQ

tomholub (commented):
Well done as far as user test - I was able to send other types of files (video in my case). The sent file worked, I'm able to read it on the other end (browser extension).

There are a few things we'll later have to improve in the future, other than what I've mentioned above. All of these 4 are for followup issues for the future, since at this stage we're just focusing on the basics.

1) if it's a bigger file, it takes a while: there should be a progress bar. We've done something similar on Android. The first 30% of the progress bar are taken up by encryption (which we have no way to track progress of) and the last 70% is upload of the message to Gmail servers, which you should be able to track progress of.
2) on dark theme, the attachment names are barely readable (also when receiving)
3) the `.pgp` filename should not be visible in that list. So you don't have to alter the filename, after all (the encryption method should be already transparently renaming them)
4) later I should have a way to remove attachment from compose screen if I changed my mind before sending

That's functionality-wise. I cannot comment too much on code, maybe @sosnovsky do you want to have a look?

FlowCrypt/Core/Core.swift
ekievsky: Right, it blocks UI for showing the spinner while request is in progress

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageAttachment.swift
ekievsky: The problem is that ImagePickerDelegate returns different dictionaries with data. E.g. a photo taken from camera doesn't have a name, path and asset but photo that taken from library has all those fields.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageAttachment.swift
ekievsky: I checked with gmail app, they sending JPG data, so I fixed

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -681,9 +681,18 @@ extension ComposeViewController: UIImagePickerControllerDelegate, UINavigationCo
...
             return
         }
-        contextToSend.attachments.append(attachment)
+        appendAttachmentIfAllowed(attachment)
         node.reloadSections(IndexSet(integer: 2), with: .automatic)
     }
+
+    private func appendAttachmentIfAllowed(_ attachment: ComposeMessageAttachment) {
+        let totalSize = contextToSend.attachments.reduce(0, { $0 + $1.size })
+        if totalSize > GeneralConstants.Global.attachmentSizeLimit {
```
tomholub, line 690: Shouldn't this be `totalSize + attachment.size > GeneralConstants.Global.attachmentSizeLimit` ?

sosnovsky:
> That's functionality-wise. I cannot comment too much on code, maybe @sosnovsky do you want to have a look?

I checked code - looks great, it helped me to better understand app structure 👍
But I noticed small bug when adding multiple attachments, created issue for it  https://github.com/FlowCrypt/flowcrypt-ios/issues/547

tomholub (approved):

tomholub:
Since remaining improvements are filed as separate issues, I've merged it.

PR #544, opened 2021-09-25
issue #508 return core response directly instead of encoding - TypeScript parts

### LINKED ISSUE #508: refactor Core response from string to string + byte array
after https://github.com/FlowCrypt/flowcrypt-ios/issues/507

currently, when we call a method in Core, for example here: https://github.com/FlowCrypt/flowcrypt-ios/blob/6367703c35bfe3df860573e8b2b8b0393b64a545/FlowCrypt/Core/Core.swift#L47

```swift
    func version() throws -> CoreRes.Version {
        let r = try call("version", jsonDict: nil, data: nil)
        return try r.json.decodeJson(as: CoreRes.Version.self)
    }
```

it will call into javascript eg here:

```ts
  public version = async (): Promise<Buffers> => {
    return fmtRes({ app_version: VERSION });
  }
```

The `fmtRes` method formats a resulting JSON + optional data into a single string. After we dropped Android support and HTTP server support, there is no point in doing this. We can instead return an object like `{json:string, data: Uint8Array}` which Swift's `JavaScriptCore` can load directly instead of parsing it.

Here's the `fmtRes` implementation:

```ts
export const fmtRes = (response: {}, data?: Buf | Uint8Array): Buffers => {
  const buffers: Buffers = [];
  buffers.push(Buf.fromUtfStr(JSON.stringify(response)));
  buffers.push(Buf.fromUtfStr('\n'));
  if (data) {
    buffers.push(data);
  }
  return buffers;
}
``` 

I suppose it should change to something like this:

```ts
type EndpointRes = {json: strring, data: Buf | Uint8Array};

export const fmtRes = (response: {}, data?: Buf | Uint8Array): EndpointRes => {
  return {
    json: JSON.stringify(response)),
    data: data || new Uint8Array(0)
  }
}
``` 

Therefore the endpoints almost don't need to change:

```ts
  public version = async (): Promise<EndpointRes> => {
    return fmtRes({ app_version: VERSION });
  }
```

Then similarly, we should no longer try to parse it on Swift side. Currently done here: https://github.com/FlowCrypt/flowcrypt-ios/blob/6367703c35bfe3df860573e8b2b8b0393b64a545/FlowCrypt/Core/Core.swift#L198

Something like this:

```diff
    private func call(_ endpoint: String, jsonData: Data, data: Data) throws -> RawRes {
        try blockUntilReadyOrThrow()
        cb_last_value = nil
        jsEndpointListener!.call(withArguments: [endpoint, String(data: jsonData, encoding: .utf8)!, data.base64EncodedString(), cb_catcher!])
-       let b64response = cb_last_value![0] as! String
-       let rawResponse = Data(base64Encoded: b64response)!
-       let separatorIndex = rawResponse.firstIndex(of: 10)!
-       let resJsonData = Data(rawResponse[...(separatorIndex - 1)])
+       // now somehow access the object from Swift 
+       let resObject = cb_last_value![0] as! Dict<any> // pseudo code
+       let resJsonData = Data(resObject["json"] as! String) // pseudo code
+       let resBytesData = Data(resObject["data"] as! [Uint8]) // pseudo code
        let error = try? resJsonData.decodeJson(as: CoreRes.Error.self)
        if let error = error {
            let errMsg = "------ js err -------\nCore \(endpoint):\n\(error.error.message)\n\(error.error.stack ?? "no stack")\n------- end js err -----"
            logger.logError(errMsg)
            
            throw CoreError.init(coreError: error) ?? CoreError.exception(errMsg)
        }
-        return RawRes(json: resJsonData, data: Data(rawResponse[(separatorIndex + 1)...]))
+        return RawRes(json: resJsonData, data: Data(resBytesData)
    }
```

IvanPizhenko:
This PR changes format of the TS Core response. 

issue #508 - only the typescript part without rebuilding js and Swift

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/Core/source/entrypoint-bare.ts
+++ b/Core/source/entrypoint-bare.ts
@@ -21,7 +21,7 @@ global.handleRequestFromHost = (endpointName: string, request: string, data: str
       cb(formatBareOutput(fmtErr(new Error(`Unknown endpoint: ${endpointName}`))));
     } else {
       handler(JSON.parse(request), [Buf.fromBase64Str(data)])
-        .then(res => cb(formatBareOutput(Buf.concat(res))))
+        .then(res => cb(formatBareOutput(Buf.fromUtfStr(JSON.stringify(res)))))
         .catch(err => cb(formatBareOutput(fmtErr(err))));
     }
   } catch (err) {
--- a/Core/source/entrypoint-bare.ts
+++ b/Core/source/entrypoint-bare.ts
@@ -6,25 +6,23 @@
 
 import { Buf } from './core/buf';
 import { Endpoints } from './mobile-interface/endpoints';
-import { fmtErr } from './mobile-interface/format-output';
+import { EndpointRes, fmtErr } from './mobile-interface/format-output';
 
 declare const global: any;
 
 const endpoints = new Endpoints();
 
-const formatBareOutput = (res: Buf) => res.toBase64Str();
-
-global.handleRequestFromHost = (endpointName: string, request: string, data: string, cb: (b64response: string) => void): void => {
+global.handleRequestFromHost = (endpointName: string, request: string, data: string, cb: (response: EndpointRes) => void): void => {
   try {
     const handler = endpoints[endpointName];
     if (!handler) {
-      cb(formatBareOutput(fmtErr(new Error(`Unknown endpoint: ${endpointName}`))));
+      cb(fmtErr(new Error(`Unknown endpoint: ${endpointName}`)));
     } else {
       handler(JSON.parse(request), [Buf.fromBase64Str(data)])
-        .then(res => cb(formatBareOutput(Buf.fromUtfStr(JSON.stringify(res)))))
-        .catch(err => cb(formatBareOutput(fmtErr(err))));
+        .then(res => cb(res))
+        .catch(err => cb(fmtErr(err)));
     }
   } catch (err) {
-    cb(formatBareOutput(fmtErr(err)));
+    cb(fmtErr(err));
   }
 };
--- a/Core/source/mobile-interface/format-output.ts
+++ b/Core/source/mobile-interface/format-output.ts
@@ -125,13 +125,13 @@ export const fmtRes = (response: {}, data?: Buf | Uint8Array): EndpointRes => {
 }
 
 export const fmtErr = (e: any): Buf => {
-  const r = fmtRes({
+  const res = fmtRes({
     error: {
       message: String(e),
       stack: e && typeof e === 'object' ? e.stack || '' : ''
     }
   });
-  return Buf.fromUtfStr(r.json);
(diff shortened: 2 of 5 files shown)


Core/source/mobile-interface/format-output.ts
```diff
@@ -116,22 +117,22 @@ export const fmtContentBlock = (allContentBlocks: MsgBlock[]): { contentBlock: M
...
-}));
+export const fmtErr = (e: any): Buf => {
+  const res = fmtRes({
+    error: {
+      message: String(e),
+      stack: e && typeof e === 'object' ? e.stack || '' : ''
+    }
+  });
+  return Buf.fromUtfStr(res.json);
+}
```
tomholub, line 135: Please have this also return `EndpointRes`, so you would just straight `return fmtRes({...`

Core/source/test.ts
```diff
@@ -31,7 +31,7 @@ ava.default('version', async t => {
 });
 
 ava.default('generateKey', async t => {
-  const { json, data } = parseResponse(await endpoints.generateKey({ variant: 'curve25519', passphrase: 'riruekfhydekdmdbsyd', userIds: [{ email: 'a@b.com', name: 'Him' }] }));
+  const { json: json, data } = parseResponse(await endpoints.generateKey({ variant: 'curve25519', passphrase: 'riruekfhydekdmdbsyd', userIds: [{ email: 'a@b.com', name: 'Him' }] }));
```
tomholub, line 34: This change is probably not needed

Core/source/entrypoint-bare.ts
```diff
@@ -21,7 +21,7 @@ global.handleRequestFromHost = (endpointName: string, request: string, data: str
       cb(formatBareOutput(fmtErr(new Error(`Unknown endpoint: ${endpointName}`))));
     } else {
       handler(JSON.parse(request), [Buf.fromBase64Str(data)])
-        .then(res => cb(formatBareOutput(Buf.concat(res))))
+        .then(res => cb(formatBareOutput(Buf.fromUtfStr(JSON.stringify(res)))))
```
tomholub, line 24: Here I would like to straight return the result back to Swift. So it would be `.then(res => cb(res))` which I guess is just `.then(cb)` ? And the type of `cb: (b64response: string) => void` will change to `cb: (res: EndpointRes) => void`.

That way, we return EndpointRes directly back to swift, which no longer has to parse it (after Swift is updated).

tomholub (commented):
It's very nice that tests didn't need to be updated, except for the utils. Looks great. See comments below.

IvanPizhenko:
@tomholub Solved above notes.

tomholub (approved):
This looks good. I'm thinking to merge it (without rebuilding the ios js file) and then start another PR with the file rebuilt. Thanks!

PR #546, opened 2021-09-26
issue 508 JSCore response format updates - Swift side

### LINKED ISSUE #508: refactor Core response from string to string + byte array

tomholub:
This is a draft PR - I've only checked in master and run `cd Core && npm run-script deploy` which updated the javascript resource in iOS app. This will break the app, and I'll be assigning this PR to be finished on the Swift side, as per #508 

this will eventually close #508

- Does not need tests (refactor only, docs or internal changes) - once existing tests pass, we are good

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Core/Core.swift
+++ b/FlowCrypt/Core/Core.swift
@@ -34,7 +34,7 @@ final class Core: KeyDecrypter, CoreComposeMessageType {
 
     private var jsEndpointListener: JSValue?
     private var cb_catcher: JSValue?
-    private var cb_last_value: [Any]?
+    private var cb_last_value: (String, [UInt8])?
     private var vm = JSVirtualMachine()!
     private var context: JSContext?
     private dynamic var started = false
@@ -171,8 +171,6 @@ final class Core: KeyDecrypter, CoreComposeMessageType {
                 self.context!.evaluateScript(jsFileSrc)
                 self.jsEndpointListener = self.context!.objectForKeyedSubscript("handleRequestFromHost")
                 self.cb_catcher = self.context!.objectForKeyedSubscript("engine_host_cb_value_formatter")
-                let cb_last_value_filler: @convention(block) ([NSObject]) -> Void = { values in self.cb_last_value = values }
-                self.context!.setObject(unsafeBitCast(cb_last_value_filler, to: AnyObject.self), forKeyedSubscript: "engine_host_cb_catcher" as (NSCopying & NSObjectProtocol)?)
                 self.ready = true
                 self.logger.logInfo("JsContext took \(trace.finish()) to start")
                 completion()
@@ -184,6 +182,10 @@ final class Core: KeyDecrypter, CoreComposeMessageType {
         let response = try call("gmailBackupSearch", jsonDict: ["acctEmail": email], data: nil)
         let result = try response.json.decodeJson(as: GmailBackupSearchResponse.self)
         return result.query
+    }
+
+    func handleCallbackResult(json: String, data: [UInt8]) {
+        cb_last_value = (json, data)
     }
 
     // MARK: Private calls
@@ -199,18 +201,13 @@ final class Core: KeyDecrypter, CoreComposeMessageType {
         try blockUntilReadyOrThrow()
         cb_last_value = nil
         jsEndpointListener!.call(withArguments: [endpoint, String(data: jsonData, encoding: .utf8)!, data.base64EncodedString(), cb_catcher!])
-        let b64response = cb_last_value![0] as! String
-        let rawResponse = Data(base64Encoded: b64response)!
-        let separatorIndex = rawResponse.firstIndex(of: 10)!
-        let resJsonData = Data(rawResponse[...(separatorIndex - 1)])
-        let error = try? resJsonData.decodeJson(as: CoreRes.Error.self)
-        if let error = error {
-            let errMsg = "------ js err -------\nCore \(endpoint):\n\(error.error.message)\n\(error.error.stack ?? "no stack")\n------- end js err -----"
-            logger.logError(errMsg)
-            
-            throw CoreError.init(coreError: error) ?? CoreError.exception(errMsg)
+        guard
+            let resJsonData = cb_last_value?.0.data(using: .utf8),
+            let rawResponse = cb_last_value?.1
+        else {
+            throw CoreError.format("JavaScript callback response not available")
         }
-        return RawRes(json: resJsonData, data: Data(rawResponse[(separatorIndex + 1)...]))
+        return RawRes(json: resJsonData, data: Data(rawResponse))
     }
     
     private func blockUntilReadyOrThrow() throws {
--- a/FlowCrypt/Core/Core.swift
+++ b/FlowCrypt/Core/Core.swift
@@ -4,6 +4,7 @@
 
 import FlowCryptCommon
 import JavaScriptCore
+import Combine
 
 enum CoreError: Error, Equatable {
     case exception(String)
@@ -39,6 +40,8 @@ final class Core: KeyDecrypter, CoreComposeMessageType {
     private var context: JSContext?
     private dynamic var started = false
     private dynamic var ready = false
+    
+    private let queue = DispatchQueue(label: "com.flowcrypt.core", qos: .background)
 
     private lazy var logger = Logger.nested(in: Self.self, with: "Js")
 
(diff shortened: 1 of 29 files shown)


FlowCrypt/Core/Core.swift
```diff
@@ -171,8 +171,6 @@ final class Core: KeyDecrypter, CoreComposeMessageType {
                 self.context!.evaluateScript(jsFileSrc)
                 self.jsEndpointListener = self.context!.objectForKeyedSubscript("handleRequestFromHost")
                 self.cb_catcher = self.context!.objectForKeyedSubscript("engine_host_cb_value_formatter")
-                let cb_last_value_filler: @convention(block) ([NSObject]) -> Void = { values in self.cb_last_value = values }
-                self.context!.setObject(unsafeBitCast(cb_last_value_filler, to: AnyObject.self), forKeyedSubscript: "engine_host_cb_catcher" as (NSCopying & NSObjectProtocol)?)
```
tomholub, line 175: You managed to remove this mess - that looks good too.

tomholub (commented):
You're on a good track. If some tests started failing, please debug them. Looks to me this is viable.

ivan-ushakov:
Now all unit tests from `FlowCryptAppTests` are green but I'm not sure about correct error handling. You have different approaches to errors passing from JavaScript code to Swift.

For example you have general error result for any exception, like this:
```
export const fmtErr = (e: any): EndpointRes => {
  return fmtRes({
    error: {
      message: String(e),
      stack: e && typeof e === 'object' ? e.stack || '' : ''
    }
  });
}

global.handleRequestFromHost = (endpointName: string, request: string, data: string, cb: (response: EndpointRes) => void): void => {
  try {
    const handler = endpoints[endpointName];
    if (!handler) {
      cb(fmtErr(new Error(`Unknown endpoint: ${endpointName}`)));
    } else {
      handler(JSON.parse(request), [Buf.fromBase64Str(data)])
        .then(res => cb(res))
        .catch(err => cb(fmtErr(err)));
    }
  } catch (err) {
    cb(fmtErr(err));
  }
};
```

On the other hand you have errors inside decrypt methods like this:
```
return { success: false, error: { type: DecryptErrTypes.keyMismatch, message: 'Missing appropriate key' }, message: prepared.message, longids, isEncrypted };
```

And in current unit tests I need to map some `DecryptErrTypes` errors to `CoreError` types.

FlowCrypt/Core/Core.swift
```diff
@@ -60,6 +60,11 @@ final class Core: KeyDecrypter, CoreComposeMessageType {
 
     func decryptKey(armoredPrv: String, passphrase: String) throws -> CoreRes.DecryptKey {
         let r = try call("decryptKey", jsonDict: ["armored": armoredPrv, "passphrases": [passphrase]], data: nil)
+
+        if let error = try? r.json.decodeJson(as: CoreRes.Error.self) {
+            throw CoreError.exception(error.error.message)
+        }
```
tomholub, line 66: Can there be a generic one in `_call` like before? All of these errors have a consistent format `{"error": {"message": ".", "stack": "..."}}`. The original `_call` error handling was generic and had logging done too which is important.

The system there should be generic but flexible enough so that you could get finer grained errors out of it for individual methods, but only when needed. Elese you have to add the same error handlers / guards into every method here which is worse.

tomholub:
> Now all unit tests from `FlowCryptAppTests` are green but I'm not sure about correct error handling. You have different approaches to errors passing from JavaScript code to Swift.
> 
> For example you have general error result for any exception, like this:
> 
> ```
> export const fmtErr = (e: any): EndpointRes => {
>   return fmtRes({
>     error: {
>       message: String(e),
>       stack: e && typeof e === 'object' ? e.stack || '' : ''
>     }
>   });
> }
> 
> global.handleRequestFromHost = (endpointName: string, request: string, data: string, cb: (response: EndpointRes) => void): void => {
>   try {
>     const handler = endpoints[endpointName];
>     if (!handler) {
>       cb(fmtErr(new Error(`Unknown endpoint: ${endpointName}`)));
>     } else {
>       handler(JSON.parse(request), [Buf.fromBase64Str(data)])
>         .then(res => cb(res))
>         .catch(err => cb(fmtErr(err)));
>     }
>   } catch (err) {
>     cb(fmtErr(err));
>   }
> };
> ```
> 
> On the other hand you have errors inside decrypt methods like this:
> 
> ```
> return { success: false, error: { type: DecryptErrTypes.keyMismatch, message: 'Missing appropriate key' }, message: prepared.message, longids, isEncrypted };
> ```
> 
> And in current unit tests I need to map some `DecryptErrTypes` errors to `CoreError` types.

I see - I would put one generic error handler into `_call` which has a `message`, optional `stack` and then the full json (either directly or as a generic Dict).
That means all methods except for `parseDecryptMsg` and `decryptFile` don't need further error handling in them.

But in these two, you could add additional error handling that parses the json (or evaluates the Dict) to choose to raise a more appropriate error.

would that make sense?

If not, then you'd have to put an error handler in each method, which is less ideal.

tomholub:
Alternatively if you want, you could have one huge error handler in _call that discerns all of these errors including the decrypt errors into their appropriate swift counterparts, and then no more error handling in any of these methods. That would also be ok with me.

FlowCrypt/Core/Core.swift
ivan-ushakov: Yes you are right. I was thinking that you need to have strong typing like you have in `MsgBlock.DecryptErr.ErrorType`. Now I see that `CoreRes.Error` has `type` parameter too.

FlowCrypt/Core/Core.swift
```diff
@@ -211,18 +213,22 @@ final class Core: KeyDecrypter, CoreComposeMessageType {
...
-        let separatorIndex = rawResponse.firstIndex(of: 10)!
-        let resJsonData = Data(rawResponse[...(separatorIndex - 1)])
+        guard
+            let resJsonData = cb_last_value?.0.data(using: .utf8),
+            let rawResponse = cb_last_value?.1
+        else {
+            throw CoreError.format("JavaScript callback response not available")
+        }
+
         let error = try? resJsonData.decodeJson(as: CoreRes.Error.self)
```
ivan-ushakov, line 223: Now we have the same error handler for all methods in JavaScript.

tomholub: Yes but test are failing - please have a look further.

tomholub: added - try now

ivan-ushakov: I forgot to add `FlowCrypt/Resources/flowcrypt-ios-prod.js.txt` because I was thinking you generate it from sources

tomholub: Sorry about that - we don't have an automated process for that yet, for now it has to be manually updated with `npm run-script deploy` and checked into the repo.

Once we auto-generate it from sources during build, we'll definitely gitignore it, to avoid confusion.

tomholub:
@ivan-ushakov do you think this is ready to merge? If yes, then you can mark it ready for review.

ivan-ushakov:
Who could be added as reviewer here?

tomholub:
I'll review it myself, and force-merge if good.

tomholub (commented):
LGTM

tomholub:
@ivan-ushakov well done - good to see we don't have to pass it as a base64 string it after all.

PR #548, opened 2021-09-27
issue #426 make decryptedKey not nullable

### LINKED ISSUE #426: decryptedKey response from Core should not be nullable

sosnovsky:
This PR makes `decryptedKey` property of `DecryptKey` not nullable.

close #426  // if this PR closes an issue

- Tests added or updated - deleted `testNoDecryptedKey` test (as `decryptedKey` can't be `nil` now), updated `testDecryptKeyWithWrongPassPhrase` and `testDecryptKeyWithWrongPassPhrase` tests to handle updated `decryptedKey` property

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Pgp/KeyMethods.swift
+++ b/FlowCrypt/Functionality/Pgp/KeyMethods.swift
@@ -35,17 +35,13 @@ final class KeyMethods: KeyMethodsType {
                 return nil
             }
 
-            guard let decrypted = try? self.decrypter.decryptKey(armoredPrv: privateKey, passphrase: passPhrase) else {
+            do {
+                _ = try self.decrypter.decryptKey(armoredPrv: privateKey, passphrase: passPhrase)
+                return key
+            } catch {
                 logger.logInfo("Filtered not decrypted key")
                 return nil
             }
-
-            guard decrypted.decryptedKey != nil else {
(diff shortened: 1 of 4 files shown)


FlowCrypt/Functionality/Pgp/KeyMethods.swift
```diff
@@ -35,17 +35,13 @@ final class KeyMethods: KeyMethodsType {
...
             }
 
-            guard let decrypted = try? self.decrypter.decryptKey(armoredPrv: privateKey, passphrase: passPhrase) else {
+            do {
+                _ = try self.decrypter.decryptKey(armoredPrv: privateKey, passphrase: passPhrase)
+                return key
+            } catch {
                 logger.logInfo("Filtered not decrypted key")
                 return nil
             }
```
tomholub, line 44: Looks good, but we should be careful to specifically catch the "pass phrase mismatch" kind of error, and still do something meaningful with other kind of errors, at a minimum log them. (but better, show to user as a modal)

Filing for another issue.

tomholub (approved):

Issue #516, opened 2021-09-22
notReady - could not fetch keys


tomholub:
Another severe regression - this shows how lacking our tests are.

Can be tested on any account with EKM set up.

Reproduce:
1) set up account managed by EKM
2) open any encrypted message

![image](https://user-images.githubusercontent.com/6306961/134329811-c9dce7d6-b264-48c8-bcca-a802af43e43e.png)

ekievsky:
@tomholub while logging in to my flowcrypt gmail account receiving 502 on `https://fes.flowcrypt.com/api/`, no issues with usual gmail account

tomholub:
Yes, that is unrelated issue that I just caused minutes ago - I'm fixing it and will let you know once fixed.

ekievsky:
got you, just letting you know, thank you!

tomholub:
it should be ok now

tomholub:
This turns out to be an intermittent issue. It went away when I logged out and set up again. I'm moving to 0.2.1, but we have to investigate this thoroughly, it suggests we are either using Realm wrongly or not checking errors properly during setup.

tomholub:
Interestingly, when this was happening, it was showing 3 keys in Settings -> Keys without issue.

tomholub:
> [Roman] After app restart it works well - seems like something isn't fully configured after sign up

tomholub:
(I removed some comments that were not related to this particular issue, for clarity)

tomholub:
At least this seems to be quite reliably reproducible.

tomholub:
Also adding @ekievsky to have a look in parallel - this issue is becoming annoying.

sosnovsky:
@tomholub is there any way to simulate new user setup? As this bug happens only for new users, but isn't reproduced for existing users.

sosnovsky:
Yeah, I finally reproduced it by resetting Xcode simulator content, as after app uninstall-install it worked well without "fetch keys" error. Thanks!

sosnovsky:
I'm reproducing it with my `flowcrypt.com` mail, but I think it happens for other accounts too.

I found where this bug happens - https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/Functionality/Services/Key%20Services/KeyService.swift#L84

Stored passphrase has 2 fingerprints, but app compares only `primaryFingerprint`(which is the first one).
[...]

PR #552, opened 2021-09-27
Bugfix/issue 525 compose disappear

### LINKED ISSUE #525: compose content disappears on app blur
1) type text in compose view
2) put app into background
3) put app into foreground

actual result: composed message content disappears

expected result: composed message content should remain unchanged

tested with reply, probably problem in both

sosnovsky:
This PR fixes content disappearing when sending app to background while on Compose mail screen.
Issue affected not only Message text, but Subject too. It occurred also when rotating phone from portrait to landscape and back.

close #525

- Tests added or updated - added UI test for checking compose screen data persistence.

### To be filled by reviewers


tomholub (approved):
Thanks for the updated test! Looks good. I'll go ahead and test it manually too.

@Kharchevskyi if manual testing is successful, I'll merge it - later please see if you have any comments regarding code.

tomholub (changes requested):
New message works well, but reply doesn't - try the following scenarios:

-----------

1) open encrypted message
2) tap reply
expected: the quoted text shows right away
actual: quoted text only appears once I start typing

-----------------

1) open encrypted message
2) tap reply
3) start typing your message
4) tap home and then come back
expected: text is exactly as I left it
actual: text is empty

sosnovsky:
I tested first scenario - it's reproducible in `master` branch too, was caused by this change https://github.com/FlowCrypt/flowcrypt-ios/commit/c9788ca1ae718f8cc59d41f3a8188e54fafc54f0#diff-42fcab7325a0eb1205afe985324f4fd5e07f1e8add0ae0f759d92fa3c30787e4R342

tomholub:
Our lacking tests are showing again. @fcvakintos will start working on improving our UI tests and having them run on CI (he's starting later this week). If you could please also update the tests to cover this case, that would be great.

sosnovsky:
I fixed found issues and added ui test for checking reply text visibility.

tomholub:
Thanks! Could you please try to tackle #516 next? If you get any leads/clues, even if incomplete, please add a comment there, which may help others. It's quite a crippling bug.

tomholub (approved):
I can confirm all of these work well now. Thanks again!

PR #562, opened 2021-09-28
Use Folder model in Providers

### LINKED ISSUE #446: user in other db objects should not be nullable?
Is there a reason for nullable user here?

```
final class FolderObject: Object {
    convenience init(
        name: String,
        path: String,
        image: Data?,
        user: UserObject?
    ) {
        self.init()
        self.name = name
        self.path = path
        self.image = image
        self.user = user
    }
```

Since an imap folder is meaningless when it's not associated with any user.

Kharchevskyi:
This PR uses Folder instead of Realm FoldersObject in app providers.
Make UserObject in FolderObject not nullable

issue #446 

**Tests**:

### To be filled by reviewers


FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -697,11 +697,11 @@ extension ComposeViewController: UIImagePickerControllerDelegate, UINavigationCo
             attachment = ComposeMessageAttachment(librarySourceMediaInfo: info)
         default: fatalError("No other image picker's sources should be used")
         }
-        guard let attachment = attachment else {
+        guard let att = attachment else {
             showAlert(message: "files_picking_photos_error_message".localized)
             return
         }
-        appendAttachmentIfAllowed(attachment)
+        appendAttachmentIfAllowed(att)
```
tomholub, line 704: Going forward I'd like to spell it out as `attachment` instead of `att` because I found that developers new to the project may often read it `att` as `attribute` and be confused (I even had someone rename it to `attribute` in another codebase).

tomholub (commented):
Code looks good. Will test the app

tomholub (approved):

tomholub:
follow-up in #600

FlowCrypt/Controllers/Compose/ComposeViewController.swift
Kharchevskyi: Got it. Just had some problems with compiling code

PR #577, opened 2021-09-30
[code ok] issue 441 remove passphrase table

### LINKED ISSUE #441: remove separate pass phrase table, add as a column on keys table
It's an odd choice for a database schema to make user entered pass phrase string a primary key.

```swift
//
//  PassPhraseInfo.swift
//  FlowCrypt
//
//  Created by Anton Kharchevskyi on 04.06.2021.
//  Copyright © 2021 FlowCrypt Limited. All rights reserved.
//

import Foundation
import RealmSwift

/// PassPhrase object to store in Realm
final class PassPhraseObject: Object {
    @objc dynamic var value: String = ""
    let allFingerprints = List<String>()

    convenience init(
        value: String = "",
        fingerprints: [String]
    ) {
        self.init()
        self.value = value
        self.allFingerprints.append(objectsIn: fingerprints)
    }

    override class func primaryKey() -> String? {
        "value"
    }
}

// MARK: - Convenience
extension PassPhraseObject {
    var primaryFingerprint: String {
        allFingerprints[0]
    }

    convenience init(_ passPhrase: PassPhrase) {
        self.init(value: passPhrase.value, fingerprints: passPhrase.fingerprints)
    }
}

extension PassPhrase {
    init(object: PassPhraseObject) {
        self.value = object.value
        self.fingerprints = Array(object.allFingerprints)
        self.date = nil
    }
}
```

sosnovsky:
This PR removes PassPhraseObject model and adds optional `passphrase` field to `KeyInfo` class.

close #441

**Tests**:
- Tests added or updated - I removed tests for methods which are not used anymore, and also added check for passphrase in `testKeyInfoInit` test

### To be filled by reviewers


FlowCrypt/Controllers/Setup/SetupBackupsViewController.swift
```diff
@@ -144,17 +144,22 @@ extension SetupBackupsViewController {
...
 
-        // save pass phrase
-        matchingKeyBackups
-            .map {
-                PassPhrase(value: passPhrase, fingerprints: $0.fingerprints)
-            }
-            .forEach {
-                passPhraseService.savePassPhrase(with: $0, inStorage: shouldStorePassPhrase)
-            }
+        if !shouldStorePassPhrase {
```
tomholub, line 147: instead of these `shouldStorePassPhrase` and `inStorage` and similar booleans, we should use an enum like `StorageMethod.permanent` vs `StorageMethod.memory`, everywhere. That would be for another PR. Then here it will say `if storageMethod == StorageMethod.memory`

FlowCrypt/Functionality/Services/Key Services/InMemoryPassPhraseStorage.swift
```diff
@@ -70,7 +61,6 @@ final class InMemoryPassPhraseStorage: PassPhraseStorageType {
 protocol InMemoryPassPhraseProviderType {
     var passPhrases: Set<PassPhrase> { get }
     func save(passPhrase: PassPhrase)
-    func removePassPhrases(with objects: PassPhrase)
```
tomholub, line 73: `removePassPhrases` - I suppose this would be used during logout? Or is there some other way to throw away all pass phrases in memory? (also in the future we'll have a button for forgetting pass phrases)

tomholub (commented):
Still need to go through the rest of the code (checked about half) but it's looking good. Left some comments.

FlowCrypt/Controllers/Setup/SetupBackupsViewController.swift
sosnovsky: Agree, I like `enum` solution 👍 
I thought on possible improvements here, but decided that this pull request already has many changes and it'll be better to work on `inStorage` update later.

FlowCrypt/Functionality/Services/Key Services/InMemoryPassPhraseStorage.swift
sosnovsky: Before deleting I checked where `removePassPhrases` is used and it was used only in 1 place - https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/Functionality/Services/Key%20Services/PassPhraseService.swift#L86

So after removing `EncryptedStorage` I removed this method too, as it wasn't used anywhere else.
It seems currently we're not deleting passphrases from memory on logout. I'll rollback this change and call `removePassPhrases` on logout.

FlowCrypt/Functionality/Services/Key Services/KeyService.swift
```diff
@@ -100,19 +102,21 @@ final class KeyService: KeyServiceType {
...
         // append keys to ensure with a pass phrase
-        if let passPhrase = passPhrase {
-            let keysToEnsure = keysInfo.map {
-                PrvKeyInfo(
-                    private: $0.private,
-                    longid: $0.primaryLongid,
-                    passphrase: passPhrase,
-                    fingerprints: Array($0.allFingerprints)
-                )
+        let keysToEnsure: [PrvKeyInfo] = keysInfo.compactMap {
```
tomholub, line 105: The original wording here was unclear. The comment should be removed, and `keysToEnsure` should be called `keysWithPassPhrases` - in another PR.

tomholub (commented):
Other than earlier comment about forgetting pass phrases on logout, the code looks good. Will now do manual testing.

FlowCrypt/Functionality/Services/Key Services/InMemoryPassPhraseStorage.swift
sosnovsky: I added back `EncryptedStorage`, but now it uses passphrases stored in `KeyInfo` table.

For deleting in-memory passphrases I created separate task https://github.com/FlowCrypt/flowcrypt-ios/issues/587, as currently `UserAccountService` doesn't have direct reference to `InMemoryPassPhraseStorage` and I'll think of better handling for this case.

tomholub (commented):
Looks good, waiting for updated xcode to try

Issue #520, opened 2021-09-23
test for not needing pass phrase for plain message


tomholub:
Test steps:
1) set up with pass phrase in session, restart app (to clear the pass phrase)
2) open plain message

expected result: message renders without asking for pass phrase

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/519#issuecomment-925572274_

fcvakintos:
hi @tomholub please review https://github.com/FlowCrypt/flowcrypt-ios/pull/698

PR #589, opened 2021-10-01
issue #519 show missing passphrase alert only for encrypted messages

### LINKED ISSUE #519: pass phrase gets wrongly prompted even if message is not encrypted
pass phrase dialog should be prompted when you attempt decrypting and get back a `need_passphrase` error.

It looks like the request for pass phrase is either done unconditionally whenever not present, or the need for pass phrase is somehow guessed.

To reproduce:
1) set up with pass phrase in session, restart app
2) open plain message

actual result: I'm asked for pass phrase
expected result: message renders without asking

sosnovsky:
This PR fixes issue when missing passphrase alert was shown for plain messages.

close #519

**Tests**:
- Tests will be added later - issue #520 

### To be filled by reviewers


FlowCrypt/Functionality/Services/Key Services/KeyService.swift
```diff
@@ -68,51 +68,16 @@ final class KeyService: KeyServiceType {
```
tomholub, line 68: Please change this method signature to not accept any pass phrase. It should just itself return [PrvKeyInfo] with pass phrases optionally included (nil when not available for that key). Based on all pass phrases that are available in storage OR in memory.

tomholub:
Looks good! The code still confuses me a little bit but will look into that another day. Anyway the code clarity has improved from before. The functionality works well.

I managed to cause a crash when I entered pass phrase to view a message, then viewed the message and clicked the back button (crash seemed to happen during the transition to inbox). I couldn't reproduce it again, and didn't have a debugger attached to catch the stack trace.

PR #590, opened 2021-10-02
Convert backupToInbox method from Promise to Future

### LINKED ISSUE #543: convert backupToInbox method from Promise to Future
Maybe we could have a separate PR later to covert this `backupToInbox` to a Future. I'll file it.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/505#discussion_r715525076_

ivan-ushakov:
close #543

### To be filled by reviewers


ivan-ushakov:
I found that it will be much easy to have service methods implemented without any internal async. So we could decide how we run that method when we use it in chain.

ivan-ushakov:
I did mistake with queue dispatching and will fix it today.

FlowCrypt/Extensions/CombineExtensions.swift
```diff
@@ -25,3 +25,14 @@ extension Publisher {
             })
     }
 }
+
+extension Publishers.SubscribeOn {
+    // swiftlint:disable line_length
+    func myFlatMap<P>(_ transform: @escaping (Self.Output) -> P) -> Publishers.FlatMap<P, Publishers.SetFailureType<Self, P.Failure>> where P: Publisher {
```
ivan-ushakov, line 31: Publishers.SubscribeOn flatMap operator available only in iOS 14 and higher so I have to create this one

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
```diff
@@ -75,51 +77,113 @@ final class SetupGenerateKeyViewController: SetupCreatePassphraseAbstractViewCon
...
-
-            let updateKey = self.attester.updateKey(
-                email: userId.email,
-                pubkey: encryptedPrv.key.public,
-                token: self.storage.token
-            )
+        showSpinner()
+
+        cancellable = Just(passPhrase)
+            .subscribe(on: DispatchQueue.global())
```
ivan-ushakov, line 83: General idea is to control externally how to we want to run method. So method has no idea about queues and sync/async.

FlowCrypt/Extensions/CombineExtensions.swift
ivan-ushakov: It was another problem and with SO I found how to handle it correctly.

tomholub:
@Kharchevskyi does this look good to you? Please have a look.

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
```diff
@@ -75,51 +77,114 @@ final class SetupGenerateKeyViewController: SetupCreatePassphraseAbstractViewCon
...
-            )
+        showSpinner()
+
+        cancellable = Just(passPhrase)
+            .setFailureType(to: Error.self)
+            .subscribe(on: DispatchQueue.global())
+            .flatMap(validateAndConfirm)
+            .flatMap(generateKey)
+            .flatMap(backupKey)
+            .flatMap(processKey)
```
tomholub, line 88: I know this is a common pattern in the iOS world but I don't like to tie results of one method to inputs of other methods THIS closely. Where you have to have a `ProcessKeyInput` returned by backupKey - that doesn't make any sense to me. Each method should be individual, inputs and outputs should not be tied to the exact chain that we happen to call it in.

Notice how the original code was written. I'd like it written with the same pattern, but using `Future` instead of `Promise`. I'm actively trying to avoid this tight chaining pattern showed here.

Look at this code:

```
            let userId = try self.getUserId()

            try awaitPromise(self.validateAndConfirmNewPassPhraseOrReject(passPhrase: passPhrase))

            let encryptedPrv = try self.core.generateKey(passphrase: passPhrase, variant: .curve25519, userIds: [userId])

            try awaitPromise(self.backupService.backupToInbox(keys: [encryptedPrv.key], for: self.user))

            let passPhrase = PassPhrase(value: passPhrase, fingerprints: encryptedPrv.key.fingerprints)

            self.keyStorage.addKeys(keyDetails: [encryptedPrv.key], source: .generated, for: self.user.email)
            self.passPhraseService.savePassPhrase(with: passPhrase, inStorage: self.shouldStorePassPhrase)

            let updateKey = self.attester.updateKey(
                email: userId.email,
                pubkey: encryptedPrv.key.public,
                token: self.storage.token
            )
```

It's very clear what inputs and outputs of each method is. I want to preserve that.

tomholub (changes requested):

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
ivan-ushakov: In provided code you have a mix of sinc and async operations. This is the reason why `await` is used. Do you want to have the same pattern used in code with `Future`?

tomholub: Yes - I'd like to use a pattern that looks very similar to the original code, with minimal or no changes to the general structure of the code.

FlowCrypt/Controllers/Settings/Backup/Backups Option Scene/BackupOptionsViewController.swift
```diff
@@ -103,19 +106,26 @@ extension BackupOptionsViewController {
 
     private func backupToInbox() {
         showSpinner()
-        backupService.backupToInbox(keys: backups, for: userId)
-            .then(on: .main) { [weak self] in
+        cancellable = Just(backups)
```
Kharchevskyi, line 109: would it be better to use 
`backupService.backupToInbox(keys: input, for: userId)` and operate with this publisher directly instead of `Just(backups)`?

FlowCrypt/Controllers/Settings/Backup/Backups Seleckt Key Scene/BackupSelectKeyViewController.swift
```diff
@@ -85,19 +88,26 @@ extension BackupSelectKeyViewController {
             .filter { $0.1 == true }
             .map(\.0)
 
-        backupService.backupToInbox(keys: backupsToSave, for: userId)
-            .then(on: .main) { [weak self] in
+        cancellable = Just(backupsToSave)
```
Kharchevskyi, line 91: same here

tomholub:
Thank you Anton. I'm unsure if your comments still apply after the PR is reworked after my comments - let's see after the rework.

FlowCrypt/Controllers/Settings/Backup/Backups Option Scene/BackupOptionsViewController.swift
ivan-ushakov: No, because you need to tell Combine that all chain should run on global queue. If you run `backupService.backupToInbox()` first, it will run on current thread.

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
ivan-ushakov: OK, I reverted changed back and added wait method for Combine Publisher. But now I don't understand the goal of refactoring.

tomholub (approved):
for me, this is exactly what I was looking for. Thank you!

@Kharchevskyi could you please give this a second look now?

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
tomholub: :-) the goal is that we depracate the Google Promise library, and instead use a subset of Combine - just the `Future`. Right now, half the code is done with Promise and half with Future so we want to standardize on just writing async code with Future.

Does that make sense?

tomholub (dismissed):
I suppose Anton is busy. To me it looks good. I'll try the app as well, and merge it if works well.

FlowCrypt/Functionality/Services/Backup Services/BackupService.swift
```diff
@@ -45,53 +45,56 @@ extension BackupService: BackupServiceType {
...
-            guard let self = self else { throw AppErr.nilSelf }
-
-            let isFullyEncryptedKeys = keys.map(\.isFullyDecrypted).contains(false)
-
-            guard isFullyEncryptedKeys else {
-                throw BackupServiceError.keyIsNotFullyEncrypted
-            }
+    func backupToInbox(keys: [KeyDetails], for userId: UserId) -> Future<Void, Error> {
+        Future { [weak self] promise in
+            DispatchQueue.global().async {
```
Kharchevskyi, line 50: This is a responsibility for subscriber to decide on which queue to run this Future.
You can subscribe on a dedicated queue
`self.backupService.backupToInbox(keys: [encryptedPrv.key], for: self.user).subscribe(on: DispatchQueue.global()).wait()`

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
ivan-ushakov: Yes, now I understand what you need.

FlowCrypt/Controllers/Settings/Backup/Backups Seleckt Key Scene/BackupSelectKeyViewController.swift
```diff
@@ -85,14 +88,14 @@ extension BackupSelectKeyViewController {
             .filter { $0.1 == true }
             .map(\.0)
 
-        backupService.backupToInbox(keys: backupsToSave, for: userId)
-            .then(on: .main) { [weak self] in
+        cancellable = backupService.backupToInbox(keys: backupsToSave, for: userId)
```
Kharchevskyi, line 91: why not to store it in cancelable set instead of assigning it?
```
backupService.backupToInbox(keys: backupsToSave, for: userId)
            .receive(on: DispatchQueue.main)
            .sinkFuture(receiveValue: { [weak self] in
                self?.hideSpinner()
                self?.navigationController?.popToRootViewController(animated: true)
            }, receiveError: { [weak self] error in
                self?.handleCommon(error: error)
            })
            .store(in: &cancellable)
```

Kharchevskyi (changes requested):

FlowCrypt/Controllers/Settings/Backup/Backups Option Scene/BackupOptionsViewController.swift
Kharchevskyi: please check comments bellow

FlowCrypt/Functionality/Services/Backup Services/BackupService.swift
ivan-ushakov: Subscribe will work only for downstream, not upstream. You could check this with playground and following example:
```
func f1(_ value: Int) -> Future<Int, Error> {
    Future { promise in
        print("\(Thread.current): f1() value = \(value)")
        promise(.success(value + 1))
    }
}

func f2(_ value: Int) -> Future<Int, Error> {
    Future { promise in
        print("\(Thread.current): f2() value = \(value)")
        promise(.success(value + 1))
    }
}

let a2 = f1(0)
    .subscribe(on: DispatchQueue.global())
    .flatMap(f2)
    .receive(on: DispatchQueue.main)
    .sink { _ in
        print("\(Thread.current): completed")
    } receiveValue: { value in
        print("\(Thread.current): sink() value = \(value)")
    }
```

Output will be:
```
<_NSMainThread: 0x60000045c540>{number = 1, name = main}: f1() value = 0
<NSThread: 0x600000459d40>{number = 5, name = (null)}: f2() value = 1
<_NSMainThread: 0x60000045c540>{number = 1, name = main}: sink() value = 2
<_NSMainThread: 0x60000045c540>{number = 1, name = main}: completed
```

ivan-ushakov: So this is the reason why `Just` was used in my original code.

FlowCrypt/Controllers/Settings/Backup/Backups Seleckt Key Scene/BackupSelectKeyViewController.swift
ivan-ushakov: I guess this two are equivalent, or?

FlowCrypt/Functionality/Services/Backup Services/BackupService.swift
Kharchevskyi: What do you mean by `downstream, not upstream`? 
A Future will execute immediately when you create it.
If you need to postpone subscription use `Deffered` for this
https://developer.apple.com/documentation/combine/deferred

```
func f1(_ value: Int) -> AnyPublisher<Int, Error> {
    Deferred {
        Future { promise in
            sleep(1)
            let new = value + 1
            print("isMainThread \(Thread.isMainThread): f1() value = \(new)")
            promise(.success(new))
        }
    }
    .eraseToAnyPublisher()
}

func f2(_ value: Int) -> Future<Int, Error> {
    Future { promise in
        sleep(1)
        let new = value + 1
        print("isMainThread \(Thread.isMainThread): f2() value = \(value)")
        promise(.success(new))
    }
}

let a2 = f1(0)
    .subscribe(on: DispatchQueue.global())
    .flatMap(f2)
    .receive(on: DispatchQueue.main)
    .sink { _ in
        print("isMainThread \(Thread.isMainThread): completed")
    } receiveValue: { value in
        print("isMainThread \(Thread.isMainThread): sink() value = \(value)")
    }
```

FlowCrypt/Controllers/Settings/Backup/Backups Seleckt Key Scene/BackupSelectKeyViewController.swift
Kharchevskyi: actually the same, I was just wondering why you decide to assign it instead of chaining in `.store`

FlowCrypt/Functionality/Services/Backup Services/BackupService.swift
ivan-ushakov: So you want to add `Deffered` in any method? Looks like a hack.

FlowCrypt/Controllers/Settings/Backup/Backups Seleckt Key Scene/BackupSelectKeyViewController.swift
ivan-ushakov: I prefer to assign

FlowCrypt/Functionality/Services/Backup Services/BackupService.swift
tomholub: @sosnovsky do you have any thoughts or preference regarding the two approaches above?

tomholub: My own design goal here is that the resulting code structure in Swift reads similar to async/await syntax in TypeScript and Kotlin, because these are the tree languages we use at our company. Doing it similarly adds a bit of consistency across codebases, and makes it easier for team members to jump across codebases, and also makes PR reviews easier.

So that would be something like (TS example)

```ts
async function bigAsyncMethod(input): Promise<X> {
  const thing = await something(input);
  if (thing > 10) {
    return await thingWasbig()
  }
  return await thingWasSmall()
}
```

which is then called like

```ts
bigAsyncMethod('hello').then(success => {
  // happy path
}).catch(error => {
  // error path
})
```

With the PR as currently written, I think we've achieved that goal. What's left to do is to settle on a good-practice code pattern for these situations in Swift, as Ivan and Anton are discussing.

tomholub: As it looks now, I may slightly prefer the `Deferred` approach as it seems to make usage of the methods a bit more flexible. But I'm not in a position to know if it's a good idea or not, since Swift is not my area.

Kharchevskyi: Yes, basically there is a responsibility for the subscriber of the Future to decide on which thread this publisher should be executed. 
Combine in general is an SDK to handle async operations, so there is a misusage of it in case you manually switch to some thread inside implementation. This breaks SOLID principles as well
You can make some convenience extension for Publisher which wraps Defferd and Promise and use this instead

tomholub: >  You can make some convenience extension for Publisher which wraps Defferd and Promise and use this instead

Do you mean it would wrap `Deferred` and `Future` ? Roughly what could it look like?

ivan-ushakov: > My own design goal here is that the resulting code structure in Swift reads similar to async/await syntax in TypeScript and Kotlin, because these are the tree languages we use at our company.

True async/await is available only with iOS 15 and coroutines. My `wait` method blocks current thread while real `await` should not. I'm not sure it is possible to get the same behaviour as TypeScript/Kotlin has.

tomholub: > True async/await is available only with iOS 15 and coroutines. My wait method blocks current thread while real await should not. I'm not sure it is possible to get the same behaviour as TypeScript/Kotlin has.

I understand that, but you had to add the `.wait()` because you were mixing Promise and Future, right? Once we are all Future, you shouldn't have to do that?

Kharchevskyi:
```
extension Future {
    static func deferred(_ attempt: @escaping ((Result<Output, Failure>) -> Void) -> Void) -> AnyPublisher<Output, Failure> {
        return Deferred {
            Future.init(attempt)
        }
        .eraseToAnyPublisher()
    }
}
```

and usage
```
func f1(_ value: Int) -> AnyPublisher<Int, Error> {
    Future.deferred { promise in
        sleep(1)
        let new = value + 1
        print("isMainThread! \(Thread.isMainThread): f1() value = \(new)")
        promise(.success(new))
    }
}
```

Or any other solution, like extending AnyPublisher, creating your own Publisher - DefferedFuture. Or so...

ivan-ushakov: > Got it. Did you have to add the `.wait()` because you were mixing Promise and Future? Once we are all Future, you shouldn't have to do that?

Since you want to use `await` pattern and don't like to chain functions with `flatMap` (like I did in original code), we have to wait for future result somehow. So we have to use `wait`.

sosnovsky: @tomholub I usually prefer method chaining, as it was in original Ivan's code, it's quite common pattern for Swift. But after reading all comments here I think `wait` solution should be better for this case - to make code similar between different platforms. Also in the future it'll be easier to convert this code to native iOS `async/await`

tomholub: Thanks Roma

@Kharchevskyi is it not possible to put Deferred a the call point rather than into the method? It feels weird to have the first called method be "special" as you did it here: https://github.com/FlowCrypt/flowcrypt-ios/pull/590#discussion_r722482417 which would mean we would be putting Deferred inside all async method which seems also weird.

Also I want to avoid method chaining if possible.

So can it be something like this?

```swift
func f1(_ value: Int) -> Future<Int, Error> {
        Future { promise in
            sleep(1)
            let new = value + 1
            print("isMainThread \(Thread.isMainThread): f1() value = \(new)")
            promise(.success(new))
        }
}

func f2(_ value: Int) -> Future<Int, Error> {
    Future { promise in
        sleep(1)
        let new = value + 1
        print("isMainThread \(Thread.isMainThread): f2() value = \(value)")
        promise(.success(new))
    }
}
```

```Swift
// pseudo code
Future.deferred {
  let a = f1(0)
  let b = f2(a + 3)
}
.subscribe(on: DispatchQueue.global())
.receive(on: DispatchQueue.main)
.sink { _ in
  print("isMainThread \(Thread.isMainThread): completed")
} receiveValue: { value in
  print("isMainThread \(Thread.isMainThread): sink() value = \(value)")
}
```

Is there really no such thing before real async/await of iOS15? Do I have to wait every time and block the thread like this?

```swift
// pseudo code
Future.deferred {
  let a = f1(0).wait()
  let b = f2(a + 3).wait()
}
.subscribe(on: DispatchQueue.global())
.receive(on: DispatchQueue.main)
.sink { _ in
  print("isMainThread \(Thread.isMainThread): completed")
} receiveValue: { value in
  print("isMainThread \(Thread.isMainThread): sink() value = \(value)")
}
```

tomholub:
I may be able to arrange dropping support for iOS 14. That means we'd transition to iOS 15 async/await syntax and stop it with these workarounds. I'll have more news soon.

tomholub:
Not sure if still relevant

```
dyld4 config: DYLD_LIBRARY_PATH=/usr/lib/system/introspection DYLD_INSERT_LIBRARIES=/Developer/usr/lib/libBacktraceRecording.dylib:/Developer/usr/lib/libMainThreadChecker.dylib:/Developer/Library/PrivateFrameworks/DTDDISupport.framework/libViewDebuggerSupport.dylib
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Unexpected nil element for willDisplayCell: <_ASTableViewCell: 0x11e243990; baseClass = UITableViewCell; frame = (0 174.333; 320 0); autoresize = W; layer = <CALayer: 0x2800c9ba0>>, <ASTableView: 0x107a4d000; baseClass = UITableView; frame = (0 0; 320 693); clipsToBounds = YES; autoresize = W+H; gestureRecognizers = <NSArray: 0x280e0aee0>; layer = <FlowCryptUI.TableNode-Layer: 0x281595340; node = <FlowCryptUI.TableNode: 0x1079c1c00>>; contentOffset: {0, -87}; contentSize: {320, 263}; adjustedContentInset: {87, 0, 37, 0}; dataSource: <ASTableViewProxy: 0x2807e9e40>>, <NSIndexPath: 0x84260b1799598112> {length = 2, path = 1 - 3}'
terminating with uncaught exception of type NSException
(lldb) 
```

ivan-ushakov:
Could I continue now with this issue and try `async/await` approach?

tomholub:
> Could I continue now with this issue and try `async/await` approach?

Yes - please try that.

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
```diff
@@ -85,7 +85,12 @@ extension SetupGenerateKeyViewController {
 
             let encryptedPrv = try self.core.generateKey(passphrase: passPhrase, variant: .curve25519, userIds: [userId])
 
-            try awaitPromise(self.backupService.backupToInbox(keys: [encryptedPrv.key], for: self.user))
+            let semaphore = DispatchSemaphore(value: 0)
```
ivan-ushakov, line 88: This method uses a lot of other methods and too big for the first step. So I just wait for async task.

FlowCryptAppTests/Core/FlowCryptCoreTests.swift
```diff
@@ -93,104 +93,80 @@ class FlowCryptCoreTests: XCTestCase {
...
-        core.composeEmail(msg: msg, fmt: .plain, pubKeys: nil)
-            .sinkFuture(
-                receiveValue: { composeEmailRes in
-                    mime = String(data: composeEmailRes.mimeEncoded, encoding: .utf8)!
-                    expectation.fulfill()
-                }, receiveError: {_ in }
-            )
-            .store(in: &cancellable)
-        wait(for: [expectation], timeout: 3)
+    func testComposeEmailPlain() async throws {
```
ivan-ushakov, line 96: Test framework allows to use `async` keyword and simplify testing of async code. No need for expectations.

FlowCrypt.xcodeproj/project.pbxproj
```diff
@@ -99,7 +99,6 @@
 		9F2AC5B1267BDED100F6149B /* GmailSearchExpressionGenerator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9F2AC5B0267BDED100F6149B /* GmailSearchExpressionGenerator.swift */; };
 		9F2F206826AEEAA60044E144 /* CombineTestExtension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9F2F206726AEEAA60044E144 /* CombineTestExtension.swift */; };
 		9F2F207326AEECFB0044E144 /* PromiseTestExtension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9FC411892681191D004C0A69 /* PromiseTestExtension.swift */; };
-		9F2F212B26B305420044E144 /* test-ci-secrets.json in Resources */ = {isa = PBXBuildFile; fileRef = 9F2F212A26B305420044E144 /* test-ci-secrets.json */; };
```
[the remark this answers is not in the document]
tomholub: Why do you have to delete it? Any errors?

It's used for ui tests - contains credentials.

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
tomholub: Ok. Can do in another PR.

tomholub:
@ivan-ushakov anything left to do for this PR to be able to mark it as ready?

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -206,26 +206,30 @@ extension ComposeViewController {
...
+        case .failure(let error):
+            handle(error: error)
+        }
+    }
+
+    private func encryptAndSend(_ message: SendableMsg) {
+        Task {
+            do {
+                try await composeMessageService.encryptAndSend(message: message, threadId: input.threadId)
+                handleSuccessfullySentMessage()
```
ivan-ushakov, line 226: Since every view controller is marked with `@MainActor` we could simply call methods from task and they will be invoked on the main thread.

FlowCrypt.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
ivan-ushakov: Now I have this file locally.

tomholub:
@Kharchevskyi does this look good to you? If the tests pass I'll merge it, but if there is something to change, we let me know and we'll get it addressed in the next PR.

PR #591, opened 2021-10-02
[code ok] Disallow attester organisational rule

### LINKED ISSUE #299: implement disallow_attester_search_for_domains OrgRule
part of https://github.com/FlowCrypt/flowcrypt-ios/issues/275

When this is present, it will contain a list of domains (defaults to empty array when not set/nil), or possibly a star `*` which means disable Attester search completely (and only rely on WKD).

You'll want to re-implement this method in ClientConfiguration or a similar place:

https://github.com/FlowCrypt/flowcrypt-browser/blob/a6d0dc021e20c43e41f906aa8783c77f304e3521/extension/js/common/org-rules.ts#L170-L184

Then you'll use that method whenever fetching public keys from Attester. If this returns false, do not send any request to Attester and act as if Attester returned zero results. Best would be to edit the Attester methods, for example like done here: https://github.com/FlowCrypt/flowcrypt-browser/blob/3efa45c841808d9e9185ff9b38d1b78be10117f4/extension/js/common/api/key-server/attester.ts#L22-L28

Please be careful that - while the OrgRules belong to the sender - inside the above method goes email address **of the recipient**.

ekievsky:
This PR updates canLookupThisRecipientOnAttester OrgRule, ignores attester api depending on canLookupThisRecipientOnAttester

close #299

**Tests**:
- Difficult to test

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Services/AttesterApi.swift
+++ b/FlowCrypt/Functionality/Services/AttesterApi.swift
@@ -28,9 +28,14 @@ final class AttesterApi: AttesterApiType {
     }
 
     private let core: Core
+    private let organisationalRules: OrganisationalRules
 
-    init(core: Core = .shared) {
+    init(
+        core: Core = .shared,
+        organisationalRulesService: OrganisationalRulesServiceType = OrganisationalRulesService()
+    ) {
         self.core = core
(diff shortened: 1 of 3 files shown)


tomholub (commented):
looks good

PR #595, opened 2021-10-03
Rename ClientConfiguration

### LINKED ISSUE #504: join or rename OrganisationalRules with ClientConfiguration
In another PR, we should refactor the naming. It's not clear what's the benefit of having an `OrganisationalRules` object that only wraps `ClientConfiguration` object. We should just have a single `ClientConfiguration` object and that's all. If we do need a wrapper for some justified reason, then we can call that `ClientConfigurationWrapper` and add a comment on why it's needed. If it's to match it to a http response, then we should call it `ClientConfigurationResponse` that contains `ClientConfiguration`.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/499#discussion_r711189057_

ekievsky:
This PR renames ClientConfiguration->ClientConfigurationWrapper;

close #504

**Tests**:

### To be filled by reviewers


tomholub:
I'm sorry, this doesn't look right either. I'll have a look later.

ekievsky:
I'm not really sure what's the goal of that. Let's discuss and find the better way.

PR #615, opened 2021-10-05
Don't add .pgp to file list when attaching files on compose screen

### LINKED ISSUE #538: don't add .pgp to file list when attaching files on compose screen
(when composing a message and adding attachment)

the .pgp filename should not be visible in that list. So you don't have to alter the filename, after all (the encryption method should be already transparently renaming them)

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/505#pullrequestreview-763195500_

ekievsky:
This PR adds .pgp extension to attachments right before sending them

close #538

**Tests** :

### To be filled by reviewers


tomholub (approved):
Code looks good

PR #625, opened 2021-10-06
issue #609 TypeScript-Core functionality to verify signatures

### LINKED ISSUE #609: TypeScript-Core functionality to verify signatures
`parseDecryptMsg` currently accepts these values: `{ keys: PrvKeyInfo[], msgPwd: string?, isEmail: boolean }`

I'd like to add `verificationPubkeys: string[]` which will accept an array of armored public keys.

This will get passed into:

```
const decryptRes = await PgpMsg.decrypt({ kisWithPp, msgPwd, encryptedData: Buf.with(rawBlock.content) });
// which is
public static decrypt: PgpMsgMethod.Decrypt = async ({ kisWithPp, encryptedData, msgPwd }) => {
``` 

in the `PgpMsg.decrypt` method you need to edit `const keys = await PgpMsg.getSortedKeys(kisWithPp, prepared.message);` to accept this array as well (optionally - can be undefined), which means updating the signature of `private static getSortedKeys = async (kiWithPp: PrvKeyInfo[], msg: OpenpgpMsgOrCleartext): Promise<SortedKeysForDecrypt> => {` to accept it.

Then finally in `getSortedKeys`, this line:

```
await PgpMsg.cryptoMsgGetSignedBy(msg, keys);
```

should be changed to something like:

```ts
if(typeof verificationPubkeys !== 'undefined') {
  keys.forVerification = [];
  for (const verificationPubkey of verificationPubkeys) {
    const { keys: keysForVerification } = await openpgp.key.readArmored(verificationPubkey);
    keys.forVerification.push(...keysForVerification);
  }
} else {
  await PgpMsg.cryptoMsgGetSignedBy(msg, keys);
}
```

Then please remove everything related to `keys.verificationContacts`, which includes changing singature of `PpgMsg.verify` to not accept a `Contact?` and not return it either.

IvanPizhenko:
This PR refactors Typescript part in order to prepare to signature verification functionality.

close #609 

- Tests added or updated

### To be filled by reviewers


Core/source/core/pgp-msg.ts
```diff
@@ -301,19 +301,18 @@ export class PgpMsg {
   private static cryptoMsgGetSignedBy = async (msg: OpenpgpMsgOrCleartext, keys: SortedKeysForDecrypt) => {
     keys.signedBy = Value.arr.unique(await PgpKey.longids(msg.getSigningKeyIds ? msg.getSigningKeyIds() : []));
     if (keys.signedBy.length && typeof Store.dbContactGet === 'function') {
-      const verificationContacts = await Store.dbContactGet(undefined, keys.signedBy);
-      keys.verificationContacts = verificationContacts.filter(contact => contact && contact.pubkey) as Contact[];
+      const signerContacts = (await Store.dbContactGet(undefined, keys.signedBy))
+        .filter(contact => contact && contact.pubkey) as Contact[];
```
tomholub, line 305: I overlooked this. Anything using `dbContactGet` won't work on iOS because that method is not implemented (and will remain that way). 

you could pass `verificationPubkeys` also into `cryptoMsgGetSignedBy`, and then I suppose you can remove the if/else in `getSortedKeys`. So `cryptoMsgGetSignedBy` would fill both `keys.signedBy` (ids) and `keys.forVerification` (the supplied ones). Then since pubkeys are supplied, you remove the `dbContactGet` here, and also the `typeof Store.dbContactGet === 'function'`.

tomholub (commented):
looking good

Core/source/core/pgp-msg.ts
tomholub: Yes, I suppose you can simplify to let it go through cryptoMsgGetSignedBy and pass verificationPubkeys in

tomholub (commented):
👍

IvanPizhenko:
So I think I've finished with changes in the code, now just need to test it all.

Core/source/mobile-interface/endpoints.ts
```diff
@@ -90,20 +90,21 @@ export class Endpoints {
     } else {
       rawBlocks.push(MsgBlock.fromContent('encryptedMsg', new Buf(Buf.concat(data))));
     }
+    for (const rawBlock of rawBlocks) console.log(JSON.stringify(rawBlock) + '\n');
```
tomholub, line 93: pls remove

Core/source/mobile-interface/format-output.ts
```diff
@@ -71,7 +72,9 @@ export const fmtContentBlock = (allContentBlocks: MsgBlock[]): { contentBlock: M
       imgsAtTheBottom.push(plainImgBlock);
     }
   }
+  var verifyRes: (VerifyRes | undefined) = undefined;
   for (const block of contentBlocks) {
+    if (!verifyRes && block.verifyRes) verifyRes = block.verifyRes;
```
tomholub, line 77: please use multiline with brakets

Core/source/mobile-interface/format-output.ts
```diff
@@ -114,7 +117,9 @@ export const fmtContentBlock = (allContentBlocks: MsgBlock[]): { contentBlock: M
     </head>
     <body>${msgContentAsHtml}</body>
   </html>`;
-  return { contentBlock: MsgBlock.fromContent('plainHtml', msgContentAsHtml), text: msgContentAsText.trim() };
+  const contentBlock = MsgBlock.fromContent('plainHtml', msgContentAsHtml);
+  contentBlock.verifyRes = verifyRes;
```
tomholub, line 121: The logic should probably be a bit more complicated, for safety.

 - if no input block was signed, then no signature on output block
 - else, if any of the input blocks contains an invalid signature, then that signature should be on the output block
 - else, if the input blocks were signed by a mixture of public keys, then pick the first one and add `mixed = true` to verifyRes (need to update type)
 - else, (at this point all available signatures are valid and from the same key) if only some input blocks were signed, add `partial = true` to verifyRes (need to update type)
 - else (at this point all are signed successfully by the same key, and no input blocks are missing signatures) pick the first verifyRes from input to pass to output

tomholub (commented):
Looks great - see comments

tomholub (approved):
This looks good. Thanks! Since this only updates TS and TS tests pass, I'll force-merge this.

PR #644, opened 2021-10-07
Reply to message on Gmail was wrongly not in the same thread

### LINKED ISSUE #492: reply to message on Gmail was wrongly not in the same thread
In my case, the following happened. I didn't try to reproduce it again.

1) I received a plain message
2) I replied to the plain message from iOS app (encrypted)
3) the resulting message was received on the other end (on Gmail)

expected: the reply should have been attached to the same thread

actual: Gmail regarded the reply as the first message of a new thread

ivan-ushakov:
close #492

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Core/CoreTypes.swift
+++ b/FlowCrypt/Core/CoreTypes.swift
@@ -139,7 +139,6 @@ struct SendableMsg: Equatable {
     let replyToMimeMsg: String?
     let atts: [Attachment]
     let pubKeys: [String]?
-    let threadId: String?
 }
 
 struct MsgBlock: Decodable {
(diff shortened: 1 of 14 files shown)


FlowCrypt/Core/CoreTypes.swift
```diff
@@ -139,6 +139,7 @@ struct SendableMsg: Equatable {
     let replyToMimeMsg: String?
     let atts: [Attachment]
     let pubKeys: [String]?
+    let threadId: String?
```
tomholub, line 142: I don't think this one needs it. It would only be meaningful if the TS code received and used it on the other end, which afaik it doesn't.

tomholub: please try that - it would be confusing to have this be passed into composeEmail where it's not used

ivan-ushakov: `threadId` added as argument

ivan-ushakov:
Looks like `iOS Tests` failed because of timeout/interruption.

tomholub:
You can try to re-run them if that's the case, in the SemaphoreCI pipeline ui.

PR #647, opened 2021-10-09
Feature/issue 495 mark message as unread

### LINKED ISSUE #495: UX - Mark message as unread

ekievsky:
This PR allows user to mark message as unread

close #495

**Tests** :

### To be filled by reviewers


FlowCrypt/Controllers/Msg/MessageViewController.swift
```diff
@@ -266,7 +265,16 @@ extension MessageViewController {
     }
 
     @objc private func handleMailTap() {
```
tomholub, line 267: Instead of `Mail` here we should call it `MarkUnread` so this would be `handleMarkUnreadTap`. If the icon is elsewhere called just `Mail` in the code, than that should be updated too.

FlowCrypt/Controllers/Msg/MessageViewController.swift
```diff
@@ -266,7 +265,16 @@ extension MessageViewController {
...
-        showToast("Marking as unread will be implemented soon")
+        messageOperationsProvider.markAsUnread(message: input.objMessage, folder: input.path)
+            .then(on: .main) { [weak self] in
+                guard let self = self else { return }
+                self.input.objMessage = self.input.objMessage.markAsRead(false)
+                self.onCompletion?(MessageAction.markAsUnread, self.input.objMessage)
```
tomholub, line 272: I suppose `MessageAction.markAsUnread` could be updated to `MessageAction.changeReadFlag` ? Since it's now used for both marking as read and unread? Unless I'm wrong?

FlowCrypt/Functionality/Mail Provider/MessageOperations Provider/Imap+MessageOperations.swift
```diff
@@ -39,6 +40,31 @@ extension Imap: MessageOperationsProvider {
...
+    // MARK: - unread
+    func markAsUnread(message: Message, folder: String) -> Promise<Void> {
+        Promise { [weak self] resolve, reject in
+            guard let self = self else { return reject(AppErr.nilSelf) }
+
+            guard let identifier = message.identifier.intId else {
+                return reject(ImapError.missedMessageInfo("intId"))
+            }
+
+            var flags: MCOMessageFlag = []
```
tomholub, line 52: Could call `flagsToRemove` for clarity.

tomholub (commented):
Code looks good with some comments, will test soon

sosnovsky (approved):
@tomholub implementation looks good, but I had some difficulties with finding how to mark mail as unread, as 'email' icon can have many different meanings.

iOS includes built-in set of icons (SF Icons) and it has many different icons for mail, maybe we can choose a better one for "mark as unread" action. We can also use icons from this set for other actions and buttons, so app will look more native to iOS. What do you think?
Screenshot 2021-10-11 at 22 28 41

Also it's quite common for iOS mail apps to have swipe actions on messages list screen, I think Flowcrypt users will find it useful. We can add "Unread" action there too.

PR #650, opened 2021-10-09
Sign outgoing messages

### LINKED ISSUE #274: sign outgoing messages
part of  #275

When outgoing message is encrypted, it should also be signed with the private key.

The implementation is to add a private key into the `Core.composeEmail` function which will cause the message to also get signed.

Currently it doesn't have any way to add private key, so we should add `signingPrv: PrvKeyInfo?` to it and pass that in.

```swift
    func composeEmail(msg: SendableMsg, fmt: MsgFmt, pubKeys: [String]?) throws -> CoreRes.ComposeEmail {
        let r = try call("composeEmail", jsonDict: [
            "text": msg.text,
            "to": msg.to,
            "cc": msg.cc,
            "bcc": msg.bcc,
            "from": msg.from,
            "subject": msg.subject,
            "replyToMimeMsg": msg.replyToMimeMsg,
            "atts": msg.atts.map { att in ["name": att.name, "type": att.type, "base64": att.base64] },
            "format": fmt.rawValue,
            "pubKeys": pubKeys,
        ], data: nil)
        // this call returned no useful json data, only bytes
        return CoreRes.ComposeEmail(mimeEncoded: r.data)
    }
```

On the receiving side, this has to be put into the `PgpMsg.encrypt` function which has a way to accept a private key.

```ts
  public composeEmail = async (uncheckedReq: any): Promise<Buffers> => {
    const req = ValidateInput.composeEmail(uncheckedReq);
    const mimeHeaders: RichHeaders = { to: req.to, from: req.from, subject: req.subject, cc: req.cc, bcc: req.bcc };
    if (req.replyToMimeMsg) {
      const previousMsg = await Mime.decode(Buf.fromUtfStr((req.replyToMimeMsg.substr(0, 10000).split('\n\n')[0] || '') + `\n\nno content`));
      const replyHeaders = Mime.replyHeaders(previousMsg);
      mimeHeaders['in-reply-to'] = replyHeaders['in-reply-to'];
      mimeHeaders['references'] = replyHeaders['references'];
    }
    if (req.format === 'plain') {
      const atts = (req.atts || []).map(({ name, type, base64 }) => new Att({ name, type, data: Buf.fromBase64Str(base64) }));
      return fmtRes({}, Buf.fromUtfStr(await Mime.encode({ 'text/plain': req.text }, mimeHeaders, atts)));
    } else if (req.format === 'encrypt-inline') {
      const encryptedAtts: Att[] = [];
      for (const att of req.atts || []) {
        const encryptedAtt = await PgpMsg.encrypt({ pubkeys: req.pubKeys, data: Buf.fromBase64Str(att.base64), filename: att.name, armor: false }) as OpenPGP.EncryptBinaryResult;
        encryptedAtts.push(new Att({ name: att.name, type: 'application/pgp-encrypted', data: encryptedAtt.message.packets.write() }))
      }
      const encrypted = await PgpMsg.encrypt({ pubkeys: req.pubKeys, data: Buf.fromUtfStr(req.text), armor: true }) as OpenPGP.EncryptArmorResult;
      return fmtRes({}, Buf.fromUtfStr(await Mime.encode({ 'text/plain': encrypted.data }, mimeHeaders, encryptedAtts)));
    } else {
      throw new Error(`Unknown format: ${req.format}`);
    }
  }
```

ivan-ushakov:
close #274

- Tests will be added later (issue #659)

### To be filled by reviewers


ivan-ushakov:
Test are broken because I need to know if I'm moving in the right direction first. Please take a look.

tomholub:
when it's a WIP you should be marking it as a draft:

image

Core/source/mobile-interface/endpoints.ts
```diff
@@ -64,8 +64,13 @@ export class Endpoints {
...
         encryptedAtts.push(new Att({ name: att.name, type: 'application/pgp-encrypted', data: encryptedAtt.message.packets.write() }))
       }
-      const encrypted = await PgpMsg.encrypt({ pubkeys: req.pubKeys, data: Buf.fromUtfStr(req.text), armor: true }) as OpenPGP.EncryptArmorResult;
-      return fmtRes({}, Buf.fromUtfStr(await Mime.encode({ 'text/plain': encrypted.data }, mimeHeaders, encryptedAtts)));
+      const signingPrv = await readArmoredKeyOrThrow(req.signingPrv.private);
+      if (await PgpKey.decrypt(signingPrv, req.signingPrv.passphrase || '')) {
+        const encrypted = await PgpMsg.encrypt({ pubkeys: req.pubKeys, signingPrv: signingPrv, data: Buf.fromUtfStr(req.text), armor: true }) as OpenPGP.EncryptArmorResult;
+        return fmtRes({}, Buf.fromUtfStr(await Mime.encode({ 'text/plain': encrypted.data }, mimeHeaders, encryptedAtts)));
+      } else {
+        throw new Error(`Unknown format: ${req.format}`);
```
tomholub, line 72: The responded error here is wrong - this is not a format issue

Core/source/mobile-interface/endpoints.ts
```diff
@@ -64,8 +64,13 @@ export class Endpoints {
         const encryptedAtt = await PgpMsg.encrypt({ pubkeys: req.pubKeys, data: Buf.fromBase64Str(att.base64), filename: att.name, armor: false }) as OpenPGP.EncryptBinaryResult;
         encryptedAtts.push(new Att({ name: att.name, type: 'application/pgp-encrypted', data: encryptedAtt.message.packets.write() }))
       }
-      const encrypted = await PgpMsg.encrypt({ pubkeys: req.pubKeys, data: Buf.fromUtfStr(req.text), armor: true }) as OpenPGP.EncryptArmorResult;
-      return fmtRes({}, Buf.fromUtfStr(await Mime.encode({ 'text/plain': encrypted.data }, mimeHeaders, encryptedAtts)));
+      const signingPrv = await readArmoredKeyOrThrow(req.signingPrv.private);
+      if (await PgpKey.decrypt(signingPrv, req.signingPrv.passphrase || '')) {
+        const encrypted = await PgpMsg.encrypt({ pubkeys: req.pubKeys, signingPrv: signingPrv, data: Buf.fromUtfStr(req.text), armor: true }) as OpenPGP.EncryptArmorResult;
```
tomholub, line 69: `signingPrv: signingPrv` can be written simply as `signingPrv`, no need to duplicate the name.

Core/source/mobile-interface/endpoints.ts
```diff
@@ -64,8 +64,13 @@ export class Endpoints {
         const encryptedAtt = await PgpMsg.encrypt({ pubkeys: req.pubKeys, data: Buf.fromBase64Str(att.base64), filename: att.name, armor: false }) as OpenPGP.EncryptBinaryResult;
         encryptedAtts.push(new Att({ name: att.name, type: 'application/pgp-encrypted', data: encryptedAtt.message.packets.write() }))
       }
-      const encrypted = await PgpMsg.encrypt({ pubkeys: req.pubKeys, data: Buf.fromUtfStr(req.text), armor: true }) as OpenPGP.EncryptArmorResult;
-      return fmtRes({}, Buf.fromUtfStr(await Mime.encode({ 'text/plain': encrypted.data }, mimeHeaders, encryptedAtts)));
+      const signingPrv = await readArmoredKeyOrThrow(req.signingPrv.private);
```
tomholub, line 67: Looks ok except the `signingPrv` should be optional. If I supply it, message gets signed. If I don't, it won't.

Core/source/mobile-interface/validate-input.ts
```diff
@@ -50,7 +50,7 @@ export class ValidateInput {
     if (!hasProp(v, 'atts', 'Attachment[]?')) {
       throw new Error('Wrong atts structure for NodeRequest.composeEmail, need: {name, type, base64}');
     }
-    if (hasProp(v, 'pubKeys', 'string[]') && v.pubKeys.length && (v.format === 'encrypt-inline' || v.format === 'encrypt-pgpmime')) {
+    if (hasProp(v, 'pubKeys', 'string[]') && v.pubKeys.length && hasProp(v, 'signingPrv', 'PrvKeyInfo') && (v.format === 'encrypt-inline' || v.format === 'encrypt-pgpmime')) {
```
tomholub, line 53: should be optional

Core/source/mobile-interface/validate-input.ts
```diff
@@ -11,7 +11,7 @@ export namespace NodeRequest {
   type Attachment = { name: string; type: string; base64: string };
   interface composeEmailBase { text: string, to: string[], cc: string[], bcc: string[], from: string, subject: string, replyToMimeMsg: string, atts?: Attachment[] };
   export interface composeEmailPlain extends composeEmailBase { format: 'plain' };
-  export interface composeEmailEncrypted extends composeEmailBase { format: 'encrypt-inline' | 'encrypt-pgpmime', pubKeys: string[] };
+  export interface composeEmailEncrypted extends composeEmailBase { format: 'encrypt-inline' | 'encrypt-pgpmime', pubKeys: string[], signingPrv: PrvKeyInfo };
```
tomholub, line 14: optional

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -194,22 +197,44 @@ extension ComposeViewController {
...
 
 // MARK: - Message Sending
 
 extension ComposeViewController {
-    private func sendMessage() {
+    private func prepareSigningKey() {
+        guard let key = try? keyService.getPrvKeyInfo().get().first else {
+            showAlert(message: "No available private key has your user id \"\(email)\" in it. Please import the appropriate private key.")
+            return
+        }
```
tomholub, line 211: Thi does not follow the logic I outlined. You're supposed to evaluate all available private keys, and choose the one that has the appropriate OpenPGP user id encoded in it, to match the sending email, as described in the issue.

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -194,22 +197,44 @@ extension ComposeViewController {
...
+
+        guard key.passphrase != nil else {
+            let alert = AlertsFactory.makePassPhraseAlert(
```
tomholub, line 214: Does `makePassPhraseAlert` also check that the pass phrase is correct for the key? (I don't have the code in front of me)

Core/source/mobile-interface/endpoints.ts
ivan-ushakov: This is true only for the TypeScript side or also for the Swift side?

ivan-ushakov:
@tomholub 
We have following code:
```
  private static cryptoMsgGetSignedBy = async (msg: OpenpgpMsgOrCleartext, keys: SortedKeysForDecrypt) => {
    keys.signedBy = Value.arr.unique(await PgpKey.longids(msg.getSigningKeyIds ? msg.getSigningKeyIds() : []));
    if (keys.signedBy.length && typeof Store.dbContactGet === 'function') {
      const verificationContacts = await Store.dbContactGet(undefined, keys.signedBy);
      keys.verificationContacts = verificationContacts.filter(contact => contact && contact.pubkey) as Contact[];
      keys.forVerification = [];
      for (const contact of keys.verificationContacts) {
        const { keys: keysForVerification } = await openpgp.key.readArmored(contact.pubkey!);
        keys.forVerification.push(...keysForVerification);
      }
    }
  }
```
but `Store.dbContactGet` always empty:
```
  static dbContactGet = async (db: void, emailOrLongid: string[]): Promise<(Contact | undefined)[]> => {
    return [];
  }
```

tomholub:
> @tomholub How to know if signature was added to message? I did some debug and `Message.prototype.sign` is invoked when I provide `signingPrv`, but I don't see signature in result MIME message. Maybe because we use default compression.

It's because the message is armored, so you wouldn't see the signature be human-readable.

Open the message in browser extension, and see if it's showing as signed.

tomholub:
> @tomholub We have following code:
> 
> ```
>   private static cryptoMsgGetSignedBy = async (msg: OpenpgpMsgOrCleartext, keys: SortedKeysForDecrypt) => {
>     keys.signedBy = Value.arr.unique(await PgpKey.longids(msg.getSigningKeyIds ? msg.getSigningKeyIds() : []));
>     if (keys.signedBy.length && typeof Store.dbContactGet === 'function') {
>       const verificationContacts = await Store.dbContactGet(undefined, keys.signedBy);
>       keys.verificationContacts = verificationContacts.filter(contact => contact && contact.pubkey) as Contact[];
>       keys.forVerification = [];
>       for (const contact of keys.verificationContacts) {
>         const { keys: keysForVerification } = await openpgp.key.readArmored(contact.pubkey!);
>         keys.forVerification.push(...keysForVerification);
>       }
>     }
>   }
> ```
> 
> but `Store.dbContactGet` always empty:
> 
> ```
>   static dbContactGet = async (db: void, emailOrLongid: string[]): Promise<(Contact | undefined)[]> => {
>     return [];
>   }
> ```

This is for signature verification - you shouldn't concern yourself with that. Ivan P. is working on that in #274 . For this issue, only message signing.

FlowCrypt/Controllers/Compose/ComposeViewController.swift
tomholub: There are things to filter on database level and then also on openpgp-metadata level. On database level, we have already been doing it - filter by the key belonging to the currently signed-in user. That gives you one or more keys (don't just pick .first)

Then, on that list of keys, you need to filter by primary and secondary OpenPGP user ids. You'll have to run `prvKeyInfo.private` through `Core.parseKeys`. Put the resulting `KeyDetails` into one big array. On each KeyDetails, evaluate `keyDetails.users` - that's what you'll be using for filtering based on my above comments. You can add a computed variable `.primaryPgpUserEmail` which is the email address of the first `user` (technically could be nil). And a computed variable `pgpUserEmails` which are email addresses of all `users`. 

This is assuming that `users` contain something like `["Joe <joe@example.com>", "Bill <bill@example.com>"]`. Which will each have to be parsed to `{name?: "Joe", "email": "joe@example.com" }` - probably our IMAP library will have a parser for this.

FlowCrypt/Functionality/Services/Key Services/KeyService.swift
```diff
@@ -80,4 +81,48 @@ final class KeyService: KeyServiceType {
 
         return .success(privateKeys)
     }
+
+    func signingKey() throws -> PrvKeyInfo? {
```
tomholub, line 85: Good to have it implemented this way, in its own method. Unfortunately I cannot read the code so well - could you take a chunk of the code and write it as a private method and name it meaningfully? Also more meaningful names than `result`. And finally, we should write unit tests for this method to make sure it behaves per the logic we expected.

Otherwise looks good overall.

ivan-ushakov: Unfortunately it is not so easy to have a unit test here because `KeyService` has some dependencies like Realm objects and Core service. I did some refactoring to make code more readable.

tomholub: I think the tests will be necessary. Let's do that later in a separate PR later. There are several tests that need to be written:
 - ui test to check proper behavior when user inputs wrong pass phrase for message signing
 - ui test to check proper behavior when user eventually inputs correct pass phrase
 - ui test to check proper behavior when pass phrase already in memory
 - unit tests to make sure that the right private key is chosen in various scenarios

FlowCrypt/Functionality/Services/Key Services/KeyService.swift
```diff
@@ -82,47 +82,47 @@ final class KeyService: KeyServiceType {
...
+                armoredOrBinary: keyInfo.`private`.data()
+            )
+
+            var keyDetails: [KeyDetails] = []
+            parsedKeys.keyDetails.forEach { keyDetails.append($0) }
+
+            return keyDetails.first {
+                let emails = $0.pgpUserEmails
+                return emails.first == email || emails.contains(where: { $0 == email })
+            }
```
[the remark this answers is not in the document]
ivan-ushakov: Yes, my bad. Now implementation should work right. Also I added unit test.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -128,19 +128,16 @@ final class ComposeMessageService {
...
+        do {
+            let r = try await core.composeEmail(
+                msg: message,
+                fmt: MsgFmt.encryptInline
+            )
+
+            try await messageGateway.sendMail(input: MessageGatewayInput(mime: r.mimeEncoded, threadId: threadId))
+        } catch {
+            throw ComposeMessageError.gatewayError(error)
+        }
```
tomholub: This needs additional scrutiny in another issue. Error may be thrown by either composeEmail or sendMail. But either way we are throwing `gatewayError`. It should be balled `composeError` (and make sure it can handle both types of errors)

FlowCrypt/Functionality/Services/Key Services/KeyService.swift
```diff
@@ -110,17 +112,25 @@ final class KeyService: KeyServiceType {
...
+                let emails = key.pgpUserEmails
+
+                if emails.first == email {
+                   return key
+                }
+
+                if foundKey == nil, emails.contains(where: { $0 == email }) {
+                    foundKey = key
+                }
             }
```
tomholub: This is what I wanted, though I wish the code was more obvious, like: (pseudocode)

```
if primaryEmailMatch = keyDetails.find({ $0.pgpUserEmails.first == email }) {
  return primaryEmailMatch
}
if anyEmailMatch = keyDetails.find({ $0.pgpUserEmails.contains(email) }) {
  return anyEmailMatch
}
return nil
```

That way, even someone who has no clue about what the code was for will figure it out quickly. As written now, it's less looping but more brain to understand (at least for me) even when I know what it's supposed to do and why.

FlowCryptAppTests/Functionality/Services/Key Services/KeyServiceTests.swift
```diff
@@ -0,0 +1,122 @@
...
+            expiration: nil,
+            users: ["Bill <bill@test.com>", "Jim <jim@example.com>"],
+            algo: nil
+        )
+        let keyParser: KeyService.KeyParser = { data in
+            return CoreRes.ParseKeys(
+                format: .armored,
+                keyDetails: [keyDetails_1, keyDetails_2]
+            )
+        }
```
tomholub, line 60: There is no need to create a KeyParser type and mock this: this mock can be avoided by instead generating the actual keys with desired properties. That will be a much more direct test of the functionality, and it will simplify production code by removing `KeyService.KeyParser` which is there only for these tests.

FlowCrypt/Functionality/Services/Key Services/KeyService.swift
ivan-ushakov: Idea here was to avoid double loop over the same data and also have only one `pgpUserEmails` since this is computational property and calculated (parsed) each time. No problem, will use your approach.

FlowCryptAppTests/Functionality/Services/Key Services/KeyServiceTests.swift
ivan-ushakov: Idea was to have pure unit test not integration one. Since we have separate test for the parsing inside Core class there is no need to test it again. Also this makes test slower because parsing and executing TS code takes time. But I can use your approach if you want.

FlowCrypt/Functionality/Services/Key Services/KeyService.swift
tomholub: Does Swift have any type for a cached (or calculated once) calculated property? In fact, the values in KeyDetails never change, can they be made immutable?

FlowCryptAppTests/Functionality/Services/Key Services/KeyServiceTests.swift
tomholub: Thank you - I'm not too worried between a 5ms test and 30ms test. What worries me is this: 

https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fmedia.giphy.com%2Fmedia%2FsYz9N4FHOLlxC%2Fgiphy.gif&f=1&nofb=1

And also complex production code.

For our codebase, mocks are more appropriate for things that are really heavy like network calls, or situations where it's not possible to easily induce the desired test behavior without a mock.

But for a bit of parsing, it's better to let the production code do its job, even if it's already covered by other tests.

ivan-ushakov: I tried to use `Core.generateKey` but private key always have only one KeyDetails for there is no way to check algorithm completely like I did in my original test. Maybe it is not possible to have several KeyDetails for the one armored private key?

tomholub: One armored private key gives you a single KeyDetails, but nothing stops you from making several private keys. After all, in the app storage, there is one private key per database record, and they all get parsed separately, each producing one KeyDetails structure.

FlowCrypt/Functionality/Services/Key Services/KeyService.swift
```diff
@@ -80,4 +81,44 @@ final class KeyService: KeyServiceType {
...
+
+        return PrvKeyInfo(keyInfo: foundKey, passphrase: passphrase)
+    }
+
+    private func findKeyByUserEmail(keysInfo: [KeyInfo], email: String) throws -> KeyInfo? {
+        let keys: [(KeyInfo, KeyDetails?)] = try keysInfo.map {
+            let parsedKeys = try self.coreService.parseKeys(
+                armoredOrBinary: $0.`private`.data()
+            )
+            return ($0, parsedKeys.keyDetails.first)
```
ivan-ushakov, line 111: Parsed private key could have only one KeyDetails. Am I right here?

tomholub (approved):
This looks good. I understand it's just a different coding philosophy, but the sooner we unite this, the better. Thanks!

PR #676, opened 2021-10-14
Feature/issue 553 multiple recipient keys

### LINKED ISSUE #337: PubLookup.lookup to return an array of public keys
after #553 

we are blindly choosing `.first` public key which is not ideal, and instead in the future we should return `[Contact]`. If we are pulling keys during compose to save them in db, save all of them

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/296#issuecomment-855239434_

### LINKED ISSUE #553: contacts store should allow more than one public key per recipient email address
Currently, I believe we restricted maximum of 1 public key per recipient in the db, and also in Settings -> Contacts.

We should instead allow more than one public key per recipient. With this, Settings -> Contacts should get updated - next to each email it should say something like `(1 public keys)` or `(3 public keys)` with 50% opacity.

Once I click on it to view the public key, instead of the single public key view that is there now, I should instead view the user's email address and then a list of public keys with details about them, some sort of large clickable "cards". 

Once I click on one of the public key cards, then I'll see the current view public key activity with more details about the public key, and on top right the icons to share/copy/delete the public key.

Each card should have:

```
Fingerprint: (primary fingerprint in small letters so it fits the line)
Primary user: some@example.com
Created: some date
Expires: some date or "never"
```

sosnovsky:
This PR changes scheme to store multiple public keys per contacts.
UI also updated to show all available contact public keys + new screen added for contact key details.

close #553
close #337 

- Tests added or updated - updated tests to use new scheme with multiple contact keys

### To be filled by reviewers


FlowCrypt/Controllers/Settings/Contacts/Contacts List/ContactKeyDetailViewController.swift
```diff
@@ -0,0 +1,139 @@
...
+    private let key: ContactKey
+    private let action: ContactKeyDetailAction?
+
+    init(
+        decorator: ContactKeyDetailDecoratorType = ContactKeyDetailDecorator(),
+        key: ContactKey,
+        action: ContactKeyDetailAction?
+    ) {
+        self.decorator = decorator
+        self.key = key
```
tomholub, line 38: let's call it `publicKey` on both left and right side. So that it's obvious in the code when comparing to private keys.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -111,18 +111,18 @@ final class ComposeMessageService {
...
-    private func getPubKeys(for recepients: [ComposeMessageRecipient]) -> Result<[String], MessageValidationError> {
-        let pubKeys = recepients.map {
-            ($0.email, contactsService.retrievePubKey(for: $0.email))
+    private func getPubKeys(for recipients: [ComposeMessageRecipient]) -> Result<[String], MessageValidationError> {
+        let pubKeys = recipients.map {
+            ($0.email, contactsService.retrievePubKeys(for: $0.email))
         }
 
-        let emailsWithoutPubKeys = pubKeys.filter { $0.1 == nil }.map(\.0)
+        let emailsWithoutPubKeys = pubKeys.filter { $0.1.isEmpty }.map(\.0)
```
tomholub, line 119: This multiple-depth reference is getting a bit confusing, could it use a named argument? At least the first one?

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -111,18 +111,18 @@ final class ComposeMessageService {
...
 
         guard emailsWithoutPubKeys.isEmpty else {
             return .failure(.noPubRecipients(emailsWithoutPubKeys))
         }
 
-        return .success(pubKeys.compactMap(\.1))
+        return .success(pubKeys.filter({ $0.1.isNotEmpty }).flatMap(\.1))
```
tomholub, line 125: Similar here

FlowCrypt/Functionality/Services/Contacts Services/LocalContactsProvider.swift
```diff
@@ -61,21 +58,37 @@ extension LocalContactsProvider: LocalContactsProviderType {
     }
 
     func searchContact(with email: String) -> Contact? {
+        guard let contactObject = find(with: email) else { return nil }
+        return Contact(contactObject)
+    }
```
tomholub, line 63: This method should be probably updated to return `[Contact]`, renamed to `searchContacts`

FlowCrypt/Functionality/Services/Contacts Services/LocalContactsProvider.swift
```diff
@@ -61,21 +58,37 @@ extension LocalContactsProvider: LocalContactsProviderType {
...
     }
 
-    func getAllContacts() -> [Contact] {
-        Array(
-            localContactsCache.realm
-                .objects(ContactObject.self)
-                .map {
-                    let keyDetail = try? core.parseKeys(armoredOrBinary: $0.pubKey.data()).keyDetails.first
-                    return Contact($0, keyDetail: keyDetail)
+    func remove(pubKey: String, for email: String) {
```
tomholub, line 77: This needs a different signature - removing by email+fingerprint of the key, instead of the key string.

tomholub: for me - Also need to double check the structure of the table later, and indexes

FlowCrypt/Functionality/Services/Contacts Services/LocalContactsProvider.swift
tomholub: `Contact` is not a good name here, actually. Let's call it, for now, `RecipientPubKey` for lack of a better name.

And yup - there may be more than one RecipientPubKey per recipient, which feels more natural to say now that we renamed it.

FlowCrypt/Functionality/Services/Contacts Services/Models/Contact.swift
```diff
@@ -12,65 +12,33 @@ struct Contact {
...
-    /// first pubkey fingerprint
-    var fingerprint: String? { fingerprints.first }
-
-    /// pubkey created date
-    let pubkeyCreated: Date?
-
-    let algo: KeyAlgo?
+    /// public keys
+    var pubKeys: [ContactKey]
 }
```
tomholub, line 19: Very good, let's call `Contact` a `RecipientWithPubKeys`

tomholub: Then `ContactKey` should be I suppose `PubKey` ? 

I think `PubKey` and `KeyDetails` will be very similar. Maybe they could be merged or further renamed, but maybe that would be another PR. As this PR ripens, it will become more obvious what to call it.

FlowCrypt/Functionality/Services/Contacts Services/Models/ContactKey.swift
```diff
@@ -0,0 +1,57 @@
...
+
+struct ContactKey {
+    let key: String
+    /// will be provided later
+    let lastSig: Date?
+    /// the date when key was retrieved from Attester, or nil
+    let lastChecked: Date?
+    /// expiration date
+    let expiresOn: Date?
+    /// all key longids, comma-separated
```
tomholub, line 19: is it still up to date that we store fingerprints comma-separated in the db? Realm doesn't have some native string array support?

FlowCrypt/Functionality/Services/Contacts Services/Models/ContactKey.swift
```diff
@@ -0,0 +1,57 @@
...
+//  Copyright © 2017-present FlowCrypt a. s. All rights reserved.
+//
+    
+import Foundation
+
+struct ContactKey {
+    let key: String
+    /// will be provided later
+    let lastSig: Date?
+    /// the date when key was retrieved from Attester, or nil
```
tomholub, line 15: `... from a public key server, or nil`

FlowCrypt/Functionality/WKDURLs/PubLookup.swift
```diff
@@ -27,14 +27,7 @@ class PubLookup: PubLookupType {
...
-            // then eg encrypt for all returned Contacts
-            // also stop throwing below - no point. Return
-            //  empty array then handle downstream
-            guard let keyDetail = keyDetails.first else {
-                throw ContactsError.keyMissing
-            }
-            resolve(Contact(email: email, keyDetail: keyDetail))
+            resolve(Contact(email: email, keyDetails: keyDetails))
         }
     }
```
tomholub, line 32: Good. With this, `getKeyDetails` doesn't need to be a separate method, anymore, its contents can be moved into `lookup`

FlowCrypt/Functionality/Services/Contacts Services/Models/ContactKey.swift
sosnovsky: No, in Realm we store fingerprints as `List<String>`, comment here is outdated. I'll remove `comma-separated` part

FlowCrypt/Controllers/Settings/Contacts/Contacts List/ContactKeyDetailViewController.swift
tomholub: sure, pubKey is good

FlowCrypt/Functionality/WKDURLs/PubLookup.swift
```diff
@@ -24,35 +24,20 @@ class PubLookup: PubLookupType {
...
-                resolve([])
+                resolve(RecipientWithPubKeys(email: email, keyDetails: []))
                 return
             }
 
             let wkdResult = try awaitPromise(self.wkd.lookupEmail(email))
             if !wkdResult.isEmpty {
-                resolve(wkdResult)
+                resolve(RecipientWithPubKeys(email: email, keyDetails: wkdResult))
             }
```
tomholub, line 37: I suppose it's missing a return here after the resolve. I'm not sure if it makes any difference, but it makes me feel better :-)

FlowCrypt/Models/Realm Models/ContactKeyObject.swift
```diff
@@ -0,0 +1,58 @@
...
+//  FlowCrypt
+//
+//  Created by Roma Sosnovsky on 11/10/21
+//  Copyright © 2017-present FlowCrypt a. s. All rights reserved.
+//
+
+import Foundation
+import RealmSwift
+
+final class PubKeyObject: Object {
```
tomholub, line 12: Please rename the file to match the class also

FlowCrypt/Models/Realm Models/ContactKeyObject.swift
```diff
@@ -0,0 +1,58 @@
...
+    @Persisted(primaryKey: true) var key: String = ""
```
tomholub, line 13: The primary fingerprint (which every OpenPGP key has) should be the primary key on this table.

For column `key` let's call it `armored` instead. It means the key in armored format, with the `----BEGIN PGP PUBLIC KEY-----` header and so on, which is what we're storing.

tomholub: That means the `PublicKey` type could be updated too to reflect.

FlowCrypt/Models/Realm Models/ContactObject.swift
```diff
@@ -10,84 +10,52 @@ import Foundation
...
     convenience init(value: String) {
         self.init()
         self.value = value
     }
 }
 
-final class ContactObject: Object {
-    @objc dynamic var email: String = ""
-    @objc dynamic var pubKey: String = ""
+final class RecipientObject: Object {
```
tomholub, line 21: Please update file name

tomholub (commented):
Very good. Please see comments. Haven't tried to run it yet.

Please also see if you are able to run the Appium tests after a change like this. I'll send you the env file. Thanks!

tomholub (approved):

PR #680, opened 2021-10-16
Fixed #674, Search screen shows white screen even in dark mode

### LINKED ISSUE #674: white background appears in dark mode theme
I found this very minor ui issue, It occurs when a user has dark mode theme enabled in the ios settings and perform a quick search that doesn't occupied the entire screen.

image

expected behavior:
unoccupied spaces should be covered by the grayish color to fill the gap.

paragdulam:
This PR fixes the UI issue in the Search screen where it shows white screen background in dark mode when search results don't cover up the entire screen.

close #674 // if this PR closes an issue

issue #674 // if it doesn't close the issue yet

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Search/SearchViewController.swift
+++ b/FlowCrypt/Controllers/Search/SearchViewController.swift
@@ -9,6 +9,7 @@
 import AsyncDisplayKit
 import FlowCryptCommon
 import FlowCryptUI
+import UIKit
 
 /**
  * View controller to search messages in folders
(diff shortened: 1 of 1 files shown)


tomholub:
> Tests (delete all except exactly one):
>    Does not need tests (refactor only, docs or internal changes)

This is not an internal change - as long as the user can observe a difference, then it's not by definition. Since it's a change of color that doesn't break functionality, this would fall under "not worth testing"

PR #690, opened 2021-10-17
save drafts during compose or reply

### LINKED ISSUE #514: save drafts during compose or reply
When I type my email or response, drafts should get continuously uploaded to Gmail API (can skip IMAP for now).

 - once a second, check if the content of the message, subject or recipients have changed compared to previous second
 - if it changed, and initiate an upload/update of draft in the background
 - (If it's the first upload of this draft, receive back the draft ID, and then use that draft id in subsequent updates of the draft)
 - if it didn't change, skip updating draft and wait until next run
 - cancel this timer when user leaves the screen through any means (back button or menu -> other place), but make sure to run this method one last time in the background while the screen is being switched, to ensure the last changes are saved
 - do not include attachments in these drafts (Gmail API is not good, would need to re-upload them each time)

ekievsky:
This PR is adding feature that saves draft encrypted mail during compose new mail or reply

close #514 

- Difficult to test (explain why). there was not tests to this modules ( gmail service etc ). if we need to cover it with tests I would prefer it as separate issue. 

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Compose/ComposeViewController.swift
+++ b/FlowCrypt/Controllers/Compose/ComposeViewController.swift
@@ -153,13 +153,16 @@ final class ComposeViewController: TableNodeViewController {
 // MARK: - Drafts
 extension ComposeViewController {
     @objc private func startTimer() {
-        saveDraftTimer = Timer.scheduledTimer(
-            timeInterval: 1,
-            target: self,
-            selector: #selector(saveDraftIfNeeded),
-            userInfo: nil,
-            repeats: true)
-        saveDraftTimer?.fire()
+        DispatchQueue.global().async { [weak self] in
+            guard let self = self else { return }
+            self.saveDraftTimer = Timer.scheduledTimer(
+                timeInterval: 1,
+                target: self,
+                selector: #selector(self.saveDraftIfNeeded),
+                userInfo: nil,
+                repeats: true)
+            self.saveDraftTimer?.fire()
+        }
     }
 
     @objc private func stopTimer() {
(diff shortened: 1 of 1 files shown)


FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -49,6 +55,9 @@ final class ComposeViewController: TableNodeViewController {
 
     private var state: State = .main
 
+    private weak var timer: Timer?
```
tomholub, line 58: Let's call it `saveDraftTimer`

FlowCrypt/Functionality/Mail Provider/Message Gateway/GmailService+send.swift
```diff
@@ -35,4 +35,58 @@ extension GmailService: MessageGateway {
...
+                }
+                if let error = error {
+                    return promise(.failure(GmailServiceError.providerError(error)))
+                }
+                promise(.success(draft))
+            }
+        }
+    }
+
+    private func resolveQueryForDraftAction(raw: String, threadId: String?, draft: GTLRGmail_Draft?) -> GTLRGmailQuery {
```
tomholub, line 62: This should be called `createQueryForDraftAction` ?

tomholub:
Need to make sure that:
 - the draft is deleted after the message is sent, so unneeded drafts don't accumulate
 - I can go to drafts folder, open a draft and be able to continue editing it

ekievsky:
@tomholub I thought that it will be separate task.

I will continue working on this feature then, thank you

PR #698, opened 2021-10-18
issue 520 appium test for not needing pass phrase for plain message

fcvakintos:
This PR for appium coverage of https://github.com/FlowCrypt/flowcrypt-ios/issues/520


appium/tests/specs/inbox/ReadEmailAfterRestartApp.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+        const senderEmail = CommonData.sender.email;
+        const emailSubject = 'Test 1';
+        const emailText = 'Test email';
+
+        const bundleId = CommonData.bundleId.id;
+
+        SplashScreen.login();
+        CreateKeyScreen.setPassPhrase();
+
+        InboxScreen.clickOnUserEmail(senderEmail);
```
tomholub, line 23: We'll have to be choosing emails in inbox by subject, there will be many more over time. Also, the subject will be more descriptive, eg in this case `plain email` instead of `Test email`.

fcvakintos: I am agree with you @tomholub , we need to send email via api before the test so we can pass dynamic subject and mail body to the test in future. I propose to create separate task for that.

tomholub: For now, you could do what you are doing - have the emails pre-arranged in the inbox instead of dynamically sending. But instead of going by sender, go by subject. Could you do that?

tomholub (approved):

PR #844, opened 2021-10-29
issue 743 contacts.other search

### LINKED ISSUE #743: contact search finds very few contacts compared to gmail
part of https://github.com/FlowCrypt/flowcrypt-ios/issues/606

On browser extension, I think the api that Google is deprecating is a lot more powerful than the People API. Could you resarch what is the proper way to search recipient using Google APIs?

Eg when I use Gmail, I can enter "joe" and it will find 20 different joe people I've communicated with before. But If I put "joe" in iOS, it will only find him if I manually created a Google contact for that person, which is not very useful.

(on the browser extension, we also loop through sent emails and create our own contact list from there. But would prefer to first see what to conventional API options are)

sosnovsky:
This PR adds search through email addresses which are not stored in user's Gmail contacts. It also needs additional auth scope `https://www.googleapis.com/auth/contacts.other.readonly` (currently it shows warning screen when user authenticates for the first time)

close #743

- Difficult to test (explain why) - it needs Google auth for testing search results.

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/CheckAuthScopes/CheckAuthScopesViewController.swift
+++ b/FlowCrypt/Controllers/CheckAuthScopes/CheckAuthScopesViewController.swift
@@ -16,6 +16,7 @@ private extension GoogleScope {
         case .userInfo: return "User Info"
         case .mail: return "Gmail"
         case .contacts: return "Contacts"
+        case .otherContacts: return "Other Contacts"
         }
     }
 }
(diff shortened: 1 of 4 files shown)


tomholub (commented):
works excellently. I've submitted the scope to google, now waiting to get approved.

tomholub (approved):

tomholub:
approved by google

PR #862, opened 2021-10-30
Clarify why did keys fail to load from storage

### LINKED ISSUE #858: clarify why did keys fail to load from storage
a line like this is ignoring the details of the error that happened:

```
            guard let keys = try? self.keyService.getPrvKeyInfo().get(), keys.isNotEmpty else {
                return reject(CoreError.notReady("Failed to load keys from storage"))
            }
```

it's bad practice. Errors should be propagated, not overwritten, to preserve details. If we want to re-throw another error, that error should wrap the original error so that original details can still be accessed.

In this particular case, there is just one error, but who knows - there may be more.

```
    func getPrvKeyInfo() -> Result<[PrvKeyInfo], KeyServiceError> {
        guard let email = currentUserEmail() else {
            return .failure(.retrieve)
        }
```

Additionally, this above error is clearly misnamed. It should be `return .failure(.missingCurrentUserEmail)` to be more precise. (to update in all such situations)

### LINKED ISSUE #859: CoreError should only be thrown for situations related to JavaScriptCore
This is complete nonsense

```swift
CoreError.notReady("Failed to load keys from storage")
```

`CoreError.notReady` is meant for situation where Core is not initialized/ready, it has nothing to do with keys storage.

All usages of `CoreError` should be reviewed to make sure they are not used in unrelated context.

ivan-ushakov:
close #858
close #859

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Resources/en.lproj/Localizable.strings
+++ b/FlowCrypt/Resources/en.lproj/Localizable.strings
@@ -48,6 +48,7 @@
 "keyServiceError_retrieve_error" = "Could not retrieve keys from DataService. Please restart the app and try again.";
 "keyServiceError_retrieve_parse" = "Could not parse keys from storage. Please reinstall the app.";
 "keyServiceError_retrieve_unexpected" = "Could not import key. Please try to relogin.";
+"keyServiceError_missing_current_email" = "Missing current email. Please restart the app and try again.";
 
 "backupServiceError_parse" = "Key can't be parsed";
 "backupServiceError_email" = "Missing user email";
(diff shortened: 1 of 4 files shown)


FlowCrypt/Resources/en.lproj/Localizable.strings
```diff
@@ -48,6 +48,7 @@
 "keyServiceError_retrieve_error" = "Could not retrieve keys from DataService. Please restart the app and try again.";
 "keyServiceError_retrieve_parse" = "Could not parse keys from storage. Please reinstall the app.";
 "keyServiceError_retrieve_unexpected" = "Could not import key. Please try to relogin.";
+"keyServiceError_missing_current_email" = "Missing current email. Please restart the app and try again.";
```
ivan-ushakov, line 51: Not sure that user will understand this. It is not possible to get this situation.

tomholub: True, thanks for raising it. Looks ok enough for now, we'll improve it later (I'll file an issue)

tomholub: Related issue for the future https://github.com/FlowCrypt/flowcrypt-ios/issues/863

tomholub (approved):

PR #854, opened 2021-10-30
issue 852 no prv found - added logs + refactor + pass phrase for signing

tomholub:
This PR synchronizes calls to Core using a queue and async methods, while refactoring the app for async (usage of Core) and fixing many other issues along the way:
- close #852
- close #819
- close #851
- close #860
- close #849 - possibly
- close #884
- close #895
- close #878
- close #681
- issue #867
- close #895

----------------------------------

**Tests**
 - new tests added, existing updated

--------------------------------

### To be filled by reviewers

I have reviewed that this PR... _(tick whichever items you personally focused on during this review)_:
- [x] addresses the issue it closes (if any)
- [x] code is readable and understandable
- [x] is accompanied with tests, or tests are not needed
- [x] is free of vulnerabilities
- [x] is documented clearly and usefully, or doesn't need documentation

Issue #812, opened 2021-10-27
PubKey type should have emails and uids type on it


tomholub:
```swift
struct PubKey {
    let armored: String
    /// will be provided later
    let lastSig: Date?
    /// the date when key was retrieved from a public key server, or nil
    let lastChecked: Date?
    /// expiration date
    let expiresOn: Date?
    /// all key longids
    let longids: [String]
    /// all key fingerprints
    let fingerprints: [String]
    /// key created date
    let created: Date?
    /// key algo
    let algo: KeyAlgo?
    /// is key revoked
    let isRevoked: Bool
}
```

When parsing a key, the pgp user ids are available, and so are pgp user id emails once parsed. These should be included in this type.

Then all usages of `armored` should be investigated, to see if it's being used solely for parsing out the emails. Such usages should be replaced directly with the newly added property.

sosnovsky:
@tomholub I added user emails to `PubKey` in https://github.com/FlowCrypt/flowcrypt-ios/commit/0b99faa62e7258513cb8c0b5deaef5b84db05546, but haven't found any usage of `armored` for parsing emails. Usually we already have recipient email and `armored` is used for parsing `KeyDetails` data, for example:
1️⃣ https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/Functionality/Services/Compose%20Message%20Service/ComposeMessageService.swift#L121
2️⃣ https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/Functionality/Services/Local%20Pub%20Key%20Services/LocalContactsProvider.swift#L117

tomholub:
Thanks. After your changes merge, I'll have another look so see if it can be streamlined.

Issue #230, opened 2020-04-28
streaming encrypt/decrypt


boby1975:
Hi
Does anyone know if is it possible in this project do "streaming encrypt and decrypt" of large files according openPGP?
Because I still can’t find a proper working solution for this task for my iOS app (swift).

Thanks in advance.

tomholub:
It's technically possible. As for our app, we have not implemented streaming yet, but eventually we will.

boby1975:
Thanks for the info. As far as I understand you are using OpenPGP.js v4.6.0. It seems that there has already appeared streaming encryption and compression.

tomholub:
Yes, but we're not taking advantage of these things yet. I'll mark this for later, there's quite a few things we have to do before we can implement streaming.

tomholub:
part of #275

PR #893, opened 2021-11-02
Feature/issue 783 unlimited attachment size

### LINKED ISSUE #783: allow opening messages of unlimited size + add download progress
Currently, messages with attachments over 5mb are disabled.

I'd like to re-enable them but during loading, show progress spinner. And allow user to use the back icon in case they don't want to wait.

_Sometime in the future, we'll separate the logic to only pull message body and not attachments first, but that's not trivial for OpenPGP messages (content is often in attachments) so that will later need a way to differentiate which attachments to pull. I'll file another issue for that._

sosnovsky:
This PR allows loading of attachments over 5mb and adds progress spinner for download process.

close #783 

- Difficult to test (explain why) - it's mostly UI change

### To be filled by reviewers


FlowCrypt/Functionality/Mail Provider/Message Provider/Gmail+Message.swift
```diff
@@ -8,34 +8,83 @@
...
-                guard let data = GTLRDecodeWebSafeBase64(raw) else {
-                    return reject(GmailServiceError.missedMessageInfo("data"))
+                guard let sizeEstimate = gmailMessage.sizeEstimate?.floatValue else {
+                    continuation.resume(throwing: GmailServiceError.missedMessageInfo("sizeEstimate"))
+                    return
                 }
-                resolve(data)
+
+                // google returns smaller estimated size
+                let totalSize = sizeEstimate * Float(1.3)
```
tomholub, line 73: I did the exact same fix on browser extension. Did you get this from there? :-)

tomholub (commented):
code looks good so far

FlowCrypt/Functionality/Mail Provider/Message Provider/Gmail+Message.swift
sosnovsky: Haha, no, I haven't seen your fix, just calculated difference between real and estimated sizes 🙂 Usually it's around 30%

sosnovsky:
It'll be nice to have decrypt progress percentage in the future too, as it takes almost the same time as download for big files.

sosnovsky:
@tomholub currently I have only https://github.com/FlowCrypt/flowcrypt-ios/issues/812 task assigned to me, can you please check on what tasks should I work after this one?

FlowCrypt/Functionality/Mail Provider/Message Provider/Gmail+Message.swift
tomholub: Careful to cap it at 100% when calculating. Else sometimes user will see 103%..

I think the difference is in base64 encoding which tends to add about a third.

sosnovsky: Yes, I handle this case with setting `1` as max progress value:
```
let progress = min(Float(received)/messageSize, 1)
```

tomholub:
> It'll be nice to have decrypt progress percentage in the future too, as it takes almost the same time as download for big files.

Download is network bound and decryption is cpu bound. You could do both at the same time, if you do streaming decryption. #230 But it's not simple to do across the Swift/JS boundary. But it's possible.

tomholub:
> @tomholub currently I have only #812 task assigned to me, can you please check on what tasks should I work after this one?

will do

sosnovsky:
> > It'll be nice to have decrypt progress percentage in the future too, as it takes almost the same time as download for big files.
> 
> Download is network bound and decryption is cpu bound. You could do both at the same time, and it should take half the time overall, if you do streaming decryption. #230 But it's not simple to do across the Swift/JS boundary. But it's possible.

Got it, it'll be a great addition for app's functionality and will make large emails loading much faster.

tomholub:
I'd like you to debug/finalize #854 . How about merging sequence? I suppose it may be easier to merge the big one first and then rebase this onto that, but we could do it the other way around if you think it's easier. Up to you.

sosnovsky:
I agree - it should be easier to merge #854 to `master` first, and then rebase it into this PR, as there is less changes here. I'll check #854 now

tomholub:
Tested, looks great

tomholub:
@sosnovsky good to merge once conflicts addressed 👍

sosnovsky:
@tomholub I merged `master` changes to this branch - the only issue was with `fetchMsg` function, it didn't execute async code for fetching mail size. I fixed it by using `async` instead of `Promise`.
This fix also affected `Imap` class (to make it conform to updated `MessageProvider` protocol) - I updated `fetchMsg` function there too  https://github.com/FlowCrypt/flowcrypt-ios/pull/893/commits/763ac4b525d7cdecf34a7d971513fb8c87a26445#diff-8012f2f1a5073946e65bdfc88fd17e2b9ddad872c955b7cf6adfd7a1621bf345R12

FlowCrypt/Controllers/Msg/MessageViewController.swift
```diff
@@ -148,26 +148,38 @@ final class MessageViewController: TableNodeViewController {
...
-        }
-        .then(on: .main) { [weak self] in
-            self?.handleReceivedMessage()
+        Task {
+            do {
+                processedMessage = try await serviceActor.fetchDecryptAndRenderMsg(message: input.objMessage,
+                                                                                   path: input.path,
+                                                                                   progressHandler: { [weak self] state in
+                                                                                        self?.handleFetchProgress(state: state)
+                                                                                   })
```
tomholub, line 161: A detail - I'd prefer formatting like:
```swift
processedMessage = try await serviceActor.fetchDecryptAndRenderMsg(
    message: input.objMessage,
    path: input.path,
    progressHandler: { [weak self] state in
        self?.handleFetchProgress(state: state)
    }
)
```
Else all of the inputs have no space due to large indentation.

FlowCrypt/Controllers/Msg/MessageViewController.swift
```diff
@@ -556,16 +554,20 @@ private actor ServiceActor {
...
 
     func checkAndPotentiallySaveEnteredPassPhrase(_ passPhrase: String) async throws -> Bool {
-        return try await messageService.checkAndPotentiallySaveEnteredPassPhrase(passPhrase)
+        try await messageService.checkAndPotentiallySaveEnteredPassPhrase(passPhrase)
     }
 
     func decryptAndProcessMessage(mime: Data) async throws -> ProcessedMessage {
-        return try await messageService.decryptAndProcessMessage(mime: mime)
+        try await messageService.decryptAndProcessMessage(mime: mime)
     }
```
tomholub, line 572: I like explicit returns for the clarity :-) but don't insist.

tomholub (approved):
👍

FlowCrypt/Controllers/Msg/MessageViewController.swift
sosnovsky: Okay, I'll use such formatting in the next commits

Issue #880, opened 2021-11-01
tests that need many cores to reproduce issues - how to run


tomholub:
Is there a way to run this code on simulator that actually uses many OS threads for the simulator?

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/854#discussion_r739864405_

The test reproduces issues on real device but not simulator.

tomholub:
<img width="649" alt="image" src="https://user-images.githubusercontent.com/6306961/139683955-c8e57d86-be09-4cfe-a9cd-a1a83858a397.png">

ivan-ushakov:
@tomholub 
You did it right and test helped me to catch concurrency bug

tomholub:
> @tomholub You did it right and test helped me to catch concurrency bug

But did you run it on real iPhone or on your mac/simulator?

ivan-ushakov:
> But did you run the test on real iPhone or on your mac/simulator?

On simulator

tomholub:
My mac has only two cores - `1.2 GHz Dual-Core Intel Core m3`, and doesn't reproduce such problem when I run it. I've added updated comment in PR review.

PR #904, opened 2021-11-03
Core class should be converted to actor

### LINKED ISSUE #869: Core class should be converted to actor
Since you want to have all `Core` class methods to be synchronised, you could use `actor` keyword. So you can remove queue and `withCheckedThrowingContinuation`

_Originally posted by @ivan-ushakov in https://github.com/FlowCrypt/flowcrypt-ios/pull/854#discussion_r739842544_

ivan-ushakov:
close #869
issue #880

### To be filled by reviewers

### DIFF
--- a/Core/source/assets/flowcrypt-ios-begin.js
+++ b/Core/source/assets/flowcrypt-ios-begin.js
@@ -35,8 +35,8 @@ const window = { navigator, crypto, setTimeout, clearTimeout };
 const self = { window, navigator, crypto };
 global.window = window;
 
-var engine_host_cb_value_formatter = function(val) {
-    coreHost.handleCallback(val.json, val.data);
+var engine_host_cb_value_formatter = function(key, val) {
+    coreHost.handleCallback(key, val.json, val.data);
 };
 
 (function() {
--- a/Core/source/assets/flowcrypt-ios-begin.js
+++ b/Core/source/assets/flowcrypt-ios-begin.js
@@ -35,8 +35,8 @@ const window = { navigator, crypto, setTimeout, clearTimeout };
 const self = { window, navigator, crypto };
 global.window = window;
 
-var engine_host_cb_value_formatter = function(key, val) {
-    coreHost.handleCallback(key, val.json, val.data);
+var engine_host_cb_value_formatter = function(callbackId, val) {
+    coreHost.handleCallback(callbackId, val.json, val.data);
 };
 
 (function() {
--- a/Core/source/entrypoint-bare.ts
+++ b/Core/source/entrypoint-bare.ts
@@ -11,17 +11,17 @@ declare const global: any;
 
 const endpoints = new Endpoints();
 
-global.handleRequestFromHost = (endpointName: string, endpointKey: string, request: string, data: Uint8Array, cb: (key: string, response: EndpointRes) => void): void => {
+global.handleRequestFromHost = (endpointName: string, callbackId: string, request: string, data: Uint8Array, cb: (key: string, response: EndpointRes) => void): void => {
   try {
     const handler = endpoints[endpointName];
     if (!handler) {
-      cb(endpointKey, fmtErr(new Error(`Unknown endpoint: ${endpointName}`)));
+      cb(callbackId, fmtErr(new Error(`Unknown endpoint: ${endpointName}`)));
     } else {
       handler(JSON.parse(request), [data])
-        .then(res => cb(endpointKey, res))
-        .catch(err => cb(endpointKey, fmtErr(err)));
+        .then(res => cb(callbackId, res))
+        .catch(err => cb(callbackId, fmtErr(err)));
     }
   } catch (err) {
-    cb(endpointKey, fmtErr(err));
+    cb(callbackId, fmtErr(err));
   }
 };
--- a/Core/source/entrypoint-bare.ts
+++ b/Core/source/entrypoint-bare.ts
@@ -11,17 +11,17 @@ declare const global: any;
 
 const endpoints = new Endpoints();
 
-global.handleRequestFromHost = (endpointName: string, request: string, data: Uint8Array, cb: (response: EndpointRes) => void): void => {
+global.handleRequestFromHost = (endpointName: string, endpointKey: string, request: string, data: Uint8Array, cb: (key: string, response: EndpointRes) => void): void => {
   try {
     const handler = endpoints[endpointName];
     if (!handler) {
-      cb(fmtErr(new Error(`Unknown endpoint: ${endpointName}`)));
+      cb(endpointKey, fmtErr(new Error(`Unknown endpoint: ${endpointName}`)));
     } else {
       handler(JSON.parse(request), [data])
-        .then(res => cb(res))
-        .catch(err => cb(fmtErr(err)));
+        .then(res => cb(endpointKey, res))
+        .catch(err => cb(endpointKey, fmtErr(err)));
     }
   } catch (err) {
-    cb(fmtErr(err));
+    cb(endpointKey, fmtErr(err));
   }
 };
--- a/FlowCrypt/Core/Core.swift
+++ b/FlowCrypt/Core/Core.swift
@@ -63,8 +63,7 @@ actor Core: KeyDecrypter, KeyParser, CoreComposeMessageType {
     private var vm = JSVirtualMachine()!
     private var context: JSContext?
 
-    private var cb_last_value = [CallbackResult]()
-    private var started = false
+    private var callbackResults = [CallbackResult]()
     private var ready = false
 
     private lazy var logger = Logger.nested(in: Self.self, with: "Js")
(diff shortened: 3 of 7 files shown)


FlowCryptAppTests/Core/FlowCryptCoreTests.swift
```diff
@@ -387,9 +388,6 @@ final class FlowCryptCoreTests: XCTestCase {
         }
     }
 
-    // this test is only meaningful on a real device
```
ivan-ushakov, line 390: Now it makes sense and also helped catch bug

FlowCrypt/Core/Core.swift
```diff
@@ -53,20 +53,22 @@ protocol KeyParser {
...
     private var cb_catcher: JSValue?
-    private var cb_last_value: (String, [UInt8])?
     private var vm = JSVirtualMachine()!
     private var context: JSContext?
-    private dynamic var started = false
-    private dynamic var ready = false
-    
-    private let queue = DispatchQueue(label: "com.flowcrypt.core", qos: .default) // todo - try also with .userInitiated
+
+    private var cb_last_value = [CallbackResult]()
```
ivan-ushakov, line 66: Let's name this variable `callbackValues` because we keep several values now

FlowCrypt/Core/Core.swift
```diff
@@ -53,20 +53,22 @@ protocol KeyParser {
...
+    private var started = false
+    private var ready = false
```
ivan-ushakov, line 68: Maybe we don't need both `ready` and `started` now

FlowCrypt/Core/Core.swift
```diff
@@ -236,47 +230,29 @@ final class Core: KeyDecrypter, KeyParser, CoreComposeMessageType {
...
-        // This will block the task for up to 1000ms if the app was just started and Core method was called before JSContext is ready
-        // It should only affect the user if Core method was called within 500-800ms of starting the app
-        let start = DispatchTime.now()
-        while !ready {
-            if start.millisecondsSince > 1000 { // already waited for 1000 ms, give up
-                throw CoreError.notReady("App Core not ready yet")
-            }
+        jsEndpointListener!.call(withArguments: [endpoint, String(data: jsonData, encoding: .utf8)!, Array<UInt8>(data), cb_catcher!])
+
+        while callbackResults.isEmpty {
```
ivan-ushakov, line 235: Since we have `CoreHost` to allow JavaScript world communicate with us, we have to call `handleCallbackResult()` method inside task. This leads to empty callback result because that task will be executed after we leave this method. To allow that task to be executed we need to "sleep" here for some small time interval.

FlowCrypt/Core/Core.swift
```diff
@@ -223,7 +217,7 @@ final class Core: KeyDecrypter, KeyParser, CoreComposeMessageType {
     }
 
     func handleCallbackResult(json: String, data: [UInt8]) {
-        cb_last_value = (json, data)
+        callbackResults.append((json, data))
     }
```
tomholub, line 221: How is it different? If this class is an actor, then single value and array is equivalent, since a single call can never produce more than one callback.

FlowCrypt/Core/Core.swift
```diff
@@ -236,47 +230,29 @@ final class Core: KeyDecrypter, KeyParser, CoreComposeMessageType {
...
-        let start = DispatchTime.now()
-        while !ready {
-            if start.millisecondsSince > 1000 { // already waited for 1000 ms, give up
-                throw CoreError.notReady("App Core not ready yet")
-            }
+        jsEndpointListener!.call(withArguments: [endpoint, String(data: jsonData, encoding: .utf8)!, Array<UInt8>(data), cb_catcher!])
+
+        while callbackResults.isEmpty {
             await Task.sleep(50 * 1_000_000) // 50ms
         }
```
tomholub, line 237: This is weird. By the time `jsEndpointListener.call` is done, the callback should be always available (JavaScriptContext should finish once the work is done). Did you have situations when it was still empty afterwards? Can you confirm again?

Otherwise, if necessary, should wait in 1ms intervals. Most calls are quick, 50ms is too long step to wait.

FlowCrypt/Core/Core.swift
```diff
@@ -236,47 +230,29 @@ final class Core: KeyDecrypter, KeyParser, CoreComposeMessageType {
     }
 
     private func call(_ endpoint: String, jsonData: Data, data: Data) async throws -> RawRes {
```
tomholub, line 232: Let's add `guard ready else { throw "Programming error: Core is not ready yet. Most likeyly startInBackgroundIfNotAlreadyRunning wasn't called first" }`

FlowCryptAppTests/Core/FlowCryptCoreTests.swift
tomholub: Let's add `This test always passes, even wrongly, on simulators running on a mac with 2 or fewer cores. Behaves meaningfully on real iPhone or simulator on a mac with many cores`

tomholub (commented):
Thanks!

FlowCrypt/Core/Core.swift
tomholub: It doesn't feel right - can you reproduce this situation with empty callback also before your refactor? I've never seen that happen.

ivan-ushakov: It is possible if you run several `Core` methods in parallel. For example this was the reason why `testCoreResponseCorrectnessUnderConcurrency()` test failed on my machine

ivan-ushakov: So without this delay we can not get any value from call-back, because while we are inside this `Core` method, any call to other `Core` methods will wait. To give `Core.handleCallbackResult()` method a chance for invocation we `await` on timer.

FlowCrypt/Core/Core.swift
ivan-ushakov: You can run several `Core` methods in parallel and one of them will override value of another since you have only single value

FlowCrypt/Core/Core.swift
tomholub: But isn't the whole point of `actor` that only one async method can happen in parallel?

Or are you saying that several methods can work in parallel, but they are synchronized on the same thread?

tomholub: I see - answer is yes - I got your answer below

FlowCrypt/Core/Core.swift
```diff
@@ -53,20 +53,21 @@ protocol KeyParser {
...
-    private dynamic var started = false
-    private dynamic var ready = false
-    
-    private let queue = DispatchQueue(label: "com.flowcrypt.core", qos: .default) // todo - try also with .userInitiated
+
+    private var callbackResults = [CallbackResult]()
+    private var ready = false
 
     private lazy var logger = Logger.nested(in: Self.self, with: "Js")
+    private lazy var resultSemaphore = DispatchSemaphore(value: 0)
```
ivan-ushakov, line 70: Unused variable

Core/source/assets/flowcrypt-ios-begin.js
```diff
@@ -35,8 +35,8 @@ const window = { navigator, crypto, setTimeout, clearTimeout };
 const self = { window, navigator, crypto };
 global.window = window;
 
-var engine_host_cb_value_formatter = function(val) {
-    coreHost.handleCallback(val.json, val.data);
+var engine_host_cb_value_formatter = function(key, val) {
+    coreHost.handleCallback(key, val.json, val.data);
 };
```
tomholub, line 40: please call it `callbackId` instead of `key`.

Since we deal with encryption keys everywhere, I'm trying to avoid the word in non-cryptographic contexts.

Core/source/entrypoint-bare.ts
```diff
@@ -11,17 +11,17 @@ declare const global: any;
 
 const endpoints = new Endpoints();
 
-global.handleRequestFromHost = (endpointName: string, request: string, data: Uint8Array, cb: (response: EndpointRes) => void): void => {
+global.handleRequestFromHost = (endpointName: string, endpointKey: string, request: string, data: Uint8Array, cb: (key: string, response: EndpointRes) => void): void => {
```
tomholub, line 14: please use `callbackId` in both usages

FlowCrypt/Core/Core.swift
```diff
@@ -230,18 +229,26 @@ actor Core: KeyDecrypter, KeyParser, CoreComposeMessageType {
     }
 
     private func call(_ endpoint: String, jsonData: Data, data: Data) async throws -> RawRes {
-        jsEndpointListener!.call(withArguments: [endpoint, String(data: jsonData, encoding: .utf8)!, Array<UInt8>(data), cb_catcher!])
+        guard ready else {
+            throw CoreError.exception("Core is not ready yet. Most likeyly startIfNotAlreadyRunning wasn't called first")
+        }
+
+        let endpointKey = NSUUID().uuidString
```
tomholub, line 236: `callbackId`

FlowCrypt/Core/CoreHost.swift
```diff
@@ -22,7 +22,7 @@ import SwiftyRSA // for rsa
     func setTimeout(_ callback: JSValue, _ ms: Double) -> String
     func clearTimeout(_ identifier: String)
 
-    func handleCallback(_ string: String, _ data: [UInt8])
+    func handleCallback(_ endpointKey: String, _ string: String, _ data: [UInt8])
```
tomholub, line 25: `callbackId`

FlowCrypt/Core/CoreHost.swift
```diff
@@ -119,9 +119,9 @@ final class CoreHost: NSObject, CoreHostExports {
         return uuid
     }
 
-    func handleCallback(_ string: String, _ data: [UInt8]) {
+    func handleCallback(_ endpointKey: String, _ string: String, _ data: [UInt8]) {
         Task {
-            await Core.shared.handleCallbackResult(json: string, data: data)
+            await Core.shared.handleCallbackResult(endpointKey: endpointKey, json: string, data: data)
         }
     }
```
tomholub: `callbackId`

tomholub (commented):
Looks great! Please rename some vals as below, else all good.

tomholub (approved):
excellent

PR #938, opened 2021-11-06
Convert remaining AttesterApiType from Promise to async

### LINKED ISSUE #934: convert remaining AttesterApiType from Promise to async
```swift
protocol AttesterApiType {
    func lookupEmail(email: String) async throws -> [KeyDetails]
    func updateKey(email: String, pubkey: String, token: String?) -> Promise<String>
    func replaceKey(email: String, pubkey: String) -> Promise<String>
    func testWelcome(email: String, pubkey: String) -> Promise<Void>
}
```

ivan-ushakov:
close #934

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
+++ b/FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
@@ -160,7 +160,8 @@ private actor Service {
             viewController: viewController
         )
 
-        _ = try await attester.testWelcome(
+        // sending welcome email is not crucial, so we don't handle errors
+        _ = try? await attester.testWelcome(
             email: userId.email,
             pubkey: encryptedPrv.key.public
         )
--- a/FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
+++ b/FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
@@ -9,7 +9,6 @@
 import AsyncDisplayKit
 import FlowCryptCommon
 import FlowCryptUI
-import Promises
 
 enum CreateKeyError: Error {
     case weakPassPhrase(_ strength: CoreRes.ZxcvbnStrengthBar)
@@ -29,7 +28,7 @@ enum CreateKeyError: Error {
  * - Here user can enter a pass phrase (can be saved in memory or in encrypted storage) and generate a key
  * - After key is generated, user will be redirected to **main flow** (inbox view)
  */
-
+@MainActor
 final class SetupGenerateKeyViewController: SetupCreatePassphraseAbstractViewController {
 
     private let backupService: BackupServiceType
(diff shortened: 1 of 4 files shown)


FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
```diff
@@ -144,25 +147,62 @@ private actor Service {
...
-        try awaitPromise(await viewController.alertAndSkipOnRejection(
-            updateKey,
-            fail: "Failed to submit Public Key")
-        )
-        let testWelcome = attester.testWelcome(email: userId.email, pubkey: encryptedPrv.key.public)
-        try awaitPromise(await viewController.alertAndSkipOnRejection(
-            testWelcome,
-            fail: "Failed to send you welcome email")
-        )
+    private func updateKeyAndWelcome(
```
tomholub, line 164: This is too complicated. You can remove the error handling/alert code in this case, and simply add two attester calls into `setupAccount`

FlowCrypt/Functionality/Services/Remote Pub Key Services/AttesterApi.swift
```diff
@@ -91,37 +92,18 @@ extension AttesterApi {
...
             headers: headers
         )
-        return Promise { () -> String in
-            let res = try awaitPromise(ApiCall.call(request))
-            return res.data.toStr()
-        }
-    }
-
-    @discardableResult
-    func replaceKey(email: String, pubkey: String) -> Promise<String> {
```
tomholub, line 101: While not used now, this will actually be needed later: on consumer usecases (that don't use the EKM), if the public key on Attester has different fingerprint than the public key being submitted, we'll have to use `replaceKey` instead of `updateKey`. I'll file an issue for this.

No need to add it back now, I can just copy/paste it below:

```
 @discardableResult
    func replaceKey(email: String, pubkey: String) -> Promise<String> {
        let request = ApiCall.Request(
            apiName: Constants.apiName,
            url: urlPub(emailOrLongid: email),
            method: .post,
            body: pubkey.data()
        )
        return Promise { () -> String in
            let res = try awaitPromise(ApiCall.call(request))
            return res.data.toStr()
        }
        let res = try await ApiCall.asyncCall(request)
        return res.data.toStr()
    }
```

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
ivan-ushakov: So we don't notify user about error here?

tomholub: I had a second look. For `attester.testWelcome`, we don't have to notify user about errors (it's low importance - this sends user a welcome message, which is not crucial). Can just directly add in `setupAccount`:

```ts
// sending welcome email is not crucial, so we don't handle errors
_ = try? await attester.testWelcome(email: email, pubkey: publicKey)
```

and remove `testWelcome` method. Then let's rename `updateKeyAndWelcome` to `submitKeyToAttesterAndShowAlertOnFailure`, and remove `testWelcome` from that. That should clean things up nicely.

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
```diff
@@ -144,27 +147,42 @@ private actor Service {
...
-            fail: "Failed to submit Public Key")
-        )
-        let testWelcome = attester.testWelcome(email: userId.email, pubkey: encryptedPrv.key.public)
-        try awaitPromise(await viewController.alertAndSkipOnRejection(
-            testWelcome,
-            fail: "Failed to send you welcome email")
+        _ = try await attester.testWelcome(
+            email: userId.email,
+            pubkey: encryptedPrv.key.public
         )
```
tomholub, line 166: I suppose this way if this throws, the whole async function will fail? As I mentioned earlier, in this one case I'd prefer a `try?`, including a // comment I mentioned earlier above, explaining why we chose `try?` this time.

tomholub (approved):

Issue #959, opened 2021-11-09
ui tests for thread rendering


tomholub:
- test that date of last message in the thread is displayed on each inbox item
 - make sure that messages are ordered by this date in inbox, with threads that have the most recent message on top

fcvakintos:
@tomholub is it correct?
email with date 22 Nov is displayed under email with date Nov 03, 2017
<img width="389" alt="Screenshot 2021-11-30 at 12 47 43" src="https://user-images.githubusercontent.com/91205853/144033968-332c8633-ba15-4172-97d7-f60446081d3a.png">
<img width="389" alt="Screenshot 2021-11-30 at 12 48 03" src="https://user-images.githubusercontent.com/91205853/144034048-3ad043b7-dfb4-4317-bad4-46d1089c7d49.png">

tomholub:
Well noticed btw

fcvakintos:
@tomholub login to e2e.enterprise.test@flowcrypt.com
and check inbox screen

fcvakintos:
@tomholub check this pls
<img width="403" alt="Screenshot 2021-11-30 at 13 30 03" src="https://user-images.githubusercontent.com/91205853/144039725-2d2ff265-fd19-4504-a00c-97ef5920b1b1.png">

tomholub:
Looks wrong indeed

tomholub:
@fcvakintos in this issue you can at least check order of messages within a particular email thread, instead of the inbox where we're still looking into potentially fixing it.

fcvakintos:
@tomholub can you confirm is it correct thread?
 I'm not sure to understand the task
 
<img width="433" alt="Screenshot 2022-02-04 at 10 52 05" src="https://user-images.githubusercontent.com/91205853/152499727-77fde7cb-72fe-4424-9d56-56a0231d1190.png">

tomholub:
Yes, this is an example of one email thread. There are several messages in it. You want to make sure they start from oldest to newest, and that the `to` and `from` make it clear who sent each message to whom.

fcvakintos:
ok, thanks

fcvakintos:
@tomholub we need to add new accessibility id for thread messages
something like we did for recipients
who from devs can do this?

fcvakintos:
and we need to add accessibility id for message body(text)
[...]

PR #961, opened 2021-11-09
issues 948 950 960 fix inbox conversations data

### LINKED ISSUE #948: Gmail provider error: the request timed out
I'm getting these relatively frequently in inbox lately.

What is the timeout on these requests? maybe it can be adjusted

Alternatively, are there many requests being called when loading inbox? Or is it one single request?

### LINKED ISSUE #950: certain inbox items are missing after the conversations update
I went to an email that was from:me to:me and I replied to it from the iOS app. Then I went to sent messages, and could not find it. Also could not find it in inbox in the app.

But I could easily find it in both browser webmail and previous version of the iOS app without threads.

However, I was not able to reproduce it on another conversation following the steps above (also tried replying to already archived message). So while I can observe it on one particular conversation anytime, I don't know how to reproduce this for other conversations.

### LINKED ISSUE #960: wrong date on thread inbox items
shows date of first message, should be date of last message

tomholub:
This PR:
 - close #960 - was date of first msg instead of last
 - close #948 - reduced concurrency
 - close #950 - just wrong ordering
 - renames various vars and methods for clarity

`for try await result in taskGroup` seems to execute in order that the tasks finish, not necessarily in the order in which the `taskGroup` array was defined. Instead of using an array to collect results, using a dict by id now, which is then converted to array in expected order.

When the order is kept correct that way, `.sort()` is no longer needed on the result. The sort was broken anyway, sorting by first message in thread. 

**Tests**:
 - tests will be added in https://github.com/FlowCrypt/flowcrypt-ios/issues/959

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Inbox/InboxRenderable.swift
+++ b/FlowCrypt/Controllers/Inbox/InboxRenderable.swift
@@ -25,12 +25,6 @@ struct InboxRenderable {
     let wrappedType: WrappedType
 }
 
-extension InboxRenderable: Comparable {
-    static func < (lhs: InboxRenderable, rhs: InboxRenderable) -> Bool {
-        lhs.date > rhs.date
-    }
-}
-
 extension InboxRenderable {
     var wrappedMessage: Message? {
         guard case .message(let message) = wrappedType else {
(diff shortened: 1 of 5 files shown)


FlowCrypt/Controllers/Inbox/InboxViewController.swift
```diff
@@ -276,7 +276,7 @@ extension InboxViewController {
         if input.data.isEmpty {
             state = .empty
         } else {
-            inboxInput = input.data.sorted()
+            inboxInput = input.data
```
tomholub, line 279: The fix is here

FlowCrypt/Controllers/Inbox/InboxRenderable.swift
```diff
@@ -56,23 +50,19 @@ extension InboxRenderable {
...
-
         self.title = sender
         self.messageCount = thread.messages.count
         self.subtitle = thread.subject ?? "message_missed_subject".localized
-
-        if let date = thread.messages.first?.date {
+        self.isRead = !thread.messages
+            .map(\.isMessageRead)
+            .contains(false)
+        let date = thread.messages.last?.date
```
tomholub, line 59: The fix is here

FlowCrypt/Functionality/Mail Provider/Threads/MessagesThreadProvider.swift
```diff
@@ -25,20 +24,21 @@ extension GmailService: MessagesThreadProvider {
...
                 }
             }
             for try await result in taskGroup {
-                messages.append(result)
+                if let id = result.identifier {
+                    messageThreadsById[id] = result
+                }
             }
-
+            let messageThreads = requests.compactMap { messageThreadsById[$0.0] }
```
tomholub, line 39: The fix is here

sosnovsky (approved):
Works good 👍

PR #967, opened 2021-11-09
issue 966 always return on continuation.resume

### LINKED ISSUE #966: continuation.resume should always return
image

if it doesn't return, there's a risk another continuation.resume may run and crash the app

tomholub:
This PR always adds a return before `continuation.resume` even if it's not needed there.

While it may look like an overkill, if we make this a convention, we will ensure to never run into the same bug again, and it will be easier to catch these bugs in PR reviews even without full context of the surrounding code. Missing return -> code smell, even if the code is otherwise correct.

close #966

**Tests**:

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Mail Provider/Threads/MessagesThreadOperationsProvider.swift
+++ b/FlowCrypt/Functionality/Mail Provider/Threads/MessagesThreadOperationsProvider.swift
@@ -91,9 +91,9 @@ extension GmailService: MessagesThreadOperationsProvider {
 
             self.gmailService.executeQuery(query) { _, _, error in
                 if let error = error {
-                    continuation.resume(throwing: GmailServiceError.providerError(error))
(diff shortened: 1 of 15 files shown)


FlowCrypt/Functionality/Mail Provider/Threads/MessagesThreadOperationsProvider.swift
```diff
@@ -91,9 +91,9 @@ extension GmailService: MessagesThreadOperationsProvider {
 
             self.gmailService.executeQuery(query) { _, _, error in
                 if let error = error {
-                    continuation.resume(throwing: GmailServiceError.providerError(error))
+                    return continuation.resume(throwing: GmailServiceError.providerError(error))
                 }
-                continuation.resume(returning: ())
+                return continuation.resume(returning: ())
             }
```
tomholub, line 97: This was the bug that made me do this

sosnovsky (approved):

PR #983, opened 2021-11-10
Rename all realm objects to end with RealmObject

### LINKED ISSUE #979: Rename all realm objects to end with RealmObject
First, we should rename all realm objects to end with `RealmObject`, eg `PrivateKeyRealmObject`. That will make them stand out.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/877#issuecomment-964889697_

ivan-ushakov:
close #979

### To be filled by reviewers


tomholub:
@ivan-ushakov does any of this cause a need for migration? Eg the app would now store it in a differently named table?

ivan-ushakov:
> @ivan-ushakov does any of this cause a need for migration? Eg the app would now store it in a differently named table?

Good catch. I will check this too.

tomholub (approved):

ivan-ushakov:
@tomholub 
Looks like class renaming requires migration. To avoid this we could try to move Realm objects to "namespace", like:
```
enum RealmObjects {
}

extension RealmObjects {
    class MyClass: Object {
    }
}

```

tomholub:
> @tomholub Looks like class renaming requires migration. To avoid this we could try to move Realm objects to "namespace", like:
> 
> ```
> enum RealmObjects {
> }
> 
> extension RealmObjects {
>     class MyClass: Object {
>     }
> }
> ```

As long as we always have to access it with `RealmObject.MyClass` than that's fine. Actually could just call `Db.MyClass` ? Or `Table.MyClass` ?

ivan-ushakov:
@tomholub 
Namespace approach will not work:

```
*** Terminating app due to uncaught exception 'RLMException', reason: 'RLMObject subclasses cannot be nested within other declarations. Please move _TtCO9FlowCrypt2Db25ClientConfigurationObject to global scope.'
terminating with uncaught exception of type NSException
```

So looks like only migration is possible

tomholub:
> @tomholub Namespace approach will not work:
> 
> ```
> *** Terminating app due to uncaught exception 'RLMException', reason: 'RLMObject subclasses cannot be nested within other declarations. Please move _TtCO9FlowCrypt2Db25ClientConfigurationObject to global scope.'
> terminating with uncaught exception of type NSException
> ```
> 
> So looks like only migration is possible

Ok, let's do it then, with a migration. We'd have to clean it up one day either way.

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -35,11 +35,14 @@ final class EncryptedStorage: EncryptedStorageType {
...
     private enum EncryptedStorageSchema: CaseIterable {
         case initial
+        case version5
 
         var version: SchemaVersion {
             switch self {
             case .initial:
                 return SchemaVersion(appVersion: "0.2.0", dbSchemaVersion: 4)
+            case .version5:
+                return SchemaVersion(appVersion: "0.2.0", dbSchemaVersion: 5)
```
ivan-ushakov, line 45: @tomholub 
Is `appVersion` OK?

FlowCrypt/Models/Realm Models/ClientConfigurationRealmObject.swift
```diff
@@ -9,16 +9,15 @@
...
-    @objc dynamic var flags: Data?
-    @objc dynamic var customKeyserverUrl: String?
-    @objc dynamic var keyManagerUrl: String?
-    @objc dynamic var disallowAttesterSearchForDomains: Data?
-    @objc dynamic var enforceKeygenAlgo: String?
-    @objc dynamic var enforceKeygenExpireMonths: Int = -1
-    @objc dynamic var user: UserObject!
-    @objc dynamic var userEmail: String!
+final class ClientConfigurationRealmObject: Object {
+    @Persisted(primaryKey: true) var userEmail: String!
```
ivan-ushakov, line 13: I decided to change all `@objc dynamic` to `@Persisted` to have consistency

FlowCrypt/Models/Realm Models/RecipientRealmObject.swift
```diff
@@ -9,21 +9,11 @@
 import Foundation
 import RealmSwift
 
-final class LongId: Object {
```
ivan-ushakov, line 12: @tomholub 
I checked source code and we don't use this type. Maybe I'm wrong and you need to keep this one

tomholub: If we don't use it, throw it out :+1:

tomholub:
@sosnovsky will need to confirm that the migration is smooth when testing an upgrade from master to this branch

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
tomholub: It's outdated. Are we supposed to keep it up to date? (and for what purpose?) If yes, could you put some dynamic value in there?

FlowCrypt/Models/Realm Models/KeyInfoRealmObject.swift
```diff
@@ -19,23 +19,22 @@ enum KeyInfoError: Error {
...
+    @Persisted var user: UserRealmObject!
+
     var primaryLongid: String {
         allLongids[0]
     }
 
-    @objc dynamic var `private`: String = ""
-    @objc dynamic var `public`: String = ""
-
     let allFingerprints = List<String>()
```
ivan-ushakov, line 34: Looks like I did mistake here because of this:

> To declare properties of generic types LinkingObjects, List, and RealmProperty, use let. Generic properties cannot be represented in the Objective‑C runtime, which Realm Database uses for dynamic dispatch of dynamic properties.

So this properties must be persisted too. To avoid such mistake in future we must use `@Persisted`

ivan-ushakov: This is the reason why UI tests failed and this is good to UI test

tomholub: Yup. You can only do limited testing with unit tests. They won't keep you safe.

FlowCrypt/Functionality/DataManager/Encrypted Storage/Version5SchemaMigration.swift
```diff
@@ -0,0 +1,247 @@
...
+
+                do {
+                    try object.1(oldObject)
+                } catch {
+                    lastError = error
+                }
+            }
+        }
+
+        [
```
ivan-ushakov, line 55: Now we remove old objects

FlowCryptAppTests/Core/Models/KeyInfoTests.swift
```diff
@@ -117,7 +117,6 @@ class KeyInfoTests: XCTestCase {
         XCTAssertTrue(key.primaryFingerprint == "f1")
         XCTAssertTrue(key.primaryLongid == "l1")
 
-        XCTAssertTrue(KeyInfo.primaryKey() == "primaryFingerprint")
```
ivan-ushakov, line 120: No need for this check since we moved to `@Persisted`

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
ivan-ushakov: We don't use this property in any logic

sosnovsky (approved):
Tested migration from `master` build to this one - worked smoothly, no errors or issues with the app, can be merged ✅

PR #984, opened 2021-11-10
Removed sleep 10s in login method, changed it to loop which waiting field

### LINKED ISSUE #825: appium: evaluate alternatives to 10s sleep after login
:-) I'll file this in a new issue

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/823#discussion_r738587438_

fcvakintos:
This PR contains changes for login method, removed sleep 10s, added waitForDispalayed({reverse: true})
close #825 

**Tests** 
- Tests added or updated

### To be filled by reviewers


appium/tests/screenobjects/splash.screen.ts
```diff
@@ -156,7 +156,7 @@ class SplashScreen extends BaseScreen {
         this.clickContinueBtn();
         this.changeLanguage();
         this.gmailLogin(email, password);
-        this.signInAsGoogleAccounLabel.waitForDisplayed({reverse: true});
+        this.signInAsGoogleAccounLabel.waitForDisplayed({timeout: 10000, reverse: true});
```
tomholub: If various different screens are being switched on the way, you may need to try/wait/retry as I suggested to avoid the flakiness.

fcvakintos:
@tomholub finished with this task, run tests and no error (4/4)

appium/tests/screenobjects/create-key.screen.ts
```diff
@@ -31,6 +31,16 @@ class CreateKeyScreen extends BaseScreen {
...
+        // retrying several times because following login, we switch
+        //   from webview to our own view and then to another one several
+        //   times, which was causing flaky tests. Originally we did a 10s
+        //   delay but now instead we're retrying once per second until
+        //   we see what we expect.
+        let count = 0;
+        do {
+            browser.pause(1000);
+            count++;
+        } while(this.enterPassPhraseField.isDisplayed() !== true || count <= 15);
```
tomholub, line 43: I don't think this does what you think it does. It looks to me it will try 15 times regardless if it showed before or not. Have you tested how many iterations it does?

fcvakintos: @tomholubno, we have "||", it is checking one of arguments is true, if the filed is displayed on 5th iteration, we break from loop, but if the field is not displayed after 15 iterations we finish loop and go to next step
I will check how many iterations appear

fcvakintos: @tomholub sorry, You was right, I changed || to &&

fcvakintos: @tomholub checked iterations count
we have max 3 iterations
Screenshot 2021-11-12 at 11 02 20

tomholub: That looks better

tomholub (approved):

Issue #830, opened 2021-10-28
clarify contactsService.retrievePubkeys name + signature


tomholub:
Neither `contactsService` nor `retrievePubkeys` names make it obvious to know if the pubkeys come from local storage or from remote server. It would be good to rename it to make it obvious from the name. I'll look into that in another issue.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/821#discussion_r738623012_

tomholub:
<img width="650" alt="image" src="https://user-images.githubusercontent.com/6306961/141460295-b51119db-9658-4105-9a71-9c64eb31f983.png">

tomholub:
This will get resolved by #1175

PR #998, opened 2021-11-11
Feature/issue 278 verify signed messages

### LINKED ISSUE #278: verify signed messages
This is part of #275. Sister issues on Android: https://github.com/FlowCrypt/flowcrypt-android/issues/1096 and https://github.com/FlowCrypt/flowcrypt-android/issues/1097. First need to do #609

### LINKED ISSUE #1021: findBy:longid unused?
I don't think this is used anywhere in this PR - unless I overlooked

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/998#discussion_r749675549_

sosnovsky:
This PR adds verification for signed messages

close #278
close #1021

- Difficult to test (explain why) - currently we don't have any tests for `MessageService`

### To be filled by reviewers


FlowCrypt/Controllers/Threads/ThreadDetailsDecorator.swift
```diff
@@ -11,6 +11,12 @@ import UIKit
 
 extension ThreadMessageSenderCellNode.Input {
     init(threadMessage: ThreadDetailsViewController.Input) {
+        let signature: NSAttributedString?
+        if let processedMessage = threadMessage.processedMessage, processedMessage.messageType != .plain {
+            signature = NSAttributedString.text(from: processedMessage.signature.message, style: .regular(12), color: .white)
```
tomholub, line 16: Does setting `.white` potentially cause trouble with dark mode?

FlowCrypt/Functionality/Mail Provider/Message Provider/MessageService.swift
```diff
@@ -178,25 +208,46 @@ final class MessageService {
...
             rawMimeData: rawMimeData,
             text: text,
             attachments: attachments,
-            messageType: messageType
+            messageType: messageType,
+            signature: signature
         )
     }
 
+    private func verifySignature(longid: String?, email: String?) async -> ProcessedMessage.MessageSignature {
```
tomholub, line 235: let's call it `evaluateSignatureVerificationResult`

Also, instead of just `longid`, give it whole `VerifyRes?`.

Also edit `VerifyRes` type - it may have an `error` on it of type `String?`.

Here is equivalent method from the browser: https://github.com/FlowCrypt/flowcrypt-browser/blob/fc8db5bf7f05e6548459f463fe672c4e09c34e20/extension/chrome/elements/pgp_block_modules/pgp-block-signature-module.ts#L24

Below is pseudo code for you in the context of this swift app.

```ts
  private func renderPgpSignatureCheckResult = (
    signature: VerifyRes?, 
    sender: RecipientWithOrderedPublicKeys
  ) -> ProcessedMessage.MessageSignature {
    if signature is nil then return .unsigned // render "Message not signed" in red
    if signature.error is not nil then return .error(signature.error) // render the error in red
    if signature.signer is nil then return .unsigned // render "Message not signed" in red
    if signature.match is nil then return .unsigned // render "Message not signed" in red
    if signature.match is false then return .bad // render "Bad signature" in red
    let signingPubKey = sender.pubkeys.find { $0.longids.contains(signature.signer) }
    if signingPubKey is nil then return .missingPubkey( 
       longid: signature.signer, 
       senderEmail: sender.email
     ) // render "Missing public key \(longid) to verify signature from \(senderEmail)" in red
     return .good
  }
```

tomholub (commented):
Thanks! See details about how to understand the signature result.

FlowCrypt/Controllers/Threads/ThreadDetailsDecorator.swift
tomholub: I can see now it's on a colorful background, all good

tomholub:
Rendering is well done - thank you. I gave you vague instructions but it turned out well 👍 just please put the green bubble below the date.

```
tom@flowcrypt.com       <-   ^
Nov 08
[🔒 signed]
```

Please have it say just `signed` instead of `Signed by xxx` in the good case. Since we already check for email match.

tomholub:
And also let's do what you suggested:

```
tom@flowcrypt.com       <-   ^
Nov 08
[🔒 encrypted] [🔒 signed]
```

FlowCrypt/Functionality/Mail Provider/Message Provider/MessageService.swift
sosnovsky: Thanks for detailed explanation 👍
I checked this browser code during development too, to see how it renders signature block, it was very helpful.

sosnovsky:
> Then please make sure it collapses together with the text when I collapse that message. Unless you think it shouldn't?

Yes, I missed collapsing part, it definitely should collapse along with other message parts.

FlowCrypt/Functionality/Mail Provider/Message Provider/MessageService.swift
```diff
@@ -215,12 +217,14 @@ final class MessageService {
             let err = decryptErrBlock.decryptErr?.error
             text = "Could not decrypt:\n\(err?.type.rawValue ?? "UNKNOWN"): \(err?.message ?? "??")\n\n\n\(rawMsg)"
             messageType = .error(err?.type ?? .other)
-            signature = .unknown
+            signature = .error(rawMsg)
```
tomholub: I'd put `.error("processing error")`?. Since the the `text` already contains the issue description (that should be rendered in text)

FlowCrypt/Functionality/Mail Provider/Message Provider/MessageService.swift
```diff
@@ -288,6 +276,44 @@ final class MessageService {
...
+
+        guard let signer = signature.signer else { return .unsigned }
+
+        var pubKey: PubKey?
+
+        if let contact = await contactsService.findBy(longId: signer) {
+            pubKey = contact.pubKey(with: signer)
+        } else if let email = sender, let contact = try? await contactsService.searchContact(with: email) {
+            pubKey = contact.pubKey(with: signer)
+        }
```
tomholub, line 307: Now I'm thinking, this shouldn't even be needed anymore - since we already provided the right pubkey to TypeScript. It would not come out as `match` unless it had the right pubkey to verify.

Therefore you can remove this `pubKey` code from this method, including `pubKey != nil` below. In fact, can also remove `sender` from this method signature.

tomholub (commented):
Looking better 👍

FlowCrypt/Functionality/Mail Provider/Message Provider/MessageService.swift
sosnovsky: We can remove first part `contactsService.findBy(longId: signer)`, but second part `contactsService.searchContact(with: email)` is needed for the case when mail sender isn't stored in local user contacts and we need to fetch data using `pubLookup.lookup(with: email)`

For example, I reinstalled app and my local contacts list is empty, then verification for all messages will fail with error `no public key`

Or I'm missing something and there is another way to get sender's public key?

sosnovsky: I think I found a bug here - this lookup should be performed before message validation, if app doesn't have sender's pubkey. Fetching here is too late, as we've already tried to validate message.
I'll remove pubkey fetch here and move it to previous validation step.

tomholub: There are performance concerns - it may take too long. So I was thinking to fetch in parallel with downloading the message. That's why I did #1002 (pub lookup itself was slow) and also separated "fetch if pubkeys missing" part into #1001

But if you merge in #1002 and then also do #1001 in this PR, that would be ok with me. Otherwise can wait for later - up to you.

sosnovsky: I update `fetchVerificationPubKeys` function to fetch pub keys in case when there are no local pub keys - https://github.com/FlowCrypt/flowcrypt-ios/pull/998/commits/e304d431539c2d5bb7634672ca01ae110fae9e0d#diff-a1ec6a13dae98c2b61953533ba95947cc194b42dce60f4d2c905ef5ea9df1b64R280
So if keys are already stored locally then there won't be any delay for user.

FlowCrypt/Functionality/Mail Provider/Message Provider/MessageService.swift
```diff
@@ -278,35 +277,28 @@ final class MessageService {
...
-        if let contact = await contactsService.findBy(longId: signer) {
-            pubKey = contact.pubKey(with: signer)
-        } else if let email = sender, let contact = try? await contactsService.searchContact(with: email) {
-            pubKey = contact.pubKey(with: signer)
-        }
-
-        guard pubKey != nil && signature.match != nil else { return .missingPubkey(signer) }
+        guard signature.match != nil else { return .missingPubkey(signer) }
 
         guard signature.match == true else { return .bad }
```
tomholub: Just realized - there is more than one kind of `.good` :-) TypeScript code for reference:

```ts
  if (verifyRes && verifyRes.match) {
    if (mixedSignatures) {
      verifyRes.mixed = true;
    }
    if (signedBlockCount > 0 && signedBlockCount != contentBlocks.length) {
      verifyRes.partial = true;
    }
  }
```

That means `VerifyRes` needs to be updated to also include `mixed: Bool?` and `partial: Bool?`. Plus this logic here, too. But that's for another issue, needs more thought. It won't be a difficult change once thought through.

FlowCrypt/Functionality/Mail Provider/Message Provider/MessageService.swift
```diff
@@ -278,35 +277,28 @@ final class MessageService {
...
+        guard let sender = sender else { return [] }
+
+        let pubKeys = contactsService.retrievePubKeys(for: sender)
+        if pubKeys.isNotEmpty { return pubKeys }
+
+        guard let contact = try? await contactsService.searchContact(with: sender)
+        else { return [] }
+
+        return contact.activePubKeys.map(\.armored)
     }
```
tomholub: This really highlights the need to (later) be clearer in differentiating local and remote lookups. Because from the code, I can't tell. #830

FlowCrypt/Functionality/Mail Provider/Message Provider/MessageService.swift
```diff
@@ -278,35 +277,28 @@ final class MessageService {
...
+    private func fetchVerificationPubKeys(for sender: String?) async throws -> [String] {
+        guard let sender = sender else { return [] }
+
+        let pubKeys = contactsService.retrievePubKeys(for: sender)
+        if pubKeys.isNotEmpty { return pubKeys }
+
+        guard let contact = try? await contactsService.searchContact(with: sender)
+        else { return [] }
+
+        return contact.activePubKeys.map(\.armored)
```
tomholub, line 289: I'm not too sure about the .activePubkeys here. I'd give it all of them. I guess, an expired public key is still a good public key to verify a message that was made before the key expired. But that's for TypeScript to decide, not here.

tomholub: Also, by also giving it revoked pubkeys, we give TypeScript a chance to say the signature is bad. But if we withold the revoked pubkeys, TypeScript will say that the right pubkey is missing - which is different.

FlowCrypt/Functionality/Mail Provider/Message Provider/MessageService.swift
tomholub: This works as a compromise. We'll have to do something better in the future.

Because public keys can change over time. So just because I have a pubkey for someone, doesn't mean that they don't have a new key (eg keys may get rotated once a year in some orgs) and that they didn't sign that message with the new key that I don't have.

But as it is now, it settles 99% of cases which will do for now.

FlowCrypt/Functionality/Mail Provider/Message Provider/MessageService.swift
sosnovsky: Got it, I'll change it to `contact.pubKeys.map(\.armored)`

sosnovsky:
@tomholub I found an issue with non-encrypted messages here https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-278-verify-signed-messages/FlowCrypt/Functionality/Mail%20Provider/Message%20Provider/MessageService.swift#L168

In case when user receives plain non-encrypted mail then app tries to decrypt it, but can't find local pub keys for mail sender, so it sends request to `https://flowcrypt.com/attester/pub/` and receives 404 error. But this request takes some time and delays showing message to user.

Is there any method to find if message needs decryption or it's just plain text?
I thought about checking if message starts with `\r\n\r\n-----BEGIN PGP MESSAGE-----\r\n`, but it doesn't seem to be a reliable solution, or it should work well?

tomholub:
> @tomholub I found an issue with non-encrypted messages here https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-278-verify-signed-messages/FlowCrypt/Functionality/Mail%20Provider/Message%20Provider/MessageService.swift#L168
> 
> In case when user receives plain non-encrypted mail then app tries to decrypt it, but can't find local pub keys for mail sender, so it sends request to `https://flowcrypt.com/attester/pub/` and receives 404 error. But this request takes some time and delays showing message to user.
> 
> Is there any method to find if message needs decryption or it's just plain text? I thought about checking if message starts with `\r\n\r\n-----BEGIN PGP MESSAGE-----\r\n`, but it doesn't seem to be a reliable solution, or it should work well?

Not reliably. This is why I was thinking to fetch the message from Gmail, and fetch the pubkey, in parallel. 

But fetching a public key from Attester can be even slower, especially when it's a 404 (and Attester itself is trying to fetch them from remote places).

The proper way to do all of this, I think, is:

1) user taps a message
2) message gets downloaded from Gmail API
3) message gets processed / decrypted / verified with whatever public keys we have locally
4) result gets rendered including signature verification, and if we are not missing any public keys, then this is the last step
5) if we are missing a public key, the verification "badge" will show gray `🕒 verifying signature..`
6) at this point we fetch remote public keys, but user is no longer waiting for it
7) after remote public keys are fetched, we check if any of the keys now contain the `signer` longid. If they don't, now we render missing pub key
8) if one of the keys now contain the signer longid, we change  we re-process the whole message with that key. But only in background. The badge still says `🕒 verifying signature`. Once we've done re-processing the message, we only look at `verifyRes` and re-render the badge with the final result.

tomholub:
sometimes it gets confused when a message is signed but not encrypted (see email for how to reproduce):

image

FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
```diff
@@ -262,6 +279,18 @@ extension ThreadDetailsViewController {
         }
     }
 
+    private func tryToFetchSenderPubKeys(message: Message, folder: String, indexPath: IndexPath) {
+        Task {
+            let processedMessage = try await messageService.getAndProcessMessage(
```
tomholub, line 284: What happens to exceptions that are thrown in `Task` and not catched? Do they crash the app or are they ignored? Any logs?

Either way, some exception handling will be necessary. If lookup or something else fails, the signature badge will be stuck in loading state forever. For these situations, we may need a badge like `Failed to verify signature due to: \(error)`.

tomholub (commented):
Looking good from code - see comments

tomholub (commented):
This looks good by code. What is the status of the PR - did you want to do more testing, or something left to do?

sosnovsky:
I want to test it with `flowcrypt.compatibility@gmail.com` account, as you mentioned there is an issue with signed but not encrypted messages. But as it's quite rare case we can leave it for another issue, is it okay?

tomholub:
> I want to test it with `flowcrypt.compatibility@gmail.com` account, as you mentioned there is an issue with signed but not encrypted messages. But as it's quite rare case we can leave it for another issue, is it okay?

I'll file another issue for it, not necessary to do now.

FlowCrypt/Controllers/Msg/MessageViewController.swift
```diff
@@ -0,0 +1,442 @@
+//
```
tomholub, line 1: Maybe the this VC was brought back in by mistake?

FlowCrypt/Controllers/Threads/ThreadDetailsDecorator.swift
```diff
@@ -11,6 +11,12 @@ import UIKit
 
 extension ThreadMessageSenderCellNode.Input {
     init(threadMessage: ThreadDetailsViewController.Input) {
+        let signature: String
+        if let processedMessage = threadMessage.processedMessage, processedMessage.messageType == .encrypted {
+            signature = processedMessage.signature.message
+        } else {
+            signature = "message_not_signed".localized
+        }
```
tomholub, line 19: This is likely the reason why signed-only messages are not recognized as signed - per above code, if they are not encrypted, they are always rendered as "not signed".

Would it be possible to always use `processedMessage.signature.message` regardless if encrypted or not? If that is missing, sounds like an error as well (since when signature is missing, `processedMessage.signature.message` is supposed to be there with `notSigned` case or similar)

Can address later.

FlowCrypt/Controllers/Msg/MessageViewController.swift
sosnovsky: Yes, it's reference was deleted from the project, but file was still stored in repository, removed it.

FlowCrypt/Functionality/Services/Local Pub Key Services/LocalContactsProvider.swift
```diff
@@ -42,8 +43,17 @@ extension LocalContactsProvider: LocalContactsProviderType {
     }
 
     func retrievePubKeys(for email: String) -> [String] {
-        find(with: email)?.pubKeys
-            .map { $0.armored } ?? []
+        find(with: email)?.pubKeys.map(\.armored) ?? []
+    }
+
+    func findBy(longid: String) async -> RecipientWithSortedPubKeys? {
```
tomholub, line 49: I don't think this is used anywhere in this PR - unless I overlooked

tomholub (approved):

FlowCrypt/Controllers/Threads/ThreadDetailsDecorator.swift
tomholub: #1020

FlowCrypt/Functionality/Services/Local Pub Key Services/LocalContactsProvider.swift
sosnovsky: Yes, it was used in initial code for signature verification, but not needed anymore, removed it

tomholub (approved):

FlowCrypt/Controllers/Threads/ThreadDetailsDecorator.swift
sosnovsky: Yes, initially I thought that all signed messages are encrypted, and haven't implemented signature state check for this case.

tomholub: Check the browser extension - next to the send button, there is a chevron. That's how a signed-only message could be sent.

sosnovsky: Oh, didn't know about such feature 👍

PR #1002, opened 2021-11-11
issue 999 parallelize pubkey lookup

### LINKED ISSUE #999: parallelize pubkey lookups
WKD advanced and WKD direct should happen in parallel
WKD and Attester (if allowed) lookups should happen in parallel, but WKD should continue to be preferred when found on both

tomholub:
This PR parallelizes pubkey lookup to speed it up.

close #999

**Tests**:

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Services/Remote Pub Key Services/AttesterApi.swift
+++ b/FlowCrypt/Functionality/Services/Remote Pub Key Services/AttesterApi.swift
@@ -46,7 +46,7 @@ final class AttesterApi: AttesterApiType {
 }
 
 extension AttesterApi {
-    func lookupEmail(email: String) async throws -> [KeyDetails] {
+    func lookupEmail(_ email: String) async throws -> [KeyDetails] {
         if !(try clientConfiguration.canLookupThisRecipientOnAttester(recipient: email)) {
             return []
         }
--- a/FlowCrypt/Functionality/Services/Remote Pub Key Services/AttesterApi.swift
+++ b/FlowCrypt/Functionality/Services/Remote Pub Key Services/AttesterApi.swift
@@ -10,7 +10,7 @@ struct PubkeySearchResult {
 }
 
 protocol AttesterApiType {
-    func lookupEmail(email: String) async throws -> [KeyDetails]
+    func lookupEmail(_ email: String) async throws -> [KeyDetails]
     func updateKey(email: String, pubkey: String, token: String?) async throws -> String
     func replaceKey(email: String, pubkey: String) async throws -> String
     func testWelcome(email: String, pubkey: String) async throws
--- a/FlowCrypt/Functionality/Services/Remote Pub Key Services/AttesterApi.swift
+++ b/FlowCrypt/Functionality/Services/Remote Pub Key Services/AttesterApi.swift
@@ -4,15 +4,10 @@
 
 import Foundation
 
-struct PubkeySearchResult {
-    let email: String
-    let armored: Data?
-}
-
 protocol AttesterApiType {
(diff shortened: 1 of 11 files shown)


tomholub:
As implemented now, if Attester call fails (server down), I think it would throw even if WKD call was successfully returning a key. For now that's not bothering me since primary customer has Attester disabled using the `disallow_attester_search_for_domains=[*]` rule. But it's something that should be addressed later.

FlowCrypt/Functionality/Services/Remote Pub Key Services/AttesterApi.swift
```diff
@@ -10,7 +10,7 @@ struct PubkeySearchResult {
 }
 
 protocol AttesterApiType {
-    func lookupEmail(email: String) async throws -> [KeyDetails]
+    func lookupEmail(_ email: String) async throws -> [KeyDetails]
```
sosnovsky, line 13: We can call it just `func lookup(email: String)` instead of `func lookupEmail(_ email: String)`, so it's more Swift-style and we don't need to use underscore.

FlowCrypt/Functionality/Services/Remote Pub Key Services/PubLookup.swift
```diff
@@ -14,6 +14,16 @@ class PubLookup: PubLookupType {
     private let wkd: WkdApiType
     private let attesterApi: AttesterApiType
 
+    private enum PubLookupSource {
+        case attester
+        case wkd
+    }
+
+    private struct Result {
```
sosnovsky, line 22: Swift already has built-in `Result` type (https://developer.apple.com/documentation/swift/result), it'll be better to rename this one to something like `LookupResult`

FlowCrypt/Functionality/Services/Remote Pub Key Services/PubLookup.swift
```diff
@@ -23,12 +33,32 @@ class PubLookup: PubLookupType {
...
         }
-
-        let attesterResult = try await attesterApi.lookupEmail(email: email)
-        return RecipientWithSortedPubKeys(email: email, keyDetails: attesterResult)
+        guard let wkdResult = results.first(where: { $0.source == .wkd }) else {
+            throw AppErr.unexpected("expecting to find wkdResult")
+        }
+        guard let attesterResult = results.first(where: { $0.source == .attester }) else {
+            throw AppErr.unexpected("expecting to find attesterResult")
+        }
```
sosnovsky, line 54: Do we need to have both - `wkdResult` and `attesterResult` or only one of them will be enough?

tomholub: Since we put both into task group, and waited for whole task group to finish, we expect both to come out.

Of course, maybe only one of them actually found public key, or none, or both - that's a separate concern.

sosnovsky: Ah, I see - it won't be `nil` even when keys not found, only `keys` property will be empty and this code will work well.
Previously I thought it'll fail if no keys were found, so all good here 👍

tomholub: I don't like this code pattern much. I've used it twice in this PR. I'm doing tasks in parallel that are not 100% interchangeable, so I need to make sure they all come out and I need to distinguish them (wkd keys have higher importance than attester keys). Maybe there's another code pattern to use? Without having to guard / filter / throw for each task group result.

sosnovsky (approved):
Nice cleanup 👍

sosnovsky (approved):

Issue #997, opened 2021-11-11
use @MainActor instead of DispatchQueue.main.async


tomholub:
Also for `@MainActor` I left comment only in 1 file, but we can create separate issue for it, as there are other places where it can help us to avoid running some functions on wrong thread. For example we can wrap `showToast`, `handleError` with `@MainActor` so we won't need to check if it's on main thread or not.

_Originally posted by @sosnovsky in https://github.com/FlowCrypt/flowcrypt-ios/issues/985#issuecomment-966323269_

tomholub:
Here is related issue: https://github.com/FlowCrypt/flowcrypt-ios/issues/876 - not sure what that issue means for this one / vice versa.

PR #1022, opened 2021-11-16
issue 1019 show alert on bad backup pp

### LINKED ISSUE #1019: entering wrong pass phrase when recovering from inbox backup should show modal
1) log into a regualar gmail account (non-enterprise) that has backups in inbox
2) give it wrong pass phrase

expected: modal

actual: view disappears, enters main flow

tomholub:
This PR ensures that bad pass phrases are handled well when recovering from inbox backups.

close #1019

**Tests**:
- ui tests will be added later - will file issue shortly

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Setup/SetupBackupsViewController.swift
+++ b/FlowCrypt/Controllers/Setup/SetupBackupsViewController.swift
@@ -185,7 +185,7 @@ extension SetupBackupsViewController {
         router.signOut()
     }
 
-    @MainActor private func moveToMainFlow() {
+    private func moveToMainFlow() {
         router.proceed()
     }
 }
--- a/FlowCrypt/Controllers/Setup/SetupBackupsViewController.swift
+++ b/FlowCrypt/Controllers/Setup/SetupBackupsViewController.swift
@@ -185,7 +185,7 @@ extension SetupBackupsViewController {
         router.signOut()
     }
 
-    private func moveToMainFlow() {
+    @MainActor private func moveToMainFlow() {
         router.proceed()
     }
 }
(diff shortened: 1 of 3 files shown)


tomholub:
UI test should be written for this:
1) log into an account that is not on enterprise domain, that has backups in inbox
2) enter a wrong pass phrase during setup, click continue
3) you should be seeing a modal telling you the pass phrase is wrong. Click ok
4) you should stay on the same screen
5) enter correct pass phrase, click continue
6) you should be seeing the inbox
7) open one encrypted email, make sure it opens successfully

tomholub:
@sosnovsky could you please help - why is it complaining about ui on non-main thread when I added MainActor on all the methods? Reproduce with the steps above.

image

sosnovsky:
When trying to find fix for this issue I came across 2 useful flags for Swift concurrency - `-Xfrontend -warn-concurrency -Xfrontend -enable-actor-data-race-checks` (https://twitter.com/olebegemann/status/1421144304127463427)
These flags add additional code checks, which are not enabled by default, and they found 55 errors in our current code.

tomholub:
very useful

sosnovsky:
@tomholub I pushed commit with fix for `non-main thread` issue - it seems in some cases code which isn't inside `Task` runs on non-main thread even with `@MainActor`.

tomholub:
> @tomholub I pushed commit with fix for `non-main thread` issue - it seems in some cases code which isn't inside `Task` runs on non-main thread even with `@MainActor`.

Ah - that means, MainActor has no effect if code was not run from a Task directly. That makes sense.

tomholub:
Looking at your code changes - looks good. That means, once we do #997 everywhere, the code should behave in a more predictable manner.

tomholub:
The PR is then ready for review 👍

sosnovsky (approved):

PR #1026, opened 2021-11-16
added accessibility identifiers for elements, updated appium tests

### LINKED ISSUE #944: use accessibility identifiers for appium tests?
Could you give these elements accessibility identifiers in the source code? Then here you can refer to them by the accessibility identifiers. This way, even the smallest change on that screen (adding an element that wasn't there before) would break these tests.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/943#discussion_r744549315_

fcvakintos:
This PR contains ned Accessibility Ids for elements which using in autotests, updats for appium tests

close #944 

**Tests** 
- Tests added or updated

### To be filled by reviewers


tomholub (approved):
Very good! Whenever possible, please add accessibility identifiers to the code when writing tests. It will be much more robust to layout changes. This is excellent.

Issue #863, opened 2021-10-30
current user should be dependency-injected


tomholub:
todo after #657 is done

instead of services and providers getting their own "currentAccountEmail", we should pass it in arguments to the methods. That way we achieve two things:
 - the methods no longer have to throw just in case they miss current email (since it will be provided)
 - we can remove a group of "middlemen classes" that only add account email to methods that need it - unneeded complexity

Additionally, ViewControllers should also receive `accountEmail` in their parameters from the parent view.

Somewhere in root view or router or similar there should be one single place that decides current user email and passes it down.

tomholub:
also this pattern should get similar treatment - pass email in function parameters

![image](https://user-images.githubusercontent.com/6306961/143572066-041afef2-b53e-4b06-bc74-9bcbf905bb68.png)

PR #1030, opened 2021-11-17
Separate Realm models and app logic models

### LINKED ISSUE #877: separate Realm models and app logic models
Since we use `async/await` now, it is hard to predict which thread will use `Realm` object and this leads to crashes

### LINKED ISSUE #1032: [User Feedback] Passphrase prompt occurs every time a user attempt to send a message during an active passphrase session
Description from original feedback:
>I have my password set to temporary.  After sending a message and then when I attempt to send a second email during the same session, it again asked for my password even though I had not logged off.  The app should retain my temporary password until I log off.

I can confirm this with FlowCrypt iOS 0.4.0 (via TestFlight)

Steps to reproduce:
1. Log in to your account with the FlowCrypt iOS app
2. Enter your passphrase and **choose the remember passphrase temporarily option**
3. Write a message and send it for the first time (a pop-up prompt will appear asking for a passphrase) - enter your passphrase
4. For the second time around, write a message once again and try to send it (another pop-up prompt asking again for a passphrase)

The user doesn't need to close the app or even put the phone to sleep. I believe the passphrase session can take up to 4 hours per session.

Reference: https://mail.google.com/mail/u/2/#inbox/FMfcgzGlksJFsgDbFTfpQkRLZMBRdvzl

ivan-ushakov:
close #877
close #1032

### To be filled by reviewers


FlowCrypt/Functionality/Services/Local Pub Key Services/Models/PubKey.swift
```diff
@@ -9,6 +9,7 @@
 import Foundation
 
 struct PubKey {
+    let primaryFingerprint: String
```
ivan-ushakov, line 12: Added this property to follow `PubKeyRealmObject`

FlowCrypt/Models/Realm Models/ClientConfigurationRealmObject.swift
```diff
@@ -17,7 +17,7 @@ final class ClientConfigurationRealmObject: Object {
     @Persisted var disallowAttesterSearchForDomains: Data?
     @Persisted var enforceKeygenAlgo: String?
     @Persisted var enforceKeygenExpireMonths: Int = -1
-    @Persisted var user: UserRealmObject!
+    @Persisted var user: UserRealmObject?
```
ivan-ushakov, line 20: @sosnovsky 
I decided to avoid having force unwrap in Realm objects. Maybe I'm wrong here so please take a look

FlowCrypt/Models/Realm Models/FolderRealmObject.swift
```diff
@@ -10,23 +10,21 @@ import Foundation
 import RealmSwift
 
 final class FolderRealmObject: Object {
-    @Persisted(primaryKey: true) var path: String = ""
```
ivan-ushakov, line 13: @sosnovsky
Same for the default values for Realm properties. Why we need them?

ivan-ushakov:
@tomholub 
I think this PR requires attention and it is not possible to have it ready today, so please feel free to make new release

FlowCrypt/Models/Realm Models/FolderRealmObject.swift
tomholub: Absolutely no need for default values for Realm properties, except when we specifically want them. 

As it was before, it was wrong.

tomholub:
> @tomholub I think this PR requires attention and it is not possible to have it ready today, so please feel free to make new release

Understood. I'll make a release without it. That way, this PR will land as one of the first PRs in the next release cycle, and we'll have time to fix bugs that come up before releasing it.

tomholub:
I'll let Roma review. Thanks!

FlowCrypt/Models/Realm Models/ClientConfigurationRealmObject.swift
tomholub: I know force unwrapping is not ideal. It's just that, client configuration always belongs to some user. So I thought it was weird that the user would be optional.

Is there a way to indicate that the user object is required without force unwrapping?

tomholub: Especially, I don't want to have some bugs in the future that would cause realm objects to be saved without reference to the user they belong to - that would be a mess.

sosnovsky: I think we can have just `@Persisted var user: UserRealmObject` without `?` or `!`, as it means that `user` property is not optional and should be set during `ClientConfigurationRealmObject` initialization.

FlowCrypt/Models/Realm Models/FolderRealmObject.swift
sosnovsky: Agree, no need for default values here

FlowCrypt/Models/Realm Models/ClientConfigurationRealmObject.swift
sosnovsky: Oh, I missed that Realm needs object properties to be optional. But currently all `ClientConfigurationRealmObject` initializers have required `user` property, so there won't be possibility to create new object without `user` property.

We can also update `var activeUser: UserRealmObject?` to look like
```swift
var activeUser: UserRealmObject {
        guard let user = user else { fatalError() }
        return user
}
```
Then it won't be possible to use configuration without assigned `user`

sosnovsky (commented):
Code looks much cleaner after models separation 👍

Just found a bug - app crashes when opening Contacts screen, maybe it affects UI tests as they're not passing now
Screenshot 2021-11-18 at 09 48 16

tomholub:
> We can also update var activeUser: UserRealmObject? to look like
> 
> ```
> var activeUser: UserRealmObject {
>        guard let user = user else { fatalError() }
>        return user
> }
>
> Then it won't be possible to use configuration without assigned user

That's basically a reimplementation of `!`. Unless we can implement it with a throw, but I don't think a dynamic var can throw? (intuitively, didn't try). Uless we write getters for everything but that also sounds excessive. Maybe the `!` was the best.

sosnovsky:
@tomholub ability to throw from `var` variable was added to the Swift recently:
```swift
var activeUser: UserRealmObject {
    get throws {
        guard let user = user else { throw AppErr... }
        return user
    }
  }
```

tomholub:
> @tomholub ability to throw from `var` variable was added to the Swift recently:
> 
> ```swift
> var activeUser: UserRealmObject {
>     get throws {
>         guard let user = user else { throw AppErr... }
>         return user
>     }
>   }
> ```

This looks good. Probably for another issue/PR?

The advantage is that an exception can be captured and handled (in the general error handling somewhere in VC, wherever user action was initiated) while a forced unwrap cannot.

sosnovsky:
Yes, error catching can be implemented in another PR, here we just need to fix `Contacts` screen crash.

FlowCrypt/Controllers/SetupImap/SetupImapViewController.swift
```diff
@@ -529,8 +532,8 @@ extension SetupImapViewController {
         globalRouter.signIn(with: .other(.session(user)))
     }
 
-    private func checkCurrentUser() -> Result<UserRealmObject, UserError> {
-        guard user != UserRealmObject.empty, user.email != UserRealmObject.empty.email else {
+    private func checkCurrentUser() -> Result<User, UserError> {
+        guard user != User.empty, user.email != User.empty.email else {
             return .failure(.empty)
         }
```
tomholub, line 538: In the future all similar `Result` usages should get rewritten to throw instead, if the goal is to communicate success / failure. 

Besides, current user should always come through method parameters, as per #863 . 

Filing an issue for the `Result` refactor.

ivan-ushakov:
> Code looks much cleaner after models separation 👍
> 
> Just found a bug - app crashes when opening Contacts screen, maybe it affects UI tests as they're not passing now Screenshot 2021-11-18 at 09 48 16

Yes, this one was the reason:
```
let objects = localContactsCache.realm.objects(RecipientRealmObject.self)
    .map(Recipient.init)
```
This produces lazy map. I fixed it.

ivan-ushakov:
@tomholub 
UI test failing because we show alert if user does not have public key:
```
[0-3] RUNNING in /Users/ivan/Library/Developer/Xcode/DerivedData/FlowCrypt-dkmvlhcprnebshbeurvkormazmnk/Build/Products/Debug-iphonesimulator/FlowCrypt.app - /tests/specs/composeEmail/SentEmailToRecipientWithoutPublicKey.spec.ts
[0-3] addedRecipientEl value:   no.publickey@flowcrypt.com  
[0-3] Screenshot of failed test was saved to /Users/ivan/Documents/Projects/FlowCrypt/flowcrypt-ios/appium/tmp
[0-3] Error in "COMPOSE EMAIL:  user is able to select recipient from contact list using contact name"
Error: Can't call setValue on element with selector "-ios class chain:**/XCUIElementTypeSecureTextField" because element wasn't found
```

sosnovsky (dismissed):
Works well now 👍

tomholub:
>  UI test failing because we show alert if user does not have public key:

The app behavior of showing alert is correct. Why master isn't affected by the failing test? Are you up to master as well?

ivan-ushakov:
> > UI test failing because we show alert if user does not have public key:
> 
> The app behavior of showing alert is correct. Why master isn't affected by the failing test? Are you up to master as well?

Code in master is [the same](https://github.com/FlowCrypt/flowcrypt-ios/blob/master/appium/tests/specs/composeEmail/SentEmailToRecipientWithoutPublicKey.spec.ts)

We also don't have any green PR's at this time.

tomholub:
PR I just made from master passes test: https://github.com/FlowCrypt/flowcrypt-ios/pull/1040

You don't need to involve me for this kind of thing. You can run a similar test on master - if there is an issue, file it and investigate it, and if there isn't, investigate your own code or how tests should be updated to work with updated code.

ivan-ushakov:
@tomholub
Could you tell me why we filter users with keys here?
```
func cleanupSessions() {
    logger.logInfo("Clean up sessions")

    encryptedStorage.getAllUsers()
        .filter {
            !encryptedStorage.doesAnyKeyExist(for: $0.email)
        }
        .map {
            logger.logInfo("User session to clean up \($0.email)")
            return $0.email
        }
        .forEach(logOut)

    let users = encryptedStorage.getAllUsers()

    if !users.contains(where: { $0.isActive }), let user = users.first(where: { encryptedStorage.doesAnyKeyExist(for: $0.email ) }) {
        switchActiveSession(for: user)
    }
}
```

We log out only users without keys?

tomholub:
The example code above is messy. There shouldn't be a general `cleanupSessions` making magic decisions like that. Please refactor this freely.

Makes no sense as implemented.

ivan-ushakov:
> The example code above is messy. There shouldn't be a general `cleanupSessions` making magic decisions like that. Please refactor this freely, with one account email as an input if needed.
> 
> Makes no sense as implemented.

What we need to do in `cleanupSessions`? Just log out current user or something else?

tomholub:
Don't know - depends what context is it called from

tomholub:
Ok I think I know what the intention originally was. If you log in with google, and never actually proceed with setup of the account keys, and then an hour later you open the app again, by now your ID token from Google will expire. You try to continue setup but it will just fail.

Therefore it's trying to clear up sessions that have no keys. I guess that's not too bad, but it would need to be renamed to `deleteSessionsOfAccountsThatDontHaveKeys` and please also refactor it to use `for x in y`, I can't stand to look at these map/forEach where a simple `for` loop is sufficient.

tomholub:
Also please add comment to it to explain what is it for / why per my notes above.

ivan-ushakov:
If we log out all users without keys that means we don't log out our current user because we set pass phrase for the key. When we enter compose screen we don't need to enter pass phrase because we did it during key setup and now we have it in memory storage. So test is wrong.

```
await NewMessageScreen.clickSentButton();

await EmailScreen.enterPassPhrase(passPhrase);
await EmailScreen.clickOkButton();
```

tomholub:
> If we log out all users without keys that means we don't log out our current user because we set pass phrase for the key.

in this method we don't log out users who already have keys set up. If our current user has keys already set up, then they won't be logged out.

> When we enter compose screen we don't need to enter pass phrase because we did it during key setup and now we have it in memory storage. So test is wrong.
> await NewMessageScreen.clickSentButton();
> await EmailScreen.enterPassPhrase(passPhrase);
> await EmailScreen.clickOkButton();
> May I fix it?

That means you have fixed #1032 ? In that case yes, you can fix the test.

appium/tests/specs/composeEmail/SendEncryptedEmailAfterWrongPassPhrase.spec.ts
```diff
@@ -23,10 +23,15 @@ describe('COMPOSE EMAIL: ', () => {
     const wrongPassPhraseError = CommonData.errors.wrongPassPhrase;
     const wrongPassPhrase = "wrong";
     const senderEmail = CommonData.account.email;
+    const bundleId = CommonData.bundleId.id;
 
     await SplashScreen.login();
     await SetupKeyScreen.setPassPhrase();
 
+    //Restart app to reset pass phrase memory cache
```
ivan-ushakov, line 31: @tomholub
This does not work and I have no idea how to fix it. It should restart app because when we did `SetupKeyScreen.setPassPhrase()` we need to reset in memory pass phrase cache.

FlowCrypt/Models/Common/User.swift
```diff
@@ -0,0 +1,72 @@
...
+        self.name = userObject.name
+        self.email = userObject.email
+        self.isActive = userObject.isActive
+        self.imap = userObject.imap.flatMap(Session.init)
+        self.smtp = userObject.smtp.flatMap(Session.init)
+    }
+}
+
+extension User {
+    static var empty: User {
```
tomholub, line 63: I've never liked this `User.empty` usage. Why not nil? Let's discuss in another issue.

FlowCrypt/Models/Realm Models/ClientConfigurationRealmObject.swift
tomholub: Filed issue #1035

appium/tests/specs/composeEmail/SentEmailToRecipientWithoutPublicKey.spec.ts
```diff
@@ -26,9 +24,6 @@ describe('COMPOSE EMAIL: ', () => {
     await NewMessageScreen.checkFilledComposeEmailInfo(noPublicKeyRecipient, emailSubject, emailText);
     await NewMessageScreen.clickSentButton();
```
tomholub, line 25: Should be called `clickSendButton` - filing another issue

appium/tests/specs/composeEmail/SendEncryptedEmailAfterWrongPassPhrase.spec.ts
tomholub: I'll take a look

tomholub: I'll take a look

tomholub: Did you run this locally while watching the simulator? Here's what happens:
 - app is started, authenticated, then pass phrase entered for setup
 - app is restarted
 - app is showing splash screen again - not set up

One could deduce that very likely, we are restarting the app too soon, before it had a chance to write setup result to storage. That's why we are seeing a splash screen again.

Don't you think? I'll try to add a wait for loaded inbox, see if that fixes it.

Next time you run into similar issue, please share a video of the failing test - you will get quicker help because sometime the problem may be obvious without having to rebuild and run it.

tomholub: That makes me think - it would be good if the CI produced a video of failed tests as well. I'll file an issue.

ivan-ushakov: Maybe I'm wrong but on my side it was like reinstall not just launching the same instance

tomholub: Hm. Try to run it after the changes I just pushed

tomholub:
btw was fixing of #1032 accidental or intentional? Did you reproduce the issue before the fix, and made sure (manually) it's fixed after the fix? To ensure we can indeed close it. I was just guessing from symptoms that you described, didn't check.

ivan-ushakov:
> btw was fixing of #1032 accidental or intentional? Did you reproduce the issue before the fix, and made sure (manually) it's fixed after the fix? To ensure we can indeed close it. I was just guessing from symptoms that you described, didn't check.

Yes, I forgot to explain, sorry. 

1. We have following code for getting signing key:
```
func getSigningKey() async throws -> PrvKeyInfo? {
    guard let email = currentUserEmail() else {
        logger.logError("no current user email")
        throw AppErr.noCurrentUser
    }
    // get keys associated with this account, freeze them to pass across threads
    let keysInfo = storage.keysInfo().filter { $0.account == email }.map { object -> KeyInfoRealmObject in
        guard object.realm != nil else { return object }
        return object.detached()
    }
    guard let foundKey = try await findKeyByUserEmail(keysInfo: keysInfo, email: email) else {
        return nil
    }

    let storedPassPhrases = passPhraseService.getPassPhrases()
    let passphrase = storedPassPhrases
        .filter { $0.value.isNotEmpty }
        .first(where: { $0.primaryFingerprintOfAssociatedKey == foundKey.primaryFingerprint })?
        .value

    return PrvKeyInfo(keyInfo: foundKey, passphrase: passphrase)
}
```

2. For some reason  `foundKey.primaryFingerprint` is always "" so we can not link pass phrase from memory storage with key.

I guess that the reason of 2 is default value we used in `KeyInfoRealmObject`:
```
@Persisted(primaryKey: true) var primaryFingerprint = ""
```

tomholub:
> I guess that the reason of 2 is default value we used in KeyInfoRealmObject:
> @Persisted(primaryKey: true) var primaryFingerprint = ""

I'd murder people for these default values :-) thank you for fixing that.

tomholub (approved):

appium/tests/specs/composeEmail/SendEncryptedEmailAfterWrongPassPhrase.spec.ts
ivan-ushakov: @tomholub
You was right about delay, it works. I will pay more attention on such things in future.

tomholub: Didn't fix it with a delay - fixed with awaiting the next expected step before restarting (waiting for rendered inbox screen)

PR #1031, opened 2021-11-17
issue 902 preview attachments in web view

### LINKED ISSUE #902: preview (or directly open) attachments instead of downloading
would it be difficult / possible to view attachments, if some app allows to open it, instead of downloading it first?

ekievsky:
This PR adds attachment webview to preview attachemnts

close #902

- Does not need tests

### To be filled by reviewers


ekievsky:
@tomholub please assign next tasks on me, thank you!

tomholub:
Assingned more - thank you.

FlowCrypt/Common UI/AttachmentManager.swift
```diff
@@ -44,8 +44,8 @@ final class AttachmentManager: NSObject {
 }
 
 extension AttachmentManager: AttachmentManagerType {
-    func open(_ attachment: MessageAttachment) {
-        filesManager.saveToFilesApp(file: attachment, from: self)
+    func open(_ file: FileType) {
```
tomholub, line 47: Method names should be renamed if they no longer make sense. If it's downloading the file, and we now also have an option to view the file, then this method should be renamed to `saveToDevice` or similar.

Also, should be refactored as an async method so that the error can be thrown up and handled by wherever the action initiated, with a `Task` and a do/try/catch. This is the preferred pattern going forward.

FlowCrypt/Functionality/FilesManager/FilesManager.swift
```diff
@@ -25,6 +25,9 @@ protocol FilesManagerType {
 
     @discardableResult
     func selectFromFilesApp(from viewController: FilesManagerPresenter & UIDocumentPickerDelegate) -> Future<Void, Error>
+
+    func saveLocally(file: FileType) -> URL?
+    func remove(file: FileType)
 }
 
 class FilesManager: FilesManagerType {
```
tomholub, line 33: What is the difference between `save`, `saveFilesToApp`, `saveLocally`? Methods should be renamed to make the distinction clearer, seems all the same to me.

Also should be refactored as async methods.

FlowCrypt/Functionality/FilesManager/FilesManager.swift
```diff
@@ -89,4 +92,27 @@ class FilesManager: FilesManagerType {
...
+            create: true
+        )
+        guard let localPath = documentDirectory?.path.appending("/\(file.name)")
+        else { return nil }
+        let url = URL(fileURLWithPath: localPath)
+        try? file.data.write(to: url)
+        return url
+    }
+
+    func remove(file: FileType) {
```
tomholub, line 110: Use consistent naming to make it clearer. If above is a `saveLocally` method that is clearly related to this, then at a minimum this should be called `removeLocal` but it's still not so clear. What does local mean? In the app's own storage not accessible by other apps?

FlowCrypt/Functionality/Services/GeneralConstants.swift
```diff
@@ -13,7 +13,7 @@ enum GeneralConstants {
 
     enum Global {
         static let generalError = -1
-        static let attachmentSizeLimit: Int = 10_000_000
+        static let attachmentSizeLimit: Int = 10_485_760
```
tomholub, line 16: Either would be fine. One is 10MiB and other is 10MB, neither is more right than the other.

tomholub (commented):
General refactoring notes below.

UI tests will also need to be updated in this PR.

Can't build it because of this error:

```
Undefined symbols for architecture arm64:
  "FlowCryptUI.NavigationBarItemsView.Input.init(image: __C.UIImage?, action: (target: Any?, selector: ObjectiveC.Selector)?, accessibilityLabel: Swift.String?, onTap: (() -> ())?) -> FlowCryptUI.NavigationBarItemsView.Input", referenced from:
      (extension in FlowCrypt):FlowCrypt.MessageActionsHandler< where A: __C.UIViewController>.(setupNavigationBarItems in _E9403E9BE08F6EDB817C8A074140A1F7)(with: Swift.String?) -> () in MessageActionsHandler.o
      FlowCrypt.ContactKeyDetailViewController.(setupNavigationBarItems in _B6EC48D05BC1AF174CD6F966F23191DB)() -> () in ContactKeyDetailViewController.o
      FlowCrypt.ContactDetailViewController.(setupNavigationBarItems in _19EF4E13ACB52787E1C776470DDC7D1C)() -> () in ContactDetailViewController.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

FlowCrypt/Functionality/Services/GeneralConstants.swift
ekievsky: Reverted so it more clear

ekievsky:
@tomholub I've fixed PR comments and couldn't run and fix the tests, who may I ask in order to help me?

tomholub:
> @tomholub I've fixed PR comments and couldn't run and fix the tests, who may I ask in order to help me?

Ask right here. What issues are you having?

FlowCrypt/Functionality/FilesManager/FilesManager.swift
```diff
@@ -93,26 +90,15 @@ class FilesManager: FilesManagerType {
...
-    }
-
-    func remove(file: FileType) {
-        let fileManager = FileManager.default
-        let documentDirectory = try? fileManager.url(for: .documentDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
-        guard let localPath = documentDirectory?.path.appending("/\(file.name)")
-        else { return }
-        let url = URL(fileURLWithPath: localPath)
-        try? fileManager.removeItem(at: url)
+    func remove(file: FileType) -> Future<Void, Error> {
```
tomholub, line 93: This is still using a Future which we don't do anymore. Use an async method.

tomholub (changes requested):

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
```diff
@@ -0,0 +1,137 @@
...
+        addWebView()
+
+        filesManager.save(file: file)
+            .sink(
+                receiveCompletion: {_ in},
+                receiveValue: { [weak self] url in
+                    self?.load(with: url)
+                }
+            )
+            .store(in: &cancellable)
```
tomholub, line 67: Here use a `Task` and call the async method with try/catch. On catch, handle the error.

tomholub (changes requested):

tomholub (commented):
functionality-wise:

1) missing a loader when rendering image. Eg try a 5mb image - it takes a while (a few seconds) before it shows, and meanwhile it just shows white/blank screen with green top.

2) the view that has the webview in it should show the attachment name

3) I got a crash:

```
=================================================================
Main Thread Checker: UI API called on a background thread: -[WKWebView loadRequest:]
PID: 4635, TID: 1587561, Thread name: (none), Queue name: com.apple.root.background-qos, QoS: 9
Backtrace:
4   FlowCrypt                           0x0000000102981908 $s9FlowCrypt24AttachmentViewControllerC4load33_FDBE842BC02C43F80E5BD063943C767ALL4withy10Foundation3URLVSg_tF + 532
5   FlowCrypt                           0x00000001029816cc $s9FlowCrypt24AttachmentViewControllerC11viewDidLoadyyFy10Foundation3URLVcfU0_ + 316
6   Combine                             0x00000001991f1c04 $s7Combine11SubscribersO4SinkC7receiveyAC6DemandVxF + 96
7   Combine                             0x00000001991f153c 30730549-3E9F-3906-9A9C-60341B800C46 + 46396
8   Combine                             0x0000000199212118 30730549-3E9F-3906-9A9C-60341B800C46 + 180504
9   Combine                             0x00000001991ee248 30730549-3E9F-3906-9A9C-60341B800C46 + 33352
10  Combine                             0x00000001991edb9c 30730549-3E9F-3906-9A9C-60341B800C46 + 31644
11  Combine                             0x0000000199210268 30730549-3E9F-3906-9A9C-60341B800C46 + 172648
12  Combine                             0x000000019922413c 30730549-3E9F-3906-9A9C-60341B800C46 + 254268
13  Combine                             0x00000001991ed8c0 30730549-3E9F-3906-9A9C-60341B800C46 + 30912
14  FlowCrypt                           0x00000001028e006c $s9FlowCrypt12FilesManagerC4save4file7Combine6FutureCy10Foundation3URLVs5Error_pGAA8FileType_p_tFyys6ResultOyAKsAL_pGccfU_yycfU_ + 552
15  FlowCrypt                           0x00000001027a299c $sIeg_IeyB_TR + 48
16  libdispatch.dylib                   0x0000000109cb06d4 _dispatch_call_block_and_release + 32
17  libdispatch.dylib                   0x0000000109cb23b4 _dispatch_client_callout + 20
18  libdispatch.dylib                   0x0000000109cc6738 _dispatch_root_queue_drain + 976
19  libdispatch.dylib                   0x0000000109cc6f0c _dispatch_worker_thread2 + 196
20  libsystem_pthread.dylib             0x00000001f235a0b8 _pthread_wqthread + 228
21  libsystem_pthread.dylib             0x00000001f2359e94 start_wqthread + 8
```

tomholub:
Additionally:
4) .txt files don't work
5) .pdf files don't work
7) I assume other files like xls, xlsx, doc, docx, ppt, etc don't work also (but webview supports them - add all files that webview supports)
8) if a file is not supported, it just shows blank screen. The app should know which files it can attempt to open in webview and which ones don't work, and don't attempt the ones it doesn't support.

ekievsky:
1. done
2. done
3. no crash with async now
4. 5. 6. could you send the example of file and video, because it's weird I've tested with pptx, xls, docx, txt, png, jpeg and all worked
7. if any error comes up error is handled

ekievsky:
About the tests, I was trying to build them locally and now I am having following issue:
photo_2021-11-25_00-20-45

also I am not sure how to fix current issue with UI tests

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
```diff
@@ -56,22 +54,27 @@ final class AttachmentViewController: UIViewController {
...
-            .sink(receiveCompletion: {_ in}, receiveValue: {_ in})
-            .store(in: &cancellable)
+        remove(file: file)
+    }
+
+    private func remove(file: FileType) {
+        Task {
+            try await filesManager.remove(file: file)
+        }
+    }
```
tomholub, line 71: There should be a try/catch and a handle with modal showing the error (don't ignore errors)

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
```diff
@@ -56,22 +54,27 @@ final class AttachmentViewController: UIViewController {
...
+
+    private func saveAndStartDownload(file: FileType) {
+        Task { [weak self] in
+            let url = try await filesManager.save(file: file)
+            self?.load(with: url)
+        }
```
tomholub, line 77: Same here - handle errors

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
```diff
@@ -103,13 +106,19 @@ extension AttachmentViewController: WKNavigationDelegate {
...
 }
 
 private extension AttachmentViewController {
 
     private func load(with link: URL?) {
         guard let wrappedLink = link else {
-            errorLabel.isHidden = false
+            DispatchQueue.main.async { [weak self] in
+                self?.errorLabel.isHidden = false
+            }
```
tomholub: Use MainActor instead

tomholub:
> About the tests, I was trying to build them locally and now I am having following issue: photo_2021-11-25_00-20-45
> 
> also I am not sure how to fix current issue with UI tests

Have you followed all of the steps here? https://github.com/FlowCrypt/flowcrypt-ios/blob/master/appium/README.md

ekievsky:
yes, I did

tomholub:
What is the full folder path of your flowcrypt-ios folder?

tomholub:
> 5, 6 could you send the example of file and video, because it's weird I've tested with pptx, xls, docx, txt, png, jpeg and all worked

Try send yourself an encrypted message with small txt, pdf, xls, doc files in it. Can they be previewed?

ekievsky:
> > 5, 6 could you send the example of file and video, because it's weird I've tested with pptx, xls, docx, txt, png, jpeg and all worked
> 
> Try send yourself an encrypted message with small txt, pdf, xls, doc files in it. Can they be previewed?

That's what I did to test it. I opened the files from Compose view and received email view. It works for both.
Maybe you are sending blank files?

ekievsky:
Seems that UITest are failing for all the PRs and master.

tomholub:
> Seems that UITest are failing for all the PRs and master.

master looks fine: 

image

You can make a blank draft PR from master to check if the tests pass or not. If they don't, then raise an issue about it.

tomholub:
On `flowcrypt.compatibility` (I'm sending you credentials by email) account there is an email titled `file preview - open txt file, image, pdf, unknown`. In it, there are 4 files to test.

image

image

image

image

Let me do screenshots on the phone, then I'll post them here and send the credentials.

ekievsky:
Seems that UITest are failing for all the PRs and master.

> What is the full folder path of your flowcrypt-ios folder?

/Users/personal/git/flowcrypt-ios

tomholub:
Now it works :-) (testing latest version) except the "unknown" file has a loader that never goes away. Maybe you can reproduce this by removing ".png" from an image and sending it encrypted, then opening?

image

tomholub:
Also please adjust the header design:
 - name of file should be left-aligned, next to back button
 - download icon should be smaller, and visually together with "Download"

tomholub:
Actually instead of `Download` please use `Save`, since it's shorter - and there's limited space.

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
```diff
@@ -59,21 +60,29 @@ final class AttachmentViewController: UIViewController {
...
-
     private func remove(file: FileType) {
-        Task {
-            try await filesManager.remove(file: file)
+        Task { [weak self] in
+            do {
+                try await filesManager.remove(file: file)
+                self?.navigationController?.popViewController(animated: true)
+            } catch {
+                showAlert(message: "could_not_remove_opened_attachment".localized) {
```
tomholub, line 69: Include the `error` message in the printed string to the user. Else when we inevitably get a screenshot from user or customer when this happened, we won't be able to debug it.

Just adding `+ "
(\(error.localizedDescription))"` or something similar should already help.

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
```diff
@@ -59,21 +60,29 @@ final class AttachmentViewController: UIViewController {
...
 
     private func saveAndStartDownload(file: FileType) {
         Task { [weak self] in
-            let url = try await filesManager.save(file: file)
-            self?.load(with: url)
+            do {
+                let url = try await filesManager.save(file: file)
+                self?.load(with: url)
+            } catch {
+                showAlert(message: "could_not_open_an_attachment".localized) {
```
tomholub, line 82: Same as above, don't swallow error details: surface them for easier debugging.

tomholub (commented):
error handling feedback - still evaluating at the rest

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
ekievsky: I am not really sure that user needs the actual error from iOS, but it easy to add, I will.

FlowCrypt/Common UI/AttachmentManager.swift
```diff
@@ -45,15 +45,21 @@ final class AttachmentManager: NSObject {
...
+        }
+    }
+
+    @MainActor func openDocumentsController(from url: URL) {
+        let documentController = UIDocumentPickerViewController(forExporting: [url])
+        documentController.delegate = self
+        present(documentController, animated: true, completion: nil)
     }
 
     private func handle(_ error: Error) {
```
tomholub, line 65: This method no longer has any value and can be removed. Wherever this is used, call `await controller?.showToast("\("message_attachment_saved_with_error".localized) \(error.localizedDescription)")` directly instead.

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
tomholub: The user doesn't need it.

But the user will complain, take a screen shot and send it to me. If it's missing the iOS error, what can I possibly do with such report?

ekievsky:
> name of file should be left-aligned, next to back button

It is not really natively possible because navigation view title is always on the middle

tomholub (commented):
Improving - thanks.

tomholub:
> > name of file should be left-aligned, next to back button
> 
> It is not really natively possible because navigation view title is always on the middle

That's a real pity. @sosnovsky have you encountered this?

ekievsky:
@tomholub flowcrypt.compatibility there is 2FA, can you help me to get in?

tomholub:
> @tomholub flowcrypt.compatibility there is 2FA, can you help me to get in? it says that you need to push number 7

use the backup code instead - I've shared one in the email.

FlowCrypt/Common UI/AttachmentManager.swift
```diff
@@ -45,24 +45,24 @@ final class AttachmentManager: NSObject {
...
-            await controller?.showToast(
-                "\("message_attachment_saved_with_error".localized) \(error.localizedDescription)"
-            )
+            do {
+                let url = try await filesManager.save(file: file)
+                openDocumentsController(from: url)
+            } catch {
+                await controller?.showToast(
+                    "\("message_attachment_saved_with_error".localized) \(error.localizedDescription)"
+                )
```
tomholub, line 56: await not needed. good to follow warnings

image

ekievsky:
> use the backup code instead - I've shared one in the email. 
it still asks for 2FA "To make sure that it is really you"

tomholub:
> > use the backup code instead - I've shared one in the email.
> > it still asks for 2FA "To make sure that it is really you"

Got it. I'll grab my device. Does it still want 7 or other number?

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
ekievsky: We could use Crashlytics or similar to that to get the reports for crashes and these kind of errors.

ekievsky:
@tomholub it is 45

tomholub:
approved

ekievsky:
I'm in, thanks!

ekievsky:
@tomholub which email do I open? I went through test/attachments folder and could not find email with `unknown` file

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
tomholub: Yes, but when a particular user writes you with a screenshot, you have to match them up with the reports that may or may not be in Crashlytics. When it is much simpler to just surface it to the user.

Hiding errors from users is a bad habit too common in the apple ecosystem. When apple website itself tells me "something went wrong" when it should have said "your session expired, you should log in again" (it does that all the time) I feel like throwing my apple laptop through a window (of Apple offices). It's a bad engineering practice.

FlowCrypt/Common UI/AttachmentManager.swift
ekievsky: done

ekievsky:
@tomholub fixed issue with infinite loader.

tomholub:
> @tomholub which email do I open? I went through test/attachments folder and could not find email with `unknown` file

The subject is written on top of https://github.com/FlowCrypt/flowcrypt-ios/pull/1031#issuecomment-979467917

tomholub (commented):
getting better

ekievsky:
> > @tomholub which email do I open? I went through test/attachments folder and could not find email with `unknown` file
> 
> The subject is written on top of [#1031 (comment)](https://github.com/FlowCrypt/flowcrypt-ios/pull/1031#issuecomment-979467917)

I watched up to delete that before you commented 🤣

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+                self?.load(with: url)
+            } catch let error {
+                showAlert(
+                    message: "could_not_open_an_attachment".localized + " (\(error.localizedDescription))"
+                ) {
+                    self?.navigationController?.popViewController(animated: true)
+                }
+            }
+        }
+    }
```
tomholub, line 91: If I understand this correctly, here is what happens:
 - I tap on the attachment to preview it
 - attachment gets automatically saved somewhere (unsure where, not familiar with iOS environment)
 - attachment gets previewed automatically
 - attachment gets removed (hopefully, unless there's an error, and then it's left hanging)

I'd like to avoid saving it completely, and go straight from memory to preview without intermediate steps. You should be able to provide it to the webview as a blob directly. Is that possible?

tomholub (changes requested):

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
ekievsky: The WebView can be loaded with URLRequest only, that need to have the path. 
Maybe @sosnovsky has ever dealt with opening files in WebView from memory?

ekievsky: What is the issue with temporary saving the file?

tomholub: Security - if you drop your phone and someone unlocks and roots it, they get your attachment content. They would not be able to get that if it was kept in memory.

Consistency - files are not cleared, as implemented, if errors happened after saving them but before deleting them.

Consistency 2 - files are not cleared if there is error while clearing them.

Consistency 3 - if app crashes in between opening file and removing it, files are left over.

These three consistency issues worsen the security problem - there is bound to be files left over after you've used the app a lot, and since you bothered to use encryption in the first place, they are likely very sensitive files.

Bettery - doing work you would not need to otherwise do consumes battery. This is minor compared to the security issue.

Complexity - if it was possible to directly view the file, it would surely be simpler to implement than saving and deleting files. (also minor compared to security)

tomholub: Here you can directly load from `Data` and a mime type - seems exactly what we need.

https://developer.apple.com/documentation/webkit/wkwebview/1415011-load

```swift
func load(
  _ data: Data, 
  mimeType MIMEType: String, 
  characterEncodingName: String, 
  baseURL: URL
) -> WKNavigation?
```

ekievsky: Just found it and trying to handle that.

ekievsky:
@tomholub Something went wrong with this PR, really. I am sorry about that.

sosnovsky:
> > > name of file should be left-aligned, next to back button
> > 
> > 
> > It is not really natively possible because navigation view title is always on the middle
> 
> That's a real pity. @sosnovsky have you encountered this?

Yes, it's a default iOS behavior - all screen titles are centered, and if title is too long then it's trimmed with `...`
Left-aligning can be achieved with custom navigation bar, but I think it'll be better to use native iOS design

tomholub (commented):
@ekievsky you can see for yourself that this is much, much better then when we started. 

@sosnovsky could you have a look? Also, I think I noticed you use some sort of standard iOS icon set instead of own icons (not 100% sure). If that's true, could that be applied here for the download icon?

sosnovsky:
Yes, for new buttons I'm using built-in SF Icons (https://developer.apple.com/sf-symbols/), they're were customisable and give consistency for different buttons.

> @sosnovsky could you have a look?
what are the main issues? as your discussion is quite long here, difficult to find what needs to be fixed

tomholub:
> > @sosnovsky could you have a look?
> 
> what are the main issues? as your discussion is quite long here, difficult to find what needs to be fixed

I'm happy with the PR now, just wanted your review on it since it's iOS stuff :-)

tomholub:
> Yes, for new buttons I'm using built-in SF Icons (https://developer.apple.com/sf-symbols/), they're were customisable and give consistency for different buttons.

@ekievsky could you use this standard set of icons instead of adding custom ones?

tomholub:
I merged changes from master blindly, hopefully correctly.

sosnovsky:
I tried to preview .xlsx file, but it showed me just raw file
Simulator Screen Shot - iPhone 12 - 2021-11-26 at 15 01 55

tomholub:
I haven't tested it after the latest changes. @ekievsky have you?

sosnovsky:
Other formats like `.txt`, `.csv`, `.png`, `.md` work well

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
```diff
@@ -0,0 +1,124 @@
...
+    private func load() {
+        webView.load(
+            file.data,
+            mimeType: file.name.mimeType,
+            characterEncodingName: "UTF-8",
+            // It is a URL that should be used to resolve relative URLs within the document.
+            // As we don't resolve any relative URLs we pass just NSURL()
+            baseURL: NSURL() as URL
+        )
+    }
```
tomholub, line 103: Likely caused by the `characterEncodingName` here - it's treating it as text regardless of actual content.

That's why only file types that are inherently textual continue to work.

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -520,19 +520,27 @@ extension ComposeViewController: ASTableDelegate, ASTableDataSource {
...
-        case 1:
-            let selectedEmail = emails[safe: indexPath.row]
-            handleEndEditingAction(with: selectedEmail)
-        case 2:
-            askForContactsPermission()
-        default:
-            break
+    func tableNode(_ tableNode: ASTableNode, didSelectRowAt indexPath: IndexPath) {
+        if case let .searchEmails(emails) = state,
+           indexPath.section == 1 {
```
sosnovsky, line 525: Why we need `indexPath.section == 1` condition here?
In this case line 531 `askForContactsPermission()` won't be executed

sosnovsky (commented):
Looks good, only a couple issues - with `.xlsx` preview and `askForContactsPermission` from review comment

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
sosnovsky: I checked `.png` and `.jpg` - work well

tomholub: @sosnovsky Can we still use the original method that opens a file from url, instead of giving it raw bytes, but instead of file url that points to a local filestem, we would give it a memory-based blob url? Is there such a concept on iOS?

sosnovsky:
Also iOS has built-in tool for previewing different files - https://developer.apple.com/documentation/quicklook/qlpreviewcontroller
The only limitation is that it can't be customised

1*LpU0KYcitEKF0SP0FOX1uQ

tomholub:
qlpreviewcontroller seems better for this then raw webview. As long as the "share" button allows to save to device.

tomholub:
Oh! And there must be ability to disable the share button (for customers with strict security requirements).

tomholub:
It looks like disabling the share button involves hacky non-standard fixes, not ideal. Let's first see if we can make it work with WebView.

sosnovsky:
> qlpreviewcontroller seems better for this then raw webview. As long as the "share" button allows to save to device.

Share button allows to send file to any supported app (including Files app) - it works the same as `Share` button in the Gmail app.

> It looks like disabling the share button involves hacky non-standard fixes, not ideal. Let's first see if we can make it work with WebView.

Yes, there is only some subclassing methods for hiding share button for now.

tomholub:
> > It looks like disabling the share button involves hacky non-standard fixes, not ideal. Let's first see if we can make it work with WebView.
> 
> Yes, there is only some subclassing methods for hiding share button for now.

Does it look usable / reliable? If we went that route?

FlowCrypt/Controllers/Attachment/AttachmentViewController.swift
tomholub: Answering my own question: I think yes https://stackoverflow.com/a/23647182

tomholub:
@ekievsky things left to do:

 - [ ] try to resolve the xlsx issue by using url blob https://stackoverflow.com/a/23647182 passed into the original method you used. So you convert data to base64 string, make a blob url from it, and pass that into the method you used originally that accepted url
 - [ ] address Roma's comment https://github.com/FlowCrypt/flowcrypt-ios/pull/1031#discussion_r757492918
 - [ ] merge changes from master again, and look into failing tests

sosnovsky:
> > > It looks like disabling the share button involves hacky non-standard fixes, not ideal. Let's first see if we can make it work with WebView.
> > 
> > 
> > Yes, there is only some subclassing methods for hiding share button for now.
> 
> Does it look usable / reliable? If we went that route? (edit - will try this with webview first, since it's more flexible. If we keep running into problems, we'll try the QuickLook.

This solution looks simple and reliable https://stackoverflow.com/a/69575481/3091318

tomholub:
Do you think this PR should be reworked to use QuickLook?

sosnovsky:
Yes, I think we should try to remake this PR to use `QLPreviewController` - it should be more stable than our custom webview solution and probably supports more file types.
@ekievsky do you agree or maybe you know some disadvantages of `QLPreviewController`?

ekievsky:
@tomholub The answer from stackoverflow you sent is from 2014 year, so it's been almost 8 years since that solution works. Also the answer is for `NSImage` not Data and I tested with data and this URL doesn't give any output in WebView.
@sosnovsky I am up to any better solution, but `QLPreviewController` doesn't work with `Data` it can open `URL` only, as I told before that stackoverflow suggestions is not working.
As we can see on developer.apple Web KIt(from iOS 8+) is much newer than QLPreviewController(from iOS 4+) so I would follow the rule to use latest possible API instead of using something old.
Let me know your thoughts.

tomholub:
Then give it a data url https://stackoverflow.com/questions/23644193/can-i-create-an-nsurl-that-refers-to-in-memory-nsdata/23647182#23647182

ekievsky:
That's what I did, this URL is not openable.

sosnovsky:
@ekievsky you can create file from attachment's `Data` (as we already has it downloaded) in `tmp` folder and give this URL to `QLPreviewController` (similar solution is here https://stackoverflow.com/a/62860331/3091318)

> As we can see on developer.apple Web KIt(from iOS 8+) is much newer than QLPreviewController(from iOS 4+) so I would follow the rule to use latest possible API instead of using something old.

`WKWebView` is newer, but it has different purpose - showing web content, while `QLPreviewController` is `A specialized view controller for previewing an item.`. So probably it won't have any problems with `xlsx` formatting and will give us more native UI for previewing different types, not just web page with embedded file.

Let's try to implement `QLPreviewController` so we'll be able to compare both solutions and choose the best one.

ekievsky:
Tom said that he wants to avoid saving file even for temporary usage

tomholub:
Yes - for security reasons, the file should never be saved on the device decrypted, even if temporarily, unless the user explicitly saved it. For preview only, it should be done in memory.

tomholub:
> That's what I did, this URL is not openable.

I've just pushed a commit that fixes previewing of xlsx with WebView using a data URL. 

With that, you should be able to open attachments using `QLPreviewController`.

tomholub:
@ekievsky any luck?

tomholub:
@ekievsky

ekievsky:
I tried QuickLook and it shows the raw base64 encoded string 
Simulator Screen Shot - iPhone 13 - 2021-11-30 at 22 15 31

tomholub:
@sosnovsky I propose we could clean up / merge this PR using WebView, and then you would have a look at it sometime later to potentially switch to quicklook. Does that sound ok?

tomholub:
@ekievsky either way please merge changes from master, and start looking at running / updating ui tests. We were testing file download in one of them, which will now be broken: it expects that file will download right after tapping it, but now there is one more screen in between. That test will need to be updated.

sosnovsky:
> @sosnovsky I propose we could clean up / merge this PR using WebView, and then you would have a look at it sometime later to potentially switch to quicklook. Does that sound ok?

Sorry, missed my mention here - sounds good, so we'll have a basic working solution for now and then will try to improve it later 👍

tomholub:
@ekievsky any news updating this to master? The longer we wait, the more it will diverge.

tomholub:
@ekievsky

tomholub:
Evgenii wrote won't be able to finish this at the moment due to lack of availability. I'll see if it will be easy enough to merge.

tomholub:
Not so easy - involves UI work that I'm not too familiar with. @ivan-ushakov once you are done with currently opened PRs, please have a look to bring this to master. Thanks!

ivan-ushakov:
Let's wait for #1189

tomholub:
There's also this comment from Roma that I think wasn't addressed

image

FlowCrypt/Controllers/Compose/ComposeViewController.swift
ivan-ushakov: I don't know why we have this condition. Let's revert

appium/tests/screenobjects/email.screen.ts
```diff
@@ -7,7 +7,8 @@ const SELECTORS = {
   ENTER_PASS_PHRASE_FIELD: '-ios class chain:**/XCUIElementTypeSecureTextField',
   OK_BUTTON: '~Ok',
   WRONG_PASS_PHRASE_MESSAGE: '-ios class chain:**/XCUIElementTypeStaticText[`label == "Wrong pass phrase, please try again"`]',
-  DOWNLOAD_ATTACHMENT_BUTTON: '~attachmentDownloadButton0',
+  ATTACHMENT_CELL: '~attachmentCell0',
+  ATTACHMENT_TITLE: 'attachmentTitleLabel0',
```
ivan-ushakov, line 11: For some reason this one does not work on my machine, but identifier is correct

appium/tests/screenobjects/email.screen.ts
```diff
@@ -133,19 +138,13 @@ class EmailScreen extends BaseScreen {
     await (await this.wrongPassPhraseMessage).waitForDisplayed();
   }
 
-  attachmentName = async (name: string) => {
-    const selector = `-ios class chain:**/XCUIElementTypeStaticText[\`label == "${name}"\`]`;
```
ivan-ushakov, line 137: @fcvakintos
Is this one the only way to get attachment title?

tomholub:
We could also merge this with incomplete tests, and then you can formulate a separate issue for @fcvakintos

sosnovsky (approved):
Looks good for me now, just created separate issue for possible preview improvements in the future - https://github.com/FlowCrypt/flowcrypt-ios/issues/1201

tomholub:
> We could also merge this with incomplete tests, and then you can formulate a separate issue for @fcvakintos

I meant: if unsure how to fix the tests, you could disable the problematic part of tests so that the CI passes, then I could merge it.

tomholub (dismissed):

ivan-ushakov:
Still investigating problems with UI tests

ivan-ushakov:
@tomholub
I disabled some checks with `FIXME` tag in UI tests to get green mark

tomholub (approved):
Thanks!

PR #1040, opened 2021-11-19
update ui test name

tomholub:
This PR updates ui test name and tries running ui tests to see if they work.

no issue ref


ivan-ushakov:
OK, you are right that `master` branch has green tests. That means we have a bug in current logic.

ivan-ushakov (approved):

PR #1042, opened 2021-11-19
added move to trash and delete steps for SendEncryptedEmail spec

### LINKED ISSUE #1029: add mark unread, move to trash, delete to a sending test
In another issue:

It would be good, since we already sent a new message we don't need to also test message actions on it and delete it in the same test. 

- before opening the message, verify it's marked as unread in inbox
- after opening the message, run `checkOpenedEmail` as before
- after that, go back to sent (can use the back icon) and verify it's marked as read
- open the message again and tap "mark as read". Back to sent list - verify it's unread (this checks that we cleverly re-rendered the message after the action without need to refresh)
- refresh the sent message by pulling down on the list, then verify it's still unread (this checks that it was updated as unread on the mail server)
- open the message again and tap "Trash" icon
- go back using back icon, and verify that the message is gone from the list without refreshing
- refresh the list, make sure it's still gone
- using Menu, go to Trash folder, find the message, open it, and tap Trash again (this completely deletes the message)
- using back button, go back to Trash, see that message is not there
- refresh Trash by pulling down, see that message is not there

Then finally, change the test name by adding `+ mark unread, move to trash, delete`

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1027#discussion_r751474066_

fcvakintos:
This PR contains changes for SendEncryptedEmail spec , added scrollUp and scrollDown methods to TouchHelper for refreshing sent and trash lists, added checks for move sent email to trash and delete it from trash list

close #1029 

**Tests** 
- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/appium/tests/screenobjects/sent.screen.ts
+++ b/appium/tests/screenobjects/sent.screen.ts
@@ -1,0 +1,43 @@
+import BaseScreen from './base.screen';
+import TouchHelper from "../helpers/TouchHelper";
+
+const SELECTORS = {
+  SENT_HEADER: '-ios class chain:**/XCUIElementTypeNavigationBar[`name == "SENT"`]',
+  SEARCH_ICON: '~search icn',
+  HELP_ICON: '~help icn'
+};
+
+
+class SentScreen extends BaseScreen {
+  constructor() {
+    super(SELECTORS.SEARCH_ICON);
+  }
+
+  get searchIcon() {
+    return $(SELECTORS.SEARCH_ICON);
+  }
+
+  get helpIcon() {
+    return $(SELECTORS.HELP_ICON);
+  }
+
+  get sentHeader() {
+    return $(SELECTORS.SENT_HEADER)
+  }
+
+  checkSentScreen = async () => {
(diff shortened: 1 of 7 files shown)


appium/tests/screenobjects/sent.screen.ts
```diff
@@ -0,0 +1,43 @@
...
+import TouchHelper from "../helpers/TouchHelper";
+
+const SELECTORS = {
+  SENT_HEADER: '-ios class chain:**/XCUIElementTypeNavigationBar[`name == "SENT"`]',
+  SEARCH_ICON: '~search icn',
+  HELP_ICON: '~help icn'
+};
+
+
+class SentScreen extends BaseScreen {
```
tomholub, line 11: Inbox, sent, trash, and other such screens are all the same - they are all just a folder screen. So it should probably really just be one screen. A `MailFolderScreen` or maybe just `InboxScreen` or if you want to extend from something and still have them all separate, then `BaseFolderScreen`. What do you think?

tomholub (approved):

PR #1055, opened 2021-11-20
Use @MainActor instead of DispatchQueue.main.async

### LINKED ISSUE #997: use @MainActor instead of DispatchQueue.main.async

ivan-ushakov:
close #997

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Common UI/AttachmentManager.swift
+++ b/FlowCrypt/Common UI/AttachmentManager.swift
@@ -9,6 +9,7 @@
 import UIKit
 import Combine
 
+@MainActor
 protocol AttachmentManagerType {
     func open(_ attachment: MessageAttachment)
 }
@@ -49,12 +50,18 @@ extension AttachmentManager: AttachmentManagerType {
             .sinkFuture(
                 receiveValue: {},
                 receiveError: { [weak self] error in
-                    self?.controller?.showToast(
-                        "\("message_attachment_saved_with_error".localized) \(error.localizedDescription)"
-                    )
+                    self?.handle(error)
                 }
             )
             .store(in: &self.cancellable)
+    }
+
+    private func handle(_ error: Error) {
+        Task {
+            await controller?.showToast(
+                "\("message_attachment_saved_with_error".localized) \(error.localizedDescription)"
+            )
+        }
     }
 }
 
--- a/FlowCrypt/Controllers/Compose/ComposeViewController.swift
+++ b/FlowCrypt/Controllers/Compose/ComposeViewController.swift
@@ -21,7 +21,6 @@ private struct ComposedDraft: Equatable {
  * - User can be redirected here from *InboxViewController* by tapping on *+*
  * - Or from *ThreadDetailsViewController* controller by tapping on *reply*
  **/
-@MainActor
 final class ComposeViewController: TableNodeViewController {
     private lazy var logger = Logger.nested(Self.self)
 
@@ -932,7 +931,6 @@ extension ComposeViewController: PHPickerViewControllerDelegate {
                     })
             }
         }
-        
     }
 }
 
--- a/FlowCrypt/Controllers/Inbox/InboxViewController.swift
+++ b/FlowCrypt/Controllers/Inbox/InboxViewController.swift
@@ -7,6 +7,7 @@ import FlowCryptCommon
 import FlowCryptUI
 import Foundation
 
+@MainActor
 final class InboxViewController: ASDKViewController<ASDisplayNode> {
     private lazy var logger = Logger.nested(Self.self)
 
--- a/FlowCrypt/Controllers/MessageList Extension/MsgListViewController.swift
+++ b/FlowCrypt/Controllers/MessageList Extension/MsgListViewController.swift
@@ -8,6 +8,7 @@
 
 import UIKit
 
+@MainActor
 protocol MsgListViewController {
     func open(with message: InboxRenderable, path: String)
 
@@ -30,11 +31,9 @@ extension MsgListViewController where Self: UIViewController {
     private func openDraft(with message: Message) {
         guard let email = DataService.shared.email else { return }
 
-        Task {
-            let controller = await ComposeViewController(email: email)
-            await controller.updateWithMessage(message: message)
-            await navigationController?.pushViewController(controller, animated: true)
-        }
+        let controller = ComposeViewController(email: email)
+        controller.updateWithMessage(message: message)
+        navigationController?.pushViewController(controller, animated: true)
     }
 
     private func openMsg(with message: Message, path: String) {
--- a/FlowCrypt/Common UI/Refreshable.swift
+++ b/FlowCrypt/Common UI/Refreshable.swift
@@ -8,7 +8,7 @@
 
 import Foundation
 
- protocol Refreshable {
-
-     func startRefreshing()
- }
+@MainActor
+protocol Refreshable {
+    func startRefreshing()
+}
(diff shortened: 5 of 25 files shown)


FlowCrypt/Common UI/AttachmentManager.swift
```diff
@@ -9,6 +9,7 @@
 import UIKit
 import Combine
 
+@MainActor
```
ivan-ushakov, line 12: `@MainActor` on type makes all methods and properties to be isolated so we could avoid marking each method with `@MainActor`.

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -21,7 +21,6 @@ private struct ComposedDraft: Equatable {
  * - User can be redirected here from *InboxViewController* by tapping on *+*
  * - Or from *ThreadDetailsViewController* controller by tapping on *reply*
  **/
-@MainActor
```
ivan-ushakov, line 24: Since `TableNodeViewController` is annotated with `@MainActor` we have it on each inherited type.

FlowCrypt/Controllers/Inbox/InboxViewController.swift
```diff
@@ -7,6 +7,7 @@ import FlowCryptCommon
 import FlowCryptUI
 import Foundation
 
+@MainActor
```
ivan-ushakov, line 10: Here we need to put `@MainActor` since `ASDKViewController` is objc class.

FlowCrypt/Controllers/MessageList Extension/MsgListViewController.swift
```diff
@@ -8,6 +8,7 @@
 
 import UIKit
 
+@MainActor
```
ivan-ushakov, line 11: Each protocol intended to work with UI state should be marked with `@MainActor` to be sure that we work with it on UI thread.

tomholub, line 11: On a controller, wouldn't this cause some methods that make api calls to run on main? Or as long as we `await` these other methods/services, then the API call would not be on main?

ivan-ushakov: It is possible but we should move such methods to services. I mean that view controller is something that interacts with view so we want to be sure that this happens on UI thread. That is why I want to have `@MainActor` on each controller.

tomholub: This is the danger I described earlier - if we just set it on the whole class, things may suddenly run on main that we didn't intend to. I'm not sure it's the best idea without properly studying how @ main actor works, ensuring that api requests are not on main in this situation:

```
@mainactor
class SomeViewController {

  fun handleTap() {
    Task {
       do {
          try await apiClient.someApiCall() // <-- does this run on main or not?
          handleGood()
        } catch {
          handleErr()
        }
    }
  }

  fun handleGood() {
  }

  fun handleErr() {
  }

}
```

Can you please study the docs and also step through this with debugger to make sure?

ivan-ushakov: If we run task like this:
```
@MainActor
class SomeViewController {

	func handleTap() {
		print("\(Thread.current): handleTap()")
		Task {
			do {
				print("\(Thread.current): inside Task point 1")
				try await apiClient.someApiCall()
				print("\(Thread.current): inside Task point 2")
				handleGood()
			} catch {
				handleErr()
			}
		}
	}

	func handleGood() {
		print("\(Thread.current): handleGood()")
	}

	func handleErr() {
		print("\(Thread.current): handleErr()")
	}
}
```

Then according to documentation:
> Runs the given nonthrowing operation asynchronously as part of a new top-level task on behalf of the current actor.

So we run task on main thread because we are inside `@MainActor`, but async function may run internally on any other thread. So at "point 1" we are on main thread, but on "point 2" we could be on any other thread, it depends on `apiClient.someApiCall()`. This is the reason why we use `@MainActor`, to be sure that `handleGood()` will be executed on main thread.

On the other hand if we start task like this:
```
@MainActor
class SomeViewController {

	func handleTap() {
		print("\(Thread.current): handleTap()")
		Task.detached {
			do {
				print("\(Thread.current): inside Task point 1")
				try await self.apiClient.someApiCall()
				print("\(Thread.current): inside Task point 2")
				await self.handleGood()
			} catch {
				await self.handleErr()
			}
		}
	}

	func handleGood() {
		print("\(Thread.current): handleGood()")
	}

	func handleErr() {
		print("\(Thread.current): handleErr()")
	}
}
``` 

We have new top level task:
> Runs the given nonthrowing operation asynchronously as part of a new top-level task.

So we run task on another thread (no main), and in "point 1" we are not on main thread. But again, on "point 2" it is possible to be on new thread.

So it depends on how internally works async function.

ivan-ushakov: Maybe we need to find out what we want to run on separate thread, like network requests (already implemented in `URLSession` by Apple), our JavaScript functions (Core is actor already), etc.

tomholub: Can we mark these api classes etc with something like BackgroundActor ? :-)

ivan-ushakov: I need to check if it is possible to create our own annotations like this.

ivan-ushakov: Looks like there is no way to define your own annotations in Swift, so there is no way to have `@BackgroundActor`. On the other hand any user actor by default is "background".

tomholub: got it

tomholub: > Maybe we need to find out what we want to run on separate thread, like network requests (already implemented in URLSession by Apple), our JavaScript functions (Core is actor already), etc.

All API client classes, including Gmail API (which may not be using URLSession - not sure), and Core as you say. IMAP/SMTP client as well. So here the main thing to look into is Gmail and IMAP/SMTP.

FlowCrypt/Common UI/AttachmentManager.swift
```diff
@@ -49,13 +50,19 @@ extension AttachmentManager: AttachmentManagerType {
             .sinkFuture(
```
sosnovsky, line 50: @tomholub I think we should use `async/await` here instead of Combine for code consistency across project, and for better threads handling with `@MainActor`, can be done in another issue

FlowCrypt/Controllers/MessageList Extension/MsgListViewController.swift
sosnovsky: It's possible to create our own custom actors with `@globalActor` declaration, like
```swift
@globalActor
struct BackgroundActor {
...
}
```

But I think we don't need such `@BackgroundActor` declaration as Swift automatically runs `async/await` operations not on the main thread, so we just need to re-check our API calls to use `async`

sosnovsky (approved):
Works well ✅

FlowCrypt/Common UI/AttachmentManager.swift
tomholub: Agree - I've asked @ekievsky to do it as part of #1031

tomholub:
Then I'll merge this, and we make sure that we use actor for Gmail/IMAP/SMTP in another issue.

tomholub (approved):

FlowCrypt/Common UI/AttachmentManager.swift
tomholub: Will also file another issue to look over the remaining ones after that, if any

PR #1062, opened 2021-11-22
Revert "Use @MainActor instead of DispatchQueue.main.async"

tomholub:
Reverts FlowCrypt/flowcrypt-ios#1055 - I merged it too early, CI doesn't seem to pass.


tomholub:
this is not what I wanted to revert

Issue #953, opened 2021-11-08
reply all functionality


tomholub:
currently, we have a reply button. should also add reply all, and differentiate the two.

tomholub:
please follow Gmail: add reply-all under the menu for all messages in thread, and then add the three reply/reply-all/forward buttons on the very bottom. Thanks!

sosnovsky:
I think this task should include #921 too, as currently users don't see when mail has multiple recipients.

Also Gmail app by default triggers `reply all`, and I didn't find way to reply only for message sender (only by removing other recipients). I think it's because `reply all` is used more often than just `reply`, maybe we should use `reply` button for `reply all` and add `Reply` option to message menu?

tomholub:
Assigned that one. Looking into reply/reply all. For me, the default on Gmail app is reply-to-one-person.

I think in Gmail this is configurable.

sosnovsky:
Interesting, I haven't changed any settings in Gmail app after installing, maybe they take `reply` settings from Google account. Okay, let's do `Reply all` in message menu, and `Reply to one` as button

tomholub:
https://mail.google.com/mail/u/0/#settings/general

![image](https://user-images.githubusercontent.com/6306961/143444946-9ed6bd87-aafd-4912-a90e-4b26695a3c71.png)

But interestingly, after changing this (on desktop) it has no affect on the mobile app. Still the default is reply to one even if I reply to a conversation.

To be fair, I would expect the same as you - that default is reply all. On the other hand, our main customer seems to like reply-to-one as default (recently they submitted a bug to browser extension: https://github.com/FlowCrypt/flowcrypt-browser/issues/4077 )

On top of all of this, it defaults to reply-all for you and reply for me :shrug: 

Let's start with reply to one, based on customer feedback above. Then when we are wiser, we may adjust it.

Issue #723, opened 2021-10-21
[UX] Add 3 dots button for gmail thread message


Kharchevskyi:
or add similar 3 dots button which will show popup with options to choose option(mark as unread, delete message, archive) for  selected message. And buttons on top will acts on whole threads

_Originally posted by @Kharchevskyi in https://github.com/FlowCrypt/flowcrypt-ios/issues/608#issuecomment-948845791_

tomholub:
closing as a dupplicate the top icons that act on whole thread are already present, three dots will be added in PR #1066 and Gmail app doesn't have a way to modify individual messages anyway.

tomholub:
There's still three dots that act on whole thread (in Gmail app) but we don't need any of that functionality yet.

PR #1066, opened 2021-11-23
#900 Messages forwarding

### LINKED ISSUE #900: forwarding email functionality
the iOS app has a reply button but is missing a forward button

for now skip attachments - don't include them when forwarding

### LINKED ISSUE #952: remove expand icon when message expanded
> Works well. Maybe we should move the reply icon a tiny bit further from the chevron icon? Seem like they are a bit too close together. For spacing, can follow the space between the header icons which are comfortably apart. (people with fat fingers etc)

Agree, looks quite close. I was also thinking about removing `expand` icon when message is expanded, so it'll be only `reply` button, as currently `expand` icon looks too repetitive, as for me. What do you think?

_Originally posted by @sosnovsky in https://github.com/FlowCrypt/flowcrypt-ios/issues/947#issuecomment-963527437_

sosnovsky:
This PR adds forwarding functionality

close #900
close #952

- Tests added or updated - added ui test for forward functionality

### To be filled by reviewers


FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
```diff
@@ -118,21 +120,40 @@ extension ThreadDetailsViewController {
...
-        let replyInfo = ComposeMessageInput.ReplyInfo(
-            recipient: input.rawMessage.sender,
-            subject: input.rawMessage.subject,
+        let recipient: String?
+        switch quoteType {
+        case .reply:
+            recipient = input.rawMessage.sender
+        case .forward:
+            recipient = nil
+        }
```
tomholub, line 142: Might as well change the recipient to an array, as will be needed for replyAll anyway. #953 . Then instead of nil, it will be empty array?

tomholub (commented):
code looks good 👍 see comment below. Not required but will come handy.

FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
sosnovsky: Yeah, having array here will be more future-proof, I'll update code

sosnovsky: Should I also add UI test for messages forwarding or it'll be separate issue?

tomholub: Yes - please add UI test that opens encrypted message and taps forward. Then evaluate that the quote is present, and that there are no recipients. And the subject. That would end the test successfully (tests that don't change state are easier for now, until we test against mock of gmail api)

tomholub:
Functionality works well but I find the arrows confusing (didn't think about that before). Could you please, instead of the forward button, add three dots that pulls up a menu - where forward is one of the options (like Gmail app does it).

sosnovsky:
Then may be we should show 3 dots instead of `expand/collapse` arrow and menu will include `forward` and `collapse` actions?

tomholub (commented):
Code and test looks great, except the UX concern above.

tomholub:
> Then may be we should show 3 dots instead of `expand/collapse` arrow and menu will include `forward` and `collapse` actions?

I'd like to follow Gmail in functionality since users will be migrating from Gmail app to here. Gmail doesn't have a chevron, true. But it does allow to expand/collapse by tapping the message header. Here's a related issue: https://github.com/FlowCrypt/flowcrypt-ios/issues/952

I guess, what you proposed brings us closer there. I just don't know if it's unnecessary work, just for the "collapse" option to be removed again in #952

tomholub:
Actually it seems like we already support opening message by tapping anywhere. So it's just about collapsing. 

Actually, just by adding the collapse functionality when tapping expanded message header, and then adding a horizontal line between messages in thread, you would have also fixed #952 at the same time. Does it make sense in one PR? Or should I merge this PR and assign that one to you?

sosnovsky:
Let's do it in 1 PR, so forwarding functionality will have final UX before merging.
You can assign #952 to me and I'll include it here

tomholub:
This also becomes quite a duplicate of #723 which I'll close now.

FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
```diff
@@ -400,8 +414,8 @@ extension ThreadDetailsViewController: ASTableDelegate, ASTableDataSource {
     func tableNode(_ tableNode: ASTableNode, numberOfRowsInSection section: Int) -> Int {
         guard section > 0, input[section-1].isExpanded else { return 1 }
 
-        let count = input[section-1].processedMessage?.attachments.count ?? 0
-        return Parts.allCases.count + count
+        let attachmentsCount = input[section-1].processedMessage?.attachments.count ?? 0
+        return Parts.allCases.count + attachmentsCount
```
tomholub, line 418: :+1:

tomholub, line 418: :+1:

tomholub (commented):
code looks good

tomholub (approved):
Functionality works well. One improvement: when forwarding (unlike reply) the recipients field should be in automatic focus when the compose view renders. When replying, the focus should be on message body as it does now.

Issue #1071, opened 2021-11-23
tests that open emails in inbox should search for them first


tomholub:
For example `Error: element ("~Test 1") still not displayed after 15000ms` is because the desired email is too low on the screen now, and the test doesn't see it.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1070#issuecomment-976469472_

tomholub:
As we add more test emails into the inbox, some inbox become out of view.

To find the right email to test with, going forward, we have to search for the email and then open it that way.

tomholub:
Or categorize emails using labels in Gmail, and then open a particular label in left menu to open the email.

tomholub:
@fcvakintos after finalizing open PRs, please do this one next - it's still creating some friction as it is now.

fcvakintos:
@tomholub I will add ScrollToElement method

tomholub:
no, please absolutely do not do that :-) use the search functionality.

tomholub:
Eg for sent emails that we specifically want to see in sent folder, sure - use ScrollToElement. But for general tests that open one of the inbox emails and check contents etc, use search.

PR #1088, opened 2021-11-24
Signed-only messages not recognized as signed

### LINKED ISSUE #1020: signed-only messages not recognized as signed + ui tests
> I want to test it with `flowcrypt.compatibility` account, as you mentioned there is an issue with signed but not encrypted messages. But as it's quite rare case we can leave it for another issue, is it okay?

I'll file another issue for it, not necessary to do now.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/998#issuecomment-969305108_

ivan-ushakov:
close #1020

- Tests added or updated

### To be filled by reviewers


ivan-ushakov:
Still fighting with email crafting for UI tests

appium/tests/screenobjects/inbox.screen.ts
```diff
@@ -28,10 +28,12 @@ class InboxScreen extends BaseScreen {
...
     const selector = `~${subject}`;
-    if (await (await $(selector)).isDisplayed() !== true) {
-      await TouchHelper.scrollDown();
+    if (withScroll) {
+      if (await (await $(selector)).isDisplayed() !== true) {
+        await TouchHelper.scrollDown();
+      }
     }
     await ElementHelper.waitAndClick(await $(selector), 500);
   }
```
tomholub, line 39: This is a mess (not your fault), should be addressed here https://github.com/FlowCrypt/flowcrypt-ios/issues/1071

Was the scroll causing you trouble? It's supposed to only scroll when the email is not present, but it's sloppily written - it's testing too fast. Maybe it wasn't loaded yet. Try, instead of this `withScroll`, to add a 2000ms wait before you call `clickOnEmailBySubject`.

appium/tests/screenobjects/inbox.screen.ts
```diff
@@ -29,10 +29,8 @@ class InboxScreen extends BaseScreen {
...
 
   clickOnEmailBySubject = async (subject: string) => {
+    await browser.pause(2000); // todo: loading inbox. Fix this: wait until loader gone
     const selector = `~${subject}`;
-    if (await (await $(selector)).isDisplayed() !== true) {
-      await TouchHelper.scrollDown();
-    }
```
tomholub, line 35: This will still be needed, else other tests will start failing. Some messages are on the bottom of the list, and the app needs to scroll to see them. (until fixed with #1071 )

The adding of a pause was to make sure that initial messages can be loaded, based on which `if (await (await $(selector)).isDisplayed() !== true) {` can make a proper decision whether to scroll or not.

ivan-ushakov: In some cases I don't need to scroll. So I can keep this one and create additional method without scrolling, OK?

ivan-ushakov:
@tomholub
Do you know how to create this two?

1. signed-only message where the pubkey is not available
2. signed-only message that was tempered during transit

For the first one I can not use `mime-email-plain-signed.txt` because it has text with special symbol. Also I want to follow my convention for the subjects and text.

appium/tests/screenobjects/inbox.screen.ts
tomholub: In the cases when you don't need to scroll, `if (await (await $(selector)).isDisplayed() !== true) {` should recognize that scroll is not needed. And therefore it should not scroll. Can you confirm?

I know this is a mess as it is, I'm just trying to not make it even worse.

ivan-ushakov:
1. signed-only message where the pubkey is not available
For this one I used my email generator and created signed message with generated key pair (not published anywhere).

2. signed-only message that was tempered during transit
For this one I just sent signed email with browser extension, downloaded original message in MIME and changed message text from `Signed only message that was tempered during transit` to `Signed only message that was Tempered during transit` and imported with Thunderbird.

ivan-ushakov:
UI tests failing because I added new messages and now you need to scroll down to have them visible. I agree about using Search functionality in UI tests for this.

tomholub:
> UI tests failing because I added new messages and now you need to scroll down to have them visible. I agree about using Search functionality in UI tests for this.

in progress here https://github.com/FlowCrypt/flowcrypt-ios/pull/1103

ivan-ushakov:
I moved UI test to search approach to save some time.

tomholub:
@ivan-ushakov master tests are fixed now

appium/tests/specs/inbox/CheckAllEmailSignatureCases.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    await EmailScreen.clickBackButton();
+
+    await SearchScreen.clickBackButton();
+
+    // signed only message where the pubkey is not available
+    await MailFolderScreen.clickSearchButton();
+    await SearchScreen.searchAndClickEmailBySubject('Signed only message where the pubkey is not available');
+
+    await EmailScreen.checkEncryptionBadge('decrypt error');
+    await EmailScreen.clickBackButton();
```
tomholub, line 63: This one is not right - signed only message that is missing pubkey should have encryption badge of `not encrypted` and a signature badge that says it cannot be verified due to missing pubkey.

tomholub (commented):
This is excellent! With one problem below.

I'm not sure if the problem is in the supplied message or in the app code.

tomholub (approved):
I'll merge it either way, as it is an improvement and adds a great test.

But that case needs to be debugged.

appium/tests/specs/inbox/CheckAllEmailSignatureCases.spec.ts
ivan-ushakov: Maybe email was crafter with error. Let's create separate issue

Issue #840, opened 2021-10-28
use login hint for second sign-in if we already know the user email


tomholub:
It can be improved further: on the second sign-in attempt, supply `login hint` into the google sign-in request. That way, I don't have to choose my email again, instead my email will be pre-chosen based on the first sign-in.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/663#issuecomment-954246152_

tomholub:
Follow browser extension behavior. To understand this issue:

1) go to accounts.google.com on your account -> 3rd party apps -> flowcrypt ios -> revoke
2) login on ios app with that account, but don't tick Gmail permission during the grant screen
3) you will be prompted to login / grant again

current behavior: the user has to select their email account address again, especially if you are logged into more than one account

expected behavior: since user already signed in once (even though without granting permission), we should already know user email. The second login prompt should use that email address in a `login_hint` so that user doens't have to choose their email address again.

The very same thing applies when user then goes to compose screen and wants to grant contacts permission - `login_hint` should always be used there since user is already logged in.

ioanmo226:
@tomholub I agree that we need to provide `login_hint` in compose screen for grant contacts permission.

But for login screen, let's say user logged in with `a@gmail.com` google account(without ticking Gmail permission during the grant screen)
If we pass login_hint as `a@gmail.com` in this case, there will be no way for user to log out this `a@gmail.com` account.
Please check below screenshot.
I think we should not pass `login_hint` for login screen.

![image](https://user-images.githubusercontent.com/77344191/159523344-6935a811-b157-45c3-acdc-7a85718b4bae.png)

ioanmo226:
@tomholub

ioanmo226:
If user taps `Cancel` it will return to login screen and when user taps `continue with gmail`, above screen will appear again.
So IMHO it's better we should not set `login_hint` for login screen.

tomholub:
Ok, for now, let's only add login_hint on the contacts prompt. Let's leave this issue open though. I'd like to discuss the login flow again sometime later.

ioanmo226:
OK

PR #1090, opened 2021-11-25
#903 Include attachments in forwarded messages

### LINKED ISSUE #903: include attachments in forwarded message
after #900 

I'll file another issue for including attachments in the forwarded message

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/900#issuecomment-958945135_

sosnovsky:
This PR adds attachments support for forwarded messages

close #903

- Tests added or updated - updated UI test to check attachment presence

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
+++ b/FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
@@ -156,6 +156,7 @@ extension ThreadDetailsViewController {
             : []
 
         let subject = input.rawMessage.subject ?? "(no subject)"
+        let threadId = quoteType == .reply ? input.rawMessage.threadId : nil
 
         let replyInfo = ComposeMessageInput.MessageQuoteInfo(
             recipients: recipients,
@@ -164,7 +165,7 @@ extension ThreadDetailsViewController {
             mime: processedMessage.rawMimeData,
             sentDate: input.rawMessage.date,
             message: processedMessage.text,
-            threadId: input.rawMessage.threadId,
+            threadId: threadId,
             attachments: attachments
         )
 
--- a/FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
+++ b/FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
@@ -147,13 +147,13 @@ extension ThreadDetailsViewController {
               let processedMessage = input.processedMessage
         else { return }
 
-        let recipients: [String]
-        switch quoteType {
-        case .reply:
-            recipients = [input.rawMessage.sender].compactMap { $0 }
-        case .forward:
-            recipients = []
-        }
+        let recipients = quoteType == .reply
+            ? [input.rawMessage.sender].compactMap({ $0 })
+            : []
+
+        let attachments = quoteType == .forward
+            ? input.processedMessage?.attachments ?? []
+            : []
 
         let subject = input.rawMessage.subject ?? "(no subject)"
 
@@ -164,7 +164,8 @@ extension ThreadDetailsViewController {
             mime: processedMessage.rawMimeData,
             sentDate: input.rawMessage.date,
             message: processedMessage.text,
-            threadId: input.rawMessage.threadId
+            threadId: input.rawMessage.threadId,
+            attachments: attachments
         )
 
         let composeInput = ComposeMessageInput(type: .quote(replyInfo))
--- a/FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
+++ b/FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
@@ -446,7 +446,8 @@ extension ThreadDetailsViewController: ASTableDelegate, ASTableDataSource {
                 let attachment = message.attachments[indexPath.row - 2]
                 return AttachmentNode(
                     input: .init(
-                        msgAttachment: attachment
+                        msgAttachment: attachment,
+                        index: indexPath.row - 2
                     ),
                     onDownloadTap: { [weak self] in self?.attachmentManager.open(attachment) }
                 )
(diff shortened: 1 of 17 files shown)


sosnovsky:
@tomholub I currently have only 2 assigned tasks left - #953 and #840, can you please check which tasks should I work after these ones?

tomholub:
Very gladly :-)

appium/tests/screenobjects/new-message.screen.ts
```diff
@@ -112,6 +122,17 @@ class NewMessageScreen extends BaseScreen {
     expect(value).toEqual(`  ${recipient}  `);
   };
 
+  attachmentName = async (name: string) => {
+    const selector = `-ios class chain:**/XCUIElementTypeStaticText[\`label == "${name}"\`]`;
```
tomholub, line 126: I would add incremental accessibility ids to attachments. So first attachment would be `attachment1` then `attachment-1` and so on (or you can start with 1 if you want)

Then you would be getting attachment-0 in tests by selector, and make sure it has the right content.

I think that's the most prudent way to test.

FlowCryptUI/Nodes/AttachmentNode.swift
```diff
@@ -43,6 +43,7 @@ public final class AttachmentNode: CellNode {
         imageNode.tintColor = .gray
         buttonNode.tintColor = .gray
         deleteButtonNode.setImage(UIImage(named: "cancel")?.tinted(.gray), for: .normal)
+        deleteButtonNode.accessibilityIdentifier = "deleteAttachmentButton"
```
tomholub, line 46: Also here if possible to enumerate them? So this would be `deleteAttachmentButton1` maybe.

appium/tests/specs/inbox/CheckReplyAndForwardForEncryptedEmail.spec.ts
```diff
@@ -33,6 +34,6 @@ describe('INBOX: ', () => {
     await NewMessageScreen.clickBackButton();
     await EmailScreen.clickMenuButton();
     await EmailScreen.clickForwardButton();
-    await NewMessageScreen.checkFilledComposeEmailInfo("", forwardSubject, quoteText);
+    await NewMessageScreen.checkFilledComposeEmailInfo("", forwardSubject, quoteText, attachmentName);
```
tomholub, line 37: You could also test that the attachment disappears when you tap `deleteAttachmentButton`. It's out of scope of this PR but I don't think we're testing it anywhere.

tomholub (commented):
Looks good! Will test. Comments below.

appium/tests/screenobjects/new-message.screen.ts
sosnovsky: Yes, it should be a good solution for quicker search of elements 👍

FlowCryptUI/Nodes/AttachmentNode.swift
sosnovsky: Will add enumeration here too

tomholub:
I got this when forwarding a 9+ mb video. But maybe that's for another issue?

image

tomholub:
I suppose the `xlsx files malformed when forwarding` issue should be coupled with some unit test as it seems an easy thing to break again in the future.

FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
```diff
@@ -156,6 +156,7 @@ extension ThreadDetailsViewController {
             : []
 
         let subject = input.rawMessage.subject ?? "(no subject)"
+        let threadId = quoteType == .reply ? input.rawMessage.threadId : nil
```
tomholub, line 159: Let's test by `!= .forward`, then you don't have to adjust it (or forget to adjust it) once reply-all is added.

sosnovsky: But I think `quoteType` for reply-all will still be `reply`, but `recipients` array will contain more emails

tomholub: Got it. I was just suggesting blindly without context. All good.

sosnovsky:
> 4\. downloaded file doesn't get recognized as a xlsx file. It does have the right extension, and it does seem to have the right contents. But maybe the mime-type is wrong / not preserved. Compare to forwarding with browser extension, which works. (issue also present on a jpeg file, and likely most others)

I tried to reproduce this issue, but `.xlsx` file opens well after forwarding from iOS. Maybe it happens only with some specific files, I forwarded the same message to your email, can you please check if it works or not for you?

tomholub:
> > 4. downloaded file doesn't get recognized as a xlsx file. It does have the right extension, and it does seem to have the right contents. But maybe the mime-type is wrong / not preserved. Compare to forwarding with browser extension, which works. (issue also present on a jpeg file, and likely most others)
> 
> I tried to reproduce this issue, but `.xlsx` file opens well after forwarding from iOS. Maybe it happens only with some specific files, I forwarded the same message to your email, can you please check if it works or not for you?

The issue (missing mime type inside the encrypted enclosure) is present on all files. But the user-facing problem (can't open as a result) only manifests on some host OS, and possibly only for some files.

sosnovsky:
I tested sending and forwarding message with 2 attachments (image and xls file) - in both cases app sends right content-types (`image/jpeg` and `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`) to `Core.composeEmail` function.
Then I downloaded these attachments from browser on macOS and checked mime types - they were the same as sent by the app.
So I think we should create another issue for finding why attachments mime-type is wrong on Linux OS, as it's not only for forwarded messages. May be some difference in decrypting `.pgp` files.

tomholub:
Ok. I'll let this merge and we'll investigate it separately.

tomholub (approved):

PR #1098, opened 2021-11-25
test commnit

ekievsky:
This PR <what does it do and why?>

close #0000 // if this PR closes an issue

- Difficult to test (explain why)
- Not worth testing
- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCryptUIApplication/AppDelegate.swift
+++ b/FlowCryptUIApplication/AppDelegate.swift
@@ -8,6 +8,8 @@
 
 import UIKit
 
+
+
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
     var window: UIWindow?


tomholub:
ui tests passed, but swift tests hang on Semaphore. And then it doesn't load the contents when I click on it. Seems like s semaphore bug, you can try re-running it.

But it does demostrate that the appium tests are fine, at least on master.

image

tomholub:
I wrote semaphore about not being able to see the log on semaphore, their response:

```
Hi Tom,

Thanks for contacting us.

This might be a known issue that occurs when the page needs to display a long output log. The identified problem happens when a command writes several thousand lines of log.

In order to debug this, you could identify the commands with the largest output, they are usually installations with a progress bar, code climate reports or bundle, and redirect the output to a file:

user-command >> output.log Then it's possible to get the log if needed with cat output.log

I suggest also trying an SSH session since it could be a potential time-saver with this situation.

Let me know your findings.

Wishing you a great weekend ahead!

Kind regards,
--
Mateus Capistrano
Semaphore Support 
```

PR #1103, opened 2021-11-26
added new method -  scrollToElement, updated test, added checking Inb…

### LINKED ISSUE #1071: tests that open emails in inbox should search for them first

fcvakintos:
…ox screen after reopening app

This PR contains fixes for tests, added new method for scrolling down to element

close #1071 

**Tests** 
- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/appium/tests/helpers/TouchHelper.ts
+++ b/appium/tests/helpers/TouchHelper.ts
@@ -1,4 +1,5 @@
 
+
 class TouchHelper {
 
   /**
@@ -29,6 +30,27 @@ class TouchHelper {
       {action: 'release', options: {}},
     ]);
   }
+
+  static scrollDownToElement = async (element: WebdriverIO.Element) => {
+    const { width, height } = await driver.getWindowSize();
+    const anchor = width / 2;
+    const startPoint = height * 0.25;
+    const endPoint = height * 0.15;
+    let index = 0;
+
+    do {
+      await browser.pause(1000); // due to scroll action which takes about second
+      await driver.touchPerform([
+        {action: 'press', options: {x: anchor, y: startPoint}},
+        {action: 'wait', options: {ms: 100}},
+        {action: 'moveTo', options: {x: anchor, y: endPoint}},
+        {action: 'release', options: {}},
+      ]);
+    } while (await (await element).isDisplayed() !== true  && index++ < 14);
+
+    if(index === 15) throw new Error(`Element ${JSON.stringify(element.selector)} doesn't displayed after scroll`);
+  }
+
 }
 
 export default TouchHelper;
--- a/appium/tests/helpers/TouchHelper.ts
+++ b/appium/tests/helpers/TouchHelper.ts
@@ -34,23 +34,24 @@ class TouchHelper {
   static scrollDownToElement = async (element: WebdriverIO.Element) => {
     const { width, height } = await driver.getWindowSize();
     const anchor = width / 2;
-    const startPoint = height * 0.25;
+    const startPoint = height * 0.75;
     const endPoint = height * 0.15;
-    let index = 0;
 
-    do {
-      await browser.pause(1000); // due to scroll action which takes about second
+    // this wait can be later replaced by waiting for loader to go away before scrolling
+    await browser.pause(1000); // make sure contents are loaded first, so we don't scroll too early
+    for(let i = 0; i < 15; i++) {
+      if (await (await element).isDisplayed()) {
+        return;
+      }
       await driver.touchPerform([
         {action: 'press', options: {x: anchor, y: startPoint}},
         {action: 'wait', options: {ms: 100}},
         {action: 'moveTo', options: {x: anchor, y: endPoint}},
         {action: 'release', options: {}},
       ]);
-    } while (await (await element).isDisplayed() !== true  && index++ < 14);
-
-    if(index === 15) throw new Error(`Element ${JSON.stringify(element.selector)} doesn't displayed after scroll`);
+      await browser.pause(1000); // due to scroll action which takes about second
+    }
+    throw new Error(`Element ${JSON.stringify(element.selector)} not displayed after scroll`);
   }
-
 }
-
 export default TouchHelper;
--- a/appium/tests/screenobjects/mail-folder.screen.ts
+++ b/appium/tests/screenobjects/mail-folder.screen.ts
@@ -66,6 +66,7 @@ class MailFolderScreen extends BaseScreen {
   }
 
   clickOnEmailBySubject = async (subject: string) => {
+    await expect(await this.helpIcon).toBeDisplayed();
     const selector = `~${subject}`;
     if (await (await $(selector)).isDisplayed() !== true) {
       await TouchHelper.scrollDownToElement(await $(selector));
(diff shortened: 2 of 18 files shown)


tomholub (commented):
@fcvakintos this is an improvement but it does not close the issue. The issue was about using the search functionality to find inbox elements, instead of scrolling to them.

You have to have a long-term view. We started with 1 email. Now there's 20. There will probably be 300 by the time we are done writing all these tests. Scrolling to them is not scalable.

tomholub:
I can merge this but change the `close` to `issue` in the description, and then followup with a PR that uses the search icon.

fcvakintos:
@tomholub I will update the test

tomholub:
Do you want to do it in this PR or separately?

fcvakintos:
@tomholub in this PR

tomholub:
Ok. I marked it as draft until then. To recap:
 - for tests that check message content of a particular message, or test the reply button (without actually sending reply) etc : use search
 - for tests that go from folder to folder (sent / trash etc) use scroll

fcvakintos:
@tomholub I think search is not working correctly
I tried to search by email subject and get a lot of emails in response
try with this subject - "encrypted email"

Screenshot 2021-11-26 at 15 11 43

fcvakintos:
@tomholub and it search from all list(Sent/Inbox and other) - I think it should search only from the list of selected section

tomholub:
I think it's because "encrypted email" is included as a PGP header comment in every encrypted email

```
-----BEGIN PGP MESSAGE-----
Version: FlowCrypt iOS 0.2 Gmail Encryption
Comment: Seamlessly send and receive encrypted email

wV4DZQdhGPK/uXASAQdAB5494lpT81X77JPQp7Un547GRSUVnEBcv8IvEl5u
9Vww2GgQifhIQDu7kB3McqzL5y+24hiiALM8HRaKnJrAVAV2iKIl7xlIN6le
Nvu2tYGfwV4DFS6Vc++zZ+oSAQdAngKdfuE8RwlwZj741VPUMGHvM5KuQnHI
gYnd5o+wbWQwVWlEk6tG4AtvWrttqW/jGz36vInCs+10cfcm7YUH7XmZLrM0
OR/K6q88L/ANFwaH0sC7AamuqJVrmcAQsy0KTQNURwWpucEb7GlsweitvuyG
Z7BZBpm4CXGsmudk6eYHleEHrD5DH4VmJS1W8VgbpXXklLE3RN+s0R/73Xt/
3WDkzKnE/a8qg+SxpTtcK6QvzHSTEaVdtFVusQ5iwgrS95PD4T1SPYgwKU98
WTar1k2bf8jIKq7xPTvldlsmmsJBIyT3ns7zvM7iNu+IMfgelNTHFFfqJija
MNep5WvA035dUtPOA88DhnaSMVgcYqojc4vrZ3s6m1UMUm9Ik/PSBj7CM7Vv
KWS8jMNPhaXCCVvFZxTPKyHhHJrBT/HIZrGN/Acy8lLCUIhxMQ56PQCrTLz8
X16Z4SDkizmRHms/CBtCLLPJJqK1NpywmmG7eFcXkObpA9lP64Fe7WeQuzom
xj8+B2CwsesjstFUVIjXWWat4eSf1BC20ffa8ERa9P2akiRJzALlA7D/ovif
uOegizsO3bzFD33998lFMXgjCr61AY3UF3ZW9DRncsUIZw==
=NcWZ
-----END PGP MESSAGE-----
```

so that's not a fortunate combination of words to search. You can do the following things:
 - rename that email to something less generic
 - try to put the search term in quotes (which won't help this particular situation, but could help in general)

tomholub:
The search works on all folders on purpose - users were getting confused when they couldn't find email. You can try if searching adding `"search string here" is:inbox` will help.

fcvakintos:
@tomholub yeah, this case works `try a query like subject:"search string here" which would only search subject - hopefully
`
thanks

appium/tests/helpers/TouchHelper.ts
```diff
@@ -29,6 +30,27 @@ class TouchHelper {
       {action: 'release', options: {}},
     ]);
   }
+
+  static scrollDownToElement = async (element: WebdriverIO.Element) => {
+    const { width, height } = await driver.getWindowSize();
+    const anchor = width / 2;
+    const startPoint = height * 0.25;
+    const endPoint = height * 0.15;
```
tomholub, line 38: could probably do a more "generous" scroll like from 0.75 to 0.15. So that we don't have to scroll so many times.

appium/tests/helpers/TouchHelper.ts
```diff
@@ -29,6 +30,27 @@ class TouchHelper {
...
+
+    do {
+      await browser.pause(1000); // due to scroll action which takes about second
+      await driver.touchPerform([
+        {action: 'press', options: {x: anchor, y: startPoint}},
+        {action: 'wait', options: {ms: 100}},
+        {action: 'moveTo', options: {x: anchor, y: endPoint}},
+        {action: 'release', options: {}},
+      ]);
+    } while (await (await element).isDisplayed() !== true  && index++ < 14);
```
tomholub, line 49: This kind of loop declaration is prone to off-by-one errors. I can't confirm if it's right or not from looking at it, and I think it's probably indeed off by one.

I recommend to do:

```ts
// this wait can be later replaced by waiting for loader to go away before scrolling
await browser.pause(1000); // make sure contents are loaded first, so we don't scroll too early
for(let i = 0; i < 15; i++) {
    if (await (await element).isDisplayed()) {
        return;
    }
    await driver.touchPerform([
      {action: 'press', options: {x: anchor, y: startPoint}},
      {action: 'wait', options: {ms: 100}},
      {action: 'moveTo', options: {x: anchor, y: endPoint}},
      {action: 'release', options: {}},
    ]);
    await browser.pause(1000); // due to scroll action which takes about second
}
throw new Error(`Element ${JSON.stringify(element.selector)} not displayed after scroll`);
```

You should find it easier to read and follow when you structure it that way. And it will more likely be correct.

appium/tests/specs/inbox/CheckReplyAndForwardForEncryptedEmail.spec.ts
```diff
@@ -19,12 +19,13 @@ describe('INBOX: ', () => {
 
     const replySubject = `Re: ${emailSubject}`;
     const forwardSubject = `Fwd: ${emailSubject}`;
-    const quoteText = `On 11/5/21 at 4:15 AM ${senderEmail} wrote:\n > ${emailText}`;
+    const quoteText = `On 11/5/21 at 1:15 PM ${senderEmail} wrote:\n > ${emailText}`;
```
tomholub, line 22: I keep seeing people switch this line this back end forth between these two results. It seems to depend on the device that the test runs on, and its timezone, or some similar environment problem we cannot control.

Use a regex instead, to allow the time to be any number.

appium/tests/screenobjects/mail-folder.screen.ts
```diff
@@ -8,7 +8,8 @@ const SELECTORS = {
   CREATE_EMAIL_BUTTON: '-ios class chain:**/XCUIElementTypeButton[`label == "+"`]',
   INBOX_HEADER: '-ios class chain:**/XCUIElementTypeStaticText[`label == "INBOX"`]',
   SEARCH_ICON: '~search icn',
-  HELP_ICON: '~help icn'
+  HELP_ICON: '~help icn',
+  SEARCH_FIELD: '~Search All Mail'
```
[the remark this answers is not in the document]
fcvakintos: I added only for SEARCH_FIELD, but I dont know how to add for help icon and for search icon
who from Devs can help me with this?

tomholub (approved):
Thanks!

PR #1109, opened 2021-11-28
Rebuild InvalidStorageViewController with Texture

### LINKED ISSUE #766: storage error screen uses UIKit - guidance with ADK
Unfortunately I fail to make this screen using `ADK` so I switched to pure `UIKit`.

_Originally posted by @ivan-ushakov in https://github.com/FlowCrypt/flowcrypt-ios/issues/761#issuecomment-950375973_

What was the issue you were having?

Kharchevskyi:
This PR rebuild `InvalidStorageViewController` with Texture. Added UI Guidance

close #766 
IMG_E723E2654C50-1

**Tests**:

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Bootstrap/InvalidStorageViewController.swift
+++ b/FlowCrypt/Controllers/Bootstrap/InvalidStorageViewController.swift
@@ -6,9 +6,18 @@
 //  Copyright © 2017-present FlowCrypt a. s. All rights reserved.
 //
 
-import UIKit
+import FlowCryptUI
+import AsyncDisplayKit
 
-final class InvalidStorageViewController: UIViewController {
+@MainActor
+final class InvalidStorageViewController: TableNodeViewController {
+    private enum Parts: Int, CaseIterable {
+        case screenTitle
+        case title
+        case description
+        case button
+    }
+
     private let error: Error
     private let encryptedStorage: EncryptedStorageType
     private let router: GlobalRouterType
@@ -17,7 +26,7 @@ final class InvalidStorageViewController: UIViewController {
         self.error = error
         self.encryptedStorage = encryptedStorage
         self.router = router
-        super.init(nibName: nil, bundle: nil)
+        super.init(node: TableNode())
     }
 
     @available(*, unavailable)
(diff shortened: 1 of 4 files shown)


tomholub (commented):
This is an interesting example - I find that the UIKit example was more readable and overall more straight forward. From this perspective, it may make sense to stop writing new controllers in Texture, and start migrating away from it.

I wonder what would it look like in SwiftUI.

Kharchevskyi:
What if we would need to add one new element for example some extra button between views. We would need to relayout all views accordingly. 
With UIKit tableview this would be the same as in this PR but with instead uitableview and manual size calculations

Kharchevskyi:
I would prefer to love towards swift ui instead of ui kit

Kharchevskyi:
Initially ui was built by just adding uiviews instead of table view usage, which I think less flexible and reusable.

Also ui will be much slower and less smooth on uikit, due to main thread only usage. And which is much more important all size calculations for ui elements (for example cell height based on text) should be done manually. This would lead to completely reworking compose flow or any other flow with dynamic height

tomholub:
Of course we should not be converting from Texture to UIKit, that would be a downgrade.

What would this view look like when implemented in SwiftUI? Could you take a look?

Kharchevskyi:
This is just one example. For full understanding try for example to rewrite To UIKit - ComposeViewController where not only text cell should calculate size but also recipients part.

Kharchevskyi:
Sure. Should I try Swift ui in this pr or may I open another issue for this?

tomholub:
Actually, now is not the best time. In the future, we'll look at SwiftUI separately. So we could proceed with this PR as is.

FlowCrypt/Controllers/Bootstrap/InvalidStorageViewController.swift
```diff
@@ -6,9 +6,18 @@
 //  Copyright © 2017-present FlowCrypt a. s. All rights reserved.
 //
 
-import UIKit
+import FlowCryptUI
+import AsyncDisplayKit
+
+@MainActor
```
ivan-ushakov, line 12: `TableNodeViewController` already has `@MainActor` annotation, inheritance should work here.

FlowCrypt/Controllers/Bootstrap/InvalidStorageViewController.swift
```diff
@@ -85,3 +53,73 @@ final class InvalidStorageViewController: UIViewController {
...
+                                .regular(16),
+                                color: .mainTextColor,
+                                alignment: .center
+                            ),
+                        insets: insets,
+                        backgroundColor: .backgroundColor
+                    )
+                )
+            case .description:
+                return SetupTitleNode(
```
ivan-ushakov, line 97: Looks like if we have very long stack trace text here, we will have very long node (big height) and button will be hidden from user. User have to scroll down to see button. In my original version button is always on screen, independent of stack trace node height.

FlowCrypt/Functionality/Services/GlobalRouter.swift
```diff
@@ -123,6 +123,7 @@ extension GlobalRouter: GlobalRouterType {
         proceed(with: session)
     }
 
+    @MainActor
```
ivan-ushakov, line 126: We could go to this method only from `proceed` and it is already marked with `@MainActor`. Are you sure we need this one?

FlowCrypt/Controllers/Bootstrap/InvalidStorageViewController.swift
Kharchevskyi: Can keep this behaviour, but should we? 
Apple Human Interface Guidelines recommends to use table views for scrollable layout. Also it's more common for iOS users to scroll if content is higher than screen size. And it's quite rare to have such long localized description

FlowCrypt/Functionality/Services/GlobalRouter.swift
Kharchevskyi: I do have compiler error if this method is not marked as `@MainActor`

```
Call to main actor-isolated initializer 'init(error:encryptedStorage:router:)' in a synchronous nonisolated context
```

FlowCrypt/Controllers/Bootstrap/InvalidStorageViewController.swift
tomholub: The exception that caused us to put this in place was indeed quite long, causing a long scroll. 

I guess it's fine if the button ends up on the bottom, where user has to scroll. It's more user friendly when it's on top, but it's not too bad on a screen that should be rarely, if ever, seen.

ivan-ushakov (approved):

PR #1126, opened 2021-11-30
Show sending progress on ui

### LINKED ISSUE #1096: improve sending progress status to debug customer issue
We have a customer which is sending an email without attachments and it takes 70 seconds to send. Then it succeeds.

I'm not sure where to start debugging, it could be their environment. But something we could do to make progress more obvious:

1) differentiate different states in the loader. For example, `Encrypting`, `Sending`, `Deleting draft` or whatever it may be.
2) show a percentage even though we are not uploading any attachments. Is this possible and meaningful? Let's say the request has 1kb of data, maybe we could still show a percentage progress.

I received a video of the issue and it's really just a simple message and then send and then a spinner for 70 seconds like this:

image

I'll file another issue.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/695#issuecomment-979428783_

Kharchevskyi:
This PR update spinner during composing message

close #1096
https://user-images.githubusercontent.com/13693255/144126803-0837914c-6c6b-42c2-a232-867b6e6b1766.mp4

### To be filled by reviewers


FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -62,33 +64,41 @@ final class ComposeMessageService {
         includeAttachments: Bool = true,
         signingPrv: PrvKeyInfo?
     ) async throws -> SendableMsg {
+        state = .validatingMessage
+
         let recipients = contextToSend.recipients
         guard recipients.isNotEmpty else {
+            state = .error
```
tomholub, line 71: error state should be set once at the error handler, not each time where error is thrown

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -164,6 +164,22 @@ final class ComposeViewController: TableNodeViewController {
...
+                    return
+                }
+                self?.updateSpinner(
+                    label: message,
+                    progress: state.progress,
+                    systemImageName: "checkmark.circle"
+                )
+            }
+            .store(in: &lifeTimeCancellable)
+    }
```
tomholub, line 182: Instead of an observable, could you please use a regular callback? A bit closer to how it was before. I find it simpler and easier to work with.

tomholub (commented):
Video looked good. See code comments

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
Kharchevskyi: This is indicates state for service, not for clients code.

FlowCrypt/Controllers/Compose/ComposeViewController.swift
Kharchevskyi: Use same approach as for search functionality and photosManager inside same controller. 
ok, no problems. will rework to use callbacks instead

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
Kharchevskyi: Will get rid of it with "callbacks" rework

tomholub: It does now, but it doesn't have to be implemented that way. It doesn't have to be implemented as a state, instead a callback can be passed around to update progress element.

In fact, since errors are all handled with a modal, there doesn't need to be any progress element update when errors happen.

tomholub:
Roma - you are better suited to evaluate it.

sosnovsky:
I see a several possible improvements:
- app shows `Composing` label and then `Composing XX%`, probably second one should be replaced with `Uploading XX%`
- also `Composing` can be not clear for users, as they've already composed message, it'll be better to say `Encrypting` instead
- when message is sent users see `Sent` label with rotating spinner, which looks wrong - why it's rotating if message sending is completed. Previously we had `tick` icon when message is sent

Simulator Screen Shot - iPhone 12 - 2021-12-01 at 23 21 13

Kharchevskyi:
@sosnovsky fixed according to suggestions. Thanks
Also moved logic of showing hud based on the state directly to controller instead of update hud function

sosnovsky:
Looks good now, the only issue I've noticed is that 'Uploading 100%` stays for several seconds and then `Sent` hud isn't visible, because it appears only when back transition starts, here is video:

https://user-images.githubusercontent.com/1062093/144670340-1e8fc443-0616-48e1-91fd-d2df1cc873d2.mp4

Kharchevskyi:
That is due to gmail api. 
It sends progress 100 but actually message sends few milliseconds later. 
I can try to improve ux for this. But this can be done either by postponing sending "sent" state from service or by delaying animation on conrtoller. Do you think it worth it? @tomholub

sosnovsky:
I think previously `Sent` message was shown when progress reached 100%, and then back animation was performed when app receives `successfully sent` callback from Gmail.
So it was like a small fix/hack for not showing `Uploading 100%` for several seconds and showing `Sent` instead.

Kharchevskyi:
In this case "sent" is showing sometimes for few seconds

sosnovsky:
But it seems more user-friendly than showing `Uploading 100%` for several seconds - for users it can look like something is wrong, everything is uploaded but nothing happens.

Kharchevskyi:
Would suggest either to keep this implementation or postpone hiding sent hud for some predefined duration 0.3 sec

Kharchevskyi:
Same with sent) a bit confusing. Message is sent. But nothing is happening for few seconds. Sometimes it takes up to 3 seconds. Which i would say looks like a bug

sosnovsky:
Okay, let's leave it for now, as inbox screen already shows 'Encrypted message sent' popup after message is sent.

sosnovsky (approved):

tomholub:
I agree with Roma here. I'll file another issue for him to improve it.

tomholub:
Roma has good intuition when it comes to UX as well as code structure. When he raises a UX concern or code concern, we should listen.

PR #1127, opened 2021-11-30
Add accessibility id to navigation title

### LINKED ISSUE #1094: Add accessibility ID for headers
Looks good - less duplication.

When you encounter any `ios class chain` kind of selector, use accessibility identifiers instead.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1092#pullrequestreview-816005153_

Kharchevskyi:
This PR adds accessibility id to navigation title

close #1094

**Tests**:

### To be filled by reviewers


Kharchevskyi:
@fcvakintos can you please take a look if this solution works for you?

FlowCrypt/Controllers/Inbox/InboxViewController.swift
```diff
@@ -92,6 +92,7 @@ final class InboxViewController: ASDKViewController<ASDisplayNode> {
 extension InboxViewController {
     private func setupUI() {
         title = inboxTitle
+        navigationItem.setAccessibility(id: inboxTitle)
```
tomholub, line 95: Let's do `id: "navigationItem \(inboxTitle)"` to distinguish it from just the title

fcvakintos:
@Kharchevskyi it doesn't work for me.
no accessibility Id for headers
Screenshot 2021-12-01 at 10 07 37

Kharchevskyi:
Got it. Will then try another solution. 
Will keep you in touch

Kharchevskyi:
Screenshot 2021-12-01 at 10 49 46
@fcvakintos Please check if this what you are looking for. 
If yes, I will update PR with normal UI

fcvakintos:
@Kharchevskyi doesnt work for me
Screenshot 2021-12-01 at 11 07 05

FlowCrypt/Controllers/Settings/Backup/Backups Scene/BackupViewController.swift
```diff
@@ -10,7 +10,7 @@ import AsyncDisplayKit
 import FlowCryptUI
 
 @MainActor
```
ivan-ushakov, line 12: You can remove `@MainActor` because `TableNodeViewController` has it.

Kharchevskyi:
@fcvakintos  please check latest commit.

fcvakintos:
@Kharchevskyi doesnt work

Kharchevskyi:
let's schedule a meeting and try to find where the problem is
https://github.com/FlowCrypt/org - you can fin email ther

tomholub:
@fcvakintos could you also please demonstrate how can we inspect the layout the way you are inspecting it?

fcvakintos:
@tomholub for that I am using https://github.com/appium/appium-inspector

Kharchevskyi:
I can try to setup it and try locally, but I would need some time, because I'm not familiar with Appium

tomholub:
Please try, and document the steps so that another person not familiar with appium can do the same. Ask @fcvakintos if you get stuck.

That way we'll all be able to debug these things directly.

Kharchevskyi:
@fcvakintos can you please share your config for https://github.com/appium/appium-inspector
Screenshot 2021-12-03 at 21 59 25

fcvakintos:
@Kharchevskyi please replace path to app

{
  "platformName": "iOS",
  "iosInstallPause": 5000,
  "deviceName": "iPhone 13",
  "app": "path to app",
  "platformVersion": "15",
  "automationName": "XCUITest",
  "newCommandTimeout": 10000,
  "wdaLaunchTimeout": 300000,
  "wdaConnectionTimeout": 600000,
  "wdaStartupRetries": 4,
  "wdaStartupRetryInterval": 120000
}

tomholub:
The inspector should be added to `package.json` so it gets auto-installed. That means fewer steps for someone who wants to use it.

fcvakintos:
we don't need it in package.json

tomholub:
I didn't say we need it. I said that if we add it there, then other developers who need to use it don't have to install it separately. Isn't that true?

fcvakintos:
I am using mac app installation, not npm

tomholub:
Understood - I assumed it could be installed from npm, which is not the case.

Kharchevskyi:
@fcvakintos would this work for you?
Screenshot 2021-12-04 at 18 26 40
or

fcvakintos:
@Kharchevskyi you don't have accessibility Id for header now
check screen, here is accessibility id

Screenshot 2021-12-06 at 10 22 51

Kharchevskyi:
@fcvakintos please confirm this is what you are looking for
Screenshot 2021-12-06 at 13 36 01

fcvakintos:
@Kharchevskyi  we can use "choosePassPhrase"  - it is better i think

Kharchevskyi:
according to suggestion 
https://github.com/FlowCrypt/flowcrypt-ios/pull/1127#discussion_r759719723

fcvakintos:
@Kharchevskyi yes, but it is not correct, i think it should be 1 word, can we change it?

tomholub:
@fcvakintos if it should be one word, then I suggest `navigationItemChoosePassPhrase` which would be done dynamically something like `navigationItem\(title.replace(" ", ""))`. @Kharchevskyi could you do that?

tomholub:
@Kharchevskyi later please add to appium/README.md how to install and use the above tool and how to configure it so that others can also do it. I'm filing a separate issue.

Kharchevskyi:
Yes, sure. Will make it as one word - `navigationItemChoosePassPhrase` and so on

.gitignore
```diff
@@ -50,4 +50,5 @@ appium/tmp
 appium/.env
 appium/FlowCrypt.app
 appium/video
+appium/.env
```
tomholub, line 53: Duplicate, see line 50

FlowCryptUI/Nodes/TableViewController.swift
```diff
@@ -10,9 +10,37 @@ import AsyncDisplayKit
...
+    }
     
     public override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
         super.traitCollectionDidChange(previousTraitCollection)
         node.reloadData()
     }
 }
+
+public extension UINavigationItem {
+    func setAccessibility(id: String?) {
```
tomholub, line 26: Can this be non-optional? It's better to require it then to default to empty string.

Kharchevskyi: In current implementation we observe `didSet` event for viewcontroller title which can be nil. 
https://developer.apple.com/documentation/uikit/uiviewcontroller/1621364-title

If we want to make it non nil then we need to add this call to each viewcontroller implicitly

Kharchevskyi:
```
public override var title: String? {
        didSet {
            navigationItem.setAccessibility(id: title)
        }
    }
```

tomholub: I don't fully understand the technicalities, but: every view has a title. Therefore every view could also have a title-based accessibilityId. Therefore there shouldn't be a need for optionality.

Kharchevskyi: Not every view has a title only where we set it implicitly.

tomholub: You meant explicitly. Ok, in that case it's fine.

Kharchevskyi: sorry, explicitly.

tomholub (approved):
I assume this is ready to merge

tomholub:
@Kharchevskyi or not yet?

tomholub:
I'll merge it either way, if there is work left to do, can do another PR.

fcvakintos:
@Kharchevskyi @tomholub I found new issue after this fix, now we dont have accesibility id for menu bar items
it is after this fix i think, because now accessibility id = "INBOX" is not unique
Screenshot 2021-12-07 at 10 26 21

can we add accessibility id for menu bar items?
something like this - menubarIteminbox, menubaritemchat and etc

Kharchevskyi:
Can't be related to this changes, but will do asap

PR #1133, opened 2021-12-01
Use Realm queries, don't re-implement database code

### LINKED ISSUE #1114: use Realm queries, don't re-implement database code
This kind of code is all over the storage classes, this is just one example.

```swift
    func doesAnyKeyExist(for email: String) -> Bool {
        keysInfo()
            .map(\.account)
            .map { $0 == email }
            .contains(true)
    }
```

here's what's happening:
1) it retrieves all keys of all users
2) it maps only accounts from them
3) it maps it to email comparison
4) checks if any is true

This is not how a database should be used. It should be:

```swift
    func doesAnyKeyExist(for email: String) -> Bool {
        return storage.objects(KeyInfoRealmObject.self).where { $0.account == email }.count > 0
    }
```

Let the database do the work.

### LINKED ISSUE #1147: PR comments to clean up storage methods
Changes look good. I'm adding two more comments to update (in another PR).

ivan-ushakov:
close #1114
close #1147

### To be filled by reviewers


ivan-ushakov:
I removed `KeyDataStorageType` and `KeyDataStorage` types and moved logic to `EncryptedStorage`. So we consolidate all DB logic in one class.

FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
```diff
@@ -235,8 +235,18 @@ extension SetupManuallyEnterPassPhraseViewController {
...
         let newKeysToAdd = Array(Set(fetchedKeys).subtracting(existedKeys))
 
-        appContext.keyStorage.addKeys(keyDetails: newKeysToAdd, passPhrase: passPhrase, source: .imported, for: email)
-        appContext.keyStorage.updateKeys(keyDetails: keysToUpdate, passPhrase: passPhrase, source: .imported, for: email)
+        appContext.encryptedStorage.addKeys(
+            keyDetails: newKeysToAdd,
+            passPhrase: passPhrase,
+            source: .imported,
+            for: email
+        )
```
tomholub, line 243: Since you switched to use encrypted storage directly, can you remove some unused methods from keyStorage?

sosnovsky (approved):
Looks good 👍

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -181,34 +187,43 @@ extension EncryptedStorage {
...
+        return storage.objects(KeyInfoRealmObject.self).where({
+            $0.account == email
+        }).map(KeyInfo.init)
+    }
+
+    func findPrivateKey(by email: String) -> [String] {
+        return storage.objects(KeyInfoRealmObject.self).where({
+            $0.account == email
+        }).map(\.private)
     }
```
tomholub, line 209: This one is the exact one as above, except the map - duplicates functionality.

We should remove `findKeyInfo` to `getPrivateKeys(for email: String)` and then remove `findPrivateKey`.

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -181,34 +187,43 @@ extension EncryptedStorage {
...
 
     func publicKey() -> String? {
         storage.objects(KeyInfoRealmObject.self)
-            .map(\.public)
-            .first
+            .first.map(\.public)
     }
```
tomholub, line 214: This method is plain wrong, it picks public key regardless of which account it belongs to.

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -181,34 +187,43 @@ extension EncryptedStorage {
...
     func doesAnyKeyExist(for email: String) -> Bool {
-        keysInfo()
-            .map(\.account)
-            .map { $0 == email }
-            .contains(true)
+        let keys = storage.objects(KeyInfoRealmObject.self).where {
+            $0.account == email
+        }
+        return !keys.isEmpty
     }
```
tomholub, line 221: For clarification let's name this `doesAnyPrivateKeyExists`

tomholub (changes requested):
See comments below, while we're working with this code.

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
tomholub: I'd remove it. Wherever this is used, use `getKeypairs(for email: String).map(\.public).first`

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
ivan-ushakov: When we do `.map(\.private)` Realm reads only this property from persistent storage, but `findKeyInfo` returns array of `KeyInfo` models with all properties.

tomholub: Got it - now I understood the motivation. Still, I think the optimization is not worth another method. I'd rather less code and more clarity, especially since the overhead here is very small.

FlowCrypt/Functionality/DataManager/DataService.swift
```diff
@@ -111,7 +111,7 @@ extension DataService: DataServiceType {
 
     func validAccounts() -> [User] {
```
tomholub, line 112: `validAccounts` sould be called `getFinishedSetupUsersExceptCurrentUser` based on what it does now. Naming the methods by what they actually do sometimes highlights the ridiculousness of their logic.

More properly speaking, it should be renamed as `getFinishedSetupUsers` and it should not filter by `currentEmail`. The callers of that method should do the filtering, if that's what they need. If this method is used in many places that all need the filtering, then the method should be changed to `getFinishedSetupUsers(exceptUserEmail: String)` for clarity.

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -14,16 +14,14 @@ import UIKit
...
 
+    var activeUser: User? { get }
     func getAllUsers() -> [User]
     func saveActiveUser(with user: User)
-    var activeUser: User? { get }
-    func doesAnyKeyExist(for email: String) -> Bool
+    func doesAnyKeypairExist(for email: String) -> Bool
 
     func addKeys(keyDetails: [KeyDetails], passPhrase: String?, source: KeySource, for email: String)
     func updateKeys(keyDetails: [KeyDetails], passPhrase: String?, source: KeySource, for email: String)
```
tomholub: Let's call it `addKeypairs` and `updateKeypairs`, but really, is there any difference between them? Whether I'm adding or updating, I want the right keypair to be there in the end. I would refactor these two as a single method: `putKeypairs`, which adds or updates each keypair depending if we already have it or not (for that user email).

tomholub (approved):
Changes look good. I'm adding two more comments to update (in another PR).

Also in another PR: how about this comment from before? There are surely some methods to remove that are unused now?

image

FlowCrypt/Models/Realm Models/ClientConfigurationRealmObject.swift
```diff
@@ -10,7 +10,7 @@ import Foundation
 import RealmSwift
 
 final class ClientConfigurationRealmObject: Object {
-    @Persisted(primaryKey: true) var userEmail: String!
```
ivan-ushakov, line 13: Have to revert this back because it requires migration

FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
ivan-ushakov: Since I removed `KeyStorageType` and implementation and mentioned this in comment above, I was thinking that maybe you overlooked this. Sorry, will ask you next time.

FlowCrypt/Models/Realm Models/ClientConfigurationRealmObject.swift
tomholub: We should ideally have a test for upgrading from earlier app. For this, I suppose we would pre-build an app and check it into the repo. It would forever stay the same until it can no longer run the current iOS version we test on.

The test would:
 - open the app
 - set up an account
 - open the most recent app build
 - verify that basic functionality remains functional (eg no missed or bad migrations)

Not sure if realistic - worth a try.

FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
tomholub: Ah! Very good, I indeed overlooked it because the implementation is called `KeyDataStorage` ... and I was looking for `KeyStorage`. All good!

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
ivan-ushakov: Yes, logic looks the same except that `addKeys` has `UpdatePolicy.error`, and `updateKeys` has `UpdatePolicy.all` update policy. We could leave only `updateKeys` if we are find with such policy.

tomholub: Let's leave only `updateKeys` but call it `upsertKeypairs`

FlowCrypt/Models/Realm Models/ClientConfigurationRealmObject.swift
tomholub: Do you still have to update it in this PR? If yes, then mark it as a draft so it wouldn't merge in the meanwhile.

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
ivan-ushakov: Let's stay on `putKeypairs` if you don't mind.

FlowCrypt/Functionality/DataManager/DataService.swift
ivan-ushakov: We use this in one place - app menu, when we show list of possible accounts to switch to.

FlowCrypt/Models/Realm Models/ClientConfigurationRealmObject.swift
ivan-ushakov: Yes I have to revert Realm schema change. I did also that two requests from new issue.

FlowCrypt/Controllers/SideMenu/Menu/MyMenuViewController.swift
```diff
@@ -50,7 +50,9 @@ final class MyMenuViewController: ASDKViewController<ASDisplayNode> {
 
     private var folders: [FolderViewModel] = []
     private var serviceItems: [FolderViewModel] { FolderViewModel.menuItems }
-    private var accounts: [User] { appContext.dataService.validAccounts() }
+    private var accounts: [User] {
+        appContext.dataService.getFinishedSetupUsers(exceptUserEmail: currentUser.email)
```
ivan-ushakov, line 54: Here we use that filter.

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -160,19 +159,7 @@ extension EncryptedStorage {
...
-        try! storage.write {
-            for key in keyDetails {
-                storage.add(try! KeyInfoRealmObject(key, passphrase: passPhrase, source: source, user: user))
-            }
-        }
-    }
-
-    func updateKeys(keyDetails: [KeyDetails], passPhrase: String?, source: KeySource, for email: String) {
+    func putKeypairs(keyDetails: [KeyDetails], passPhrase: String?, source: KeySource, for email: String) {
         guard let user = getUserObject(for: email) else {
```
tomholub, line 163: I've noticed yet another issue, but that is more complicated and should be in another issue (requires migration). The private keys should have a unique index on fingerprint+email, not just fingerprint. Because if it's just fingerprint, I can't use the same key for two accounts.

tomholub (approved):

FlowCrypt/Controllers/SideMenu/Menu/MyMenuViewController.swift
tomholub: Should be called `otherAccounts` :-)

tomholub: And if it's only used here, then we should just do the filtering here instead of an argument

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
ivan-ushakov: Realm has following limitation:
> You can define only one primary key per object schema.

So we need to think how to get that index with fingerprint and email.

tomholub: probably a primary id that has both email and fingerprint in the string.

ivan-ushakov:
This one works for me locally - `sending message to user without public key produces modal`, so I will trigger tests again later.

tomholub:
> This one works for me locally - `sending message to user without public key produces modal`, so I will trigger tests again later.

When you notice a flaky test like that (passes one time but not other time) please report it with an issue and assign @fcvakintos . I've just reported it here: https://github.com/FlowCrypt/flowcrypt-ios/issues/1153

Issue #414, opened 2021-07-28
decide based on consumer / enterprise build when trying FES


tomholub:
-> did I NOT receive a response (server down or timeout) and the internet is otherwise working? -> don't use FES, **empty OrgRules**

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/408#issuecomment-888426548_

In the future, this should be decided based on if this is consumer or enterprise build.

tomholub:
Here is the full diagram of expected behavior:

![image](https://user-images.githubusercontent.com/6306961/127357555-14cba36b-39fb-4d7e-b44c-dc000ae24aed.png)

tomholub:
Ignoring the more complex part of this issue, in Swift how does one do an if-else statement that uses the build type (or bundle id) for decisions?

sosnovsky:
We can use something like this
```swift
enum BundleType: String {
    case debug = "com.flowcrypt.as.ios.debug",
         enterprise = "com.flowcrypt.as.ios.enterprise",
         release = "com.flowcrypt.as.ios.consumer"
}

var bundleType: BundleType {
    guard let bundleIdentifier = Bundle.main.bundleIdentifier else { return .debug }
    return BundleType(rawValue: bundleIdentifier) ?? .debug
}
```

tomholub:
Ok - that helps :+1:

tomholub:
I just saw this in one of appium tests. My guess is that internet was spotty and so call to FES failed, so it went with default config. This highlights the need to not allow this on enterprise builds, where FES is expected to be running.

<img width="217" alt="image" src="https://user-images.githubusercontent.com/6306961/144727494-f523cfa7-bc1a-4de5-bdf8-91d1c8aee2ad.png">

PR #1152, opened 2021-12-02
A server with the specified hostname could not be found

### LINKED ISSUE #1139: A server with the specified hostname could not be found
Please see where exactly this error is coming from, and add the server URL to the error message so that we know which one it is.

image

ivan-ushakov:
close #1139

- Not worth testing

### To be filled by reviewers


FlowCrypt/Functionality/Services/Account Server Services/EnterpriseServerApi.swift
```diff
@@ -79,8 +79,8 @@ class EnterpriseServerApi: EnterpriseServerApiType {
 
             return urlString
         } catch {
-            if let httpError = error as? HttpErr,
-               let nsError = httpError.error as NSError?,
+            if let apiError = error as? ApiError,
```
ivan-ushakov, line 82: @tomholub 
Maybe it is a bad idea to check tolerated `NSError` codes at business logic level?

tomholub: Maybe. It's not the worst. Please add a convenience method like `self.shouldTolerateWhenCallingOpportunistically(error: Error)`. Because I'd like to later improve that method with #414 . Enterprise build will not be tolerating missing FES (later) and that convenience method will be a good place to put it.

tomholub (commented):
This will definitely need a test because it's too easy to break. Maybe a unit test would be enough. Like, `expect getActiveFesUrl("user@nonexistentdomain.test") to be nil` (without throwing).

Please see if that test would have uncovered the problem in the original implementation. If yes, add the test. Thanks!

ivan-ushakov:
> This will definitely need a test because it's too easy to break. Maybe a unit test would be enough. Like, `expect getActiveFesUrl("user@nonexistentdomain.test") to be nil` (without throwing).
> 
> Please see if that test would have uncovered the problem in the original implementation. If yes, add the test. Thanks!

Do you want to have integration test here with real network request?

tomholub:
> > This will definitely need a test because it's too easy to break. Maybe a unit test would be enough. Like, `expect getActiveFesUrl("user@nonexistentdomain.test") to be nil` (without throwing).
> > Please see if that test would have uncovered the problem in the original implementation. If yes, add the test. Thanks!
> 
> Do you want to have integration test here with real network request?

That would be best. When I say unit test, I usually mean integration test :-)

tomholub:
I'll bring in my changes, they were also touching this code

tomholub (approved):

PR #1158, opened 2021-12-03
issue 1006 appium app to call FES mock

### LINKED ISSUE #1006: Create a FES mock for UI tests, make the tested app use it
For this, will need a way to mock FES responses. Similar to #711 - need our own FES instance for UI tests and need the app to listen to it. I suppose it would be easier to start with FES since the API is simple and obvious.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/736#issuecomment-968105488_

### LINKED ISSUE #1012: optionally send API requests to a mock URL
I understand we could probably do another build, to have eg enterprise/consumer/debug/mock builds where the mock build would have different URLs to call.

But that would mean we would have to build too apps for UI testing: both debug and mock if we want to test live as well as mocked version (which we do).

Is there a way to build one app but somehow toggle if it should use mock or live APIs? One idea is that the splash screen would have an actual toggle there, but that may make app code too complicated. Is there some way to run an app with parameters to one time tell it to use mock apis and other times not? (something like unix command line arguments)

tomholub:
This PR adds first tests that use FES mock to run. They still need some work, will continue with #1162 

close #1006
close #1012

**Tests**:
- tests extended

### To be filled by reviewers


appium/README.md
```diff
@@ -10,12 +10,16 @@
...
-Run this in `appium` folder. 
+## Run tests
 
-`npm test` - build `FlowCrypt.app` and run all ui tests
+Run these in `appium` folder. `live` means real production APIs, `mock` means local mock APIs. 
 
-`npm run-script only.test.all` - run all ui tests without building the `.app`. Use this if you already built the `.app` before, and now only want to change the UI test spec without rebuilding the app
+To run a particular test:
+- `npm run-script test.live "user is able to view text email"`
+- `npm run-script test.mock "app setup fails with bad EKM URL"`
```
tomholub, line 23: @sosnovsky @fcvakintos @ivan-ushakov @Kharchevskyi @ekievsky I have updated the commands to run tests.

So instead of `npm run-script only.test.filter "some test"`
it now is **`npn run-script test.live "some test"`**.

appium/tests/specs/live/login/GmailLogin.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+  SplashScreen,
+  SetupKeyScreen,
+  MenuBarScreen,
+  MailFolderScreen
+} from '../../../screenobjects/all-screens';
+
+
+describe('LOGIN: ', () => {
+
+  it('user is able to login via gmail', async () => {
```
tomholub, line 11: We could remove this spec, since now it's covered in many others

appium/tests/specs/mock/setup/SetupFailsWithInconsistentClientConfiguration.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+describe('SETUP: ', () => {
+
+  it('app setup fails with bad EKM URL', async () => {
+    const mockApi = new MockApi();
+    mockApi.fesConfig = { clientConfiguration: { key_manager_url: 'INTENTIONAL BAD URL' } };
+    await mockApi.withMockedApis(async () => {
+      await SplashScreen.login();
+      // todo - replace the following pause with wait for modal error
+      //   that says "Please check if key manager url set correctly"
+      await browser.pause(5000);
```
tomholub, line 16: For now this passes blindly without checking result, will be finished in #1162

appium/tests/specs/mock/setup/SetupShowsMeaningfulErrorFromFES.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+
+describe('SETUP: ', () => {
+
+  it('setup shows meaningful error when FES returns 400', async () => {
+    const mockApi = new MockApi();
+    mockApi.fesConfig = { returnError: { code: 400, message: "some client err" } };
+    await mockApi.withMockedApis(async () => {
+      await SplashScreen.login();
+      // todo - replace the following pause with wait for modal error
+      await browser.pause(5000);
```
tomholub, line 15: Here too #1162

PR #1165, opened 2021-12-04
Use throwing var methods for realm references to other objects

### LINKED ISSUE #1035: use throwing var methods for realm references to other objects
> @tomholub ability to throw from `var` variable was added to the Swift recently:
> 
> ```swift
> var activeUser: UserRealmObject {
>     get throws {
>         guard let user = user else { throw AppErr... }
>         return user
>     }
>   }
> ```

This looks good. Probably for another issue/PR?

The advantage is that an exception can be captured and handled (in the general error handling somewhere in VC, wherever user action was initiated) while a forced unwrap cannot.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1030#issuecomment-972716809_

### LINKED ISSUE #1174: global router -> switch active user: error handling
I'll file an issue for this

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1165#discussion_r762540652_

ivan-ushakov:
close #1035
close #1174

### To be filled by reviewers


FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
```diff
@@ -222,7 +222,7 @@ extension ThreadDetailsViewController {
                 input[index].rawMessage = updatedMessage
                 node.reloadSections(IndexSet(integer: index), with: .fade)
             } catch {
-                showToast("Could not mark message as read: \(error)")
+                showToast("Could not mark message as read: \(error.localizedDescription)")
```
ivan-ushakov, line 225: Localisation?

FlowCrypt/Functionality/Services/Folders Services/FoldersService.swift
```diff
@@ -53,18 +53,22 @@ final class FoldersService: FoldersServiceType {
         let fetchedFolders = try await self.remoteFoldersProvider.fetchFolders()
         return try await withCheckedThrowingContinuation { continuation in
             DispatchQueue.main.async {
-                // TODO: - Ticket? - instead of removing all folders remove only
-                // those folders which are in DB and not in remoteFolders
-                self.localFoldersProvider.removeFolders(for: user.email)
+                do {
+                    // TODO: - Ticket? - instead of removing all folders remove only
+                    // those folders which are in DB and not in remoteFolders
+                    try self.localFoldersProvider.removeFolders(for: user.email)
```
ivan-ushakov, line 59: @tomholub
Maybe you need this transactional? Now this are independent DB changes.

FlowCrypt/Models/Realm Models/KeyInfoRealmObject.swift
```diff
@@ -70,7 +70,8 @@ extension KeyInfoRealmObject {
 
 extension KeyInfoRealmObject {
     /// associated user email
-    var account: String? {
-        user?.email
+    var account: String {
+        guard let email = user?.email else { fatalError() }
```
ivan-ushakov, line 74: I'm not sure it is good to use `throws` here because having such error already makes app wrong. So it is better to guaranty that we always have `user` property when we create/update object.

ivan-ushakov:
Looks like this PR is more about removing `EncryptedCacheService` and changing `try?` to `try`.

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -22,6 +22,8 @@ protocol EncryptedStorageType {
     func putKeypairs(keyDetails: [KeyDetails], passPhrase: String?, source: KeySource, for email: String)
     func getKeypairs(by email: String) -> [KeyInfo]
 
+    func find(with email: String) -> Recipient?
```
ivan-ushakov, line 25: Maybe `getRecipient`?

FlowCrypt/Functionality/Services/Local Pub Key Services/LocalContactsProvider.swift
```diff
@@ -8,66 +8,73 @@
...
 
 extension LocalContactsProvider: LocalContactsProviderType {
-    func updateLastUsedDate(for email: String) {
-        let recipient = find(with: email)
+    func retrievePubKeys(for email: String) -> [String] {
+        guard let object = find(with: email) else { return [] }
 
-        try? localContactsCache.realm.write {
-            recipient?.lastUsed = Date()
+        do {
```
ivan-ushakov, line 41: I don't throw here because problem with updating timestamp does not mean we don't have public keys.

FlowCrypt/Functionality/Services/Folders Services/FoldersService.swift
tomholub: Maybe, you can use your judgement

FlowCrypt/Models/Realm Models/KeyInfoRealmObject.swift
tomholub: Sure, assuming our own code in the future will never be buggy. Would you bet your neck on that? I wouldn't :-)

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
tomholub: agree

FlowCrypt/Functionality/Services/Local Pub Key Services/LocalContactsProvider.swift
tomholub: I'd throw either way - it's a good habit.

FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
tomholub: can be spelled both ways

ivan-ushakov: I mean do we need to move this to our localisation file?

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
ivan-ushakov: This one not needed anymore

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -10,16 +10,15 @@ import FlowCryptCommon
 import RealmSwift
 import UIKit
 
-// swiftlint:disable force_try
```
ivan-ushakov, line 13: Now all methods with `try` are throwing so we don't need this hint anymore

FlowCrypt/Resources/en.lproj/Localizable.strings
```diff
@@ -160,8 +160,11 @@
 
 // Contacts
 "contacts_screen_title" = "Contacts";
+"contacs_screen_load_error" = "Failed to load recipients: %@";
```
ivan-ushakov, line 163: `contacts`

FlowCrypt/Controllers/Bootstrap/FatalErrorViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+//  FlowCrypt
+//
+//  Created by  Ivan Ushakov on 05.12.2021
+//  Copyright © 2017-present FlowCrypt a. s. All rights reserved.
+//
+
+import FlowCryptUI
+import AsyncDisplayKit
+
+final class FatalErrorViewController: TableNodeViewController {
```
ivan-ushakov, line 12: This one is for cases when it is not possible to continue using app and you have to restart it

FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
tomholub: I see. Sure - that never hurts.

FlowCrypt/Controllers/Settings/Contacts/Contacts List/ContactDetailViewController.swift
```diff
@@ -91,7 +91,11 @@ extension ContactDetailViewController {
 
         recipient.remove(pubKey: keyToRemove)
         if let fingerprint = keyToRemove.fingerprint, fingerprint.isNotEmpty {
-            contactsProvider.removePubKey(with: fingerprint, for: recipient.email)
+            do {
+                try contactsProvider.removePubKey(with: fingerprint, for: recipient.email)
+            } catch {
+                showToast("contact_detail_remove_public_key_error".localizeWithArguments(error.localizedDescription))
+            }
         }
```
tomholub, line 99: This should be made a transaction in storage (can be another PR later) else it produces inconsistent results.

FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
```diff
@@ -228,41 +228,41 @@ extension SetupManuallyEnterPassPhraseViewController {
...
-            newKeysToAdd
+            try newKeysToAdd
                 .map {
                     PassPhrase(value: passPhrase, fingerprintsOfAssociatedKey: $0.fingerprints)
                 }
                 .forEach {
-                    appContext.passPhraseService.savePassPhrase(with: $0, storageMethod: storageMethod)
+                    try appContext.passPhraseService.savePassPhrase(with: $0, storageMethod: storageMethod)
                 }
         }
```
tomholub, line 267: I'd like to ban `forEach` in this codebase. Let's instead use`for thing in things {`, in this case `for newKeyToAdd in newKeysToAdd {`. I'll make another issue, to do this across the codebase.

FlowCrypt/Functionality/Services/Folders Services/LocalFoldersProvider.swift
```diff
@@ -11,27 +11,45 @@ import RealmSwift
...
     }
+}
 
+extension LocalFoldersProvider: LocalFoldersProviderType {
     func fetchFolders(for userEmail: String) -> [FolderViewModel] {
-        return folderCache.getAll(for: userEmail).compactMap(FolderViewModel.init)
+        storage.objects(FolderRealmObject.self).where {
+            $0.user.email == userEmail
+        }.compactMap(FolderViewModel.init)
     }
```
tomholub, line 35: I prefer to see explicit returns - maybe there is some linter for this? That would be nice.

FlowCrypt/Functionality/Services/GlobalRouter.swift
```diff
@@ -123,11 +126,15 @@ extension GlobalRouter: GlobalRouterType {
...
-            return
+        do {
+            guard let session = try appContext.userAccountService.switchActiveSessionFor(user: user) else {
+                logger.logWarning("Can't switch active user with \(user.email)")
+                return
+            }
+            proceed(with: appContext.withSession(session))
+        } catch {
+            // TODO
         }
```
tomholub, line 137: I'll file an issue for this

FlowCrypt/Functionality/Services/Local Pub Key Services/ContactsService.swift
```diff
@@ -62,12 +62,10 @@ extension ContactsService: ContactsProviderType {
...
-        return publicKeys
+        return localContactsProvider.retrievePubKeys(for: email)
     }
 
-    func removePubKey(with fingerprint: String, for email: String) {
-        localContactsProvider.removePubKey(with: fingerprint, for: email)
+    func removePubKey(with fingerprint: String, for email: String) throws {
+        try localContactsProvider.removePubKey(with: fingerprint, for: email)
     }
 }
```
tomholub, line 71: Useless methods, Let's delete then, and use `localContactsProvider` directly

FlowCrypt/Functionality/Services/Local Pub Key Services/ContactsService.swift
```diff
@@ -46,12 +46,12 @@ extension ContactsService: ContactsProviderType {
...
-            localContactsProvider.save(recipient: recipient)
+            try localContactsProvider.save(recipient: recipient)
             return recipient
         }
 
         let recipient = try await pubLookup.lookup(email: email)
-        localContactsProvider.updateKeys(for: recipient)
+        try localContactsProvider.updateKeys(for: recipient)
         return contact
     }
```
tomholub, line 56: We could find a new home for this method, and then delete `ContactsService`.

FlowCrypt/Functionality/Services/Local Pub Key Services/LocalContactsProvider.swift
```diff
@@ -8,66 +8,77 @@
...
+            try storage.write {
+                object.lastUsed = Date()
+            }
```
tomholub, line 48: For another issue - need to review usage. Just because a public key is retrieved doesn't mean that we should update lastUsed. lastUsed should only be updated when user explicitly uses the recipient, like writing a message to them.

FlowCrypt/Models/Realm Models/KeyInfoRealmObject.swift
tomholub: Still prefer throws here rather than `fatalError`.

FlowCrypt/Controllers/Bootstrap/FatalErrorViewController.swift
tomholub: This is very similar to the other error screen we have. It should be the same screen and same VC, with a boolean init parameter to offer to reset storage or not.

tomholub (approved):

tomholub (changes requested):
I meant "request changes". I'll file a bunch of these issues for later, but some remain.

FlowCrypt/Models/Realm Models/KeyInfoRealmObject.swift
ivan-ushakov: It makes code hard to use. We use this property in select queries, and having all such method throws will make code verbose.

FlowCrypt/Controllers/Bootstrap/FatalErrorViewController.swift
ivan-ushakov: It is strange to have universal controller with all possible error handling logic.

tomholub:
Btw it's a great improvement overall 👍

tomholub:
Please also have a look at:
 - https://github.com/FlowCrypt/flowcrypt-ios/pull/1165#discussion_r762541426
 - https://github.com/FlowCrypt/flowcrypt-ios/pull/1165#discussion_r762541537

FlowCrypt/Models/Realm Models/KeyInfoRealmObject.swift
tomholub: I'd still like to look into this in the future. Will file another issue.

FlowCrypt/Controllers/Bootstrap/FatalErrorViewController.swift
tomholub: It's not strange to me. A "fatal error" kind of page will always look about the same, no matter the underlying problem. Sometimes it will have a button and sometimes not. But they are very similar, aren't they? I'll file another issue.

PR #1166, opened 2021-12-04
Replace Combine Future with async/await if any left

### LINKED ISSUE #1058: replace Combine Future with async/await if any left
@tomholub I think we should use `async/await` here instead of Combine for code consistency across project, and for better threads handling with `@MainActor`, can be done in another issue

_Originally posted by @sosnovsky in https://github.com/FlowCrypt/flowcrypt-ios/pull/1055#discussion_r754064992_

ivan-ushakov:
close #1058

### To be filled by reviewers


ivan-ushakov:
We still have one `import Combine` in `ComposeViewController.swift` because we use it for some UI logic.

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -1060,50 +1077,49 @@ extension ComposeViewController {
...
+        Task {
+            await photosManager.selectPhoto(from: self)
+        }
+    }
+
+    private func selectFromFilesApp() {
+        Task {
+            await filesManager.selectFromFilesApp(from: self)
+        }
+    }
```
tomholub, line 1121: Here in `takePhoto` we should distinguish the exact "missing access" photo before running `showNoAccessToCameraAlert`. Else a regular error modal.

`selectPhoto` and `selectFromFilesApp` should have errors handled with a modal?

Let's look at this in the future.

FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
```diff
@@ -488,7 +488,12 @@ extension ThreadDetailsViewController: ASTableDelegate, ASTableDataSource {
                     msgAttachment: attachment,
                     index: attachmentIndex
                 ),
-                onDownloadTap: { [weak self] in self?.attachmentManager.open(attachment) }
+                onDownloadTap: { [weak self] in
+                    guard let self = self else { return }
+                    Task {
+                        await self.attachmentManager.open(attachment)
+                    }
```
tomholub, line 495: Also here errors should be handled with a modal.

tomholub (approved):

tomholub:
I'm good to merge this, but some tests are unhappy.

FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
ivan-ushakov: Error is handled in `open` method. Do you want to move it here? I guess this was done to avoid code duplication.
```
func open(_ attachment: MessageAttachment) async {
    do {
        try await filesManager.saveToFilesApp(file: attachment, from: self)
    } catch {
        controller?.showToast(
            "message_attachment_saved_with_error".localizeWithArguments(error.localizedDescription)
        )
    }
}
```

FlowCrypt/Controllers/Compose/ComposeViewController.swift
ivan-ushakov: We don't have errors in `selectPhoto` and `selectFromFilesApp`, so I don't use `throws`.

ivan-ushakov:
> I'm good to merge this, but some tests are unhappy.

It works fine locally. We already have issue for this test.

PR #1188, opened 2021-12-07
#1146 Fix multiple keys EKM setup

### LINKED ISSUE #1146: when setting up with keys from EKM and pass phrase in memory, it doesn't save pass phrase of second key
> > for 1st and 4th messages I need to set pass phrase
> 
> Is this right after you've re-set up the e2e account?

I've just tested this. It's a bug. I'll file an issue for this right now. You can write the test so that it gives it the pass phrase, but add a comment that that's because of a bug and should be removed once the github issue is fixed. Link that issue in the comment.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1095#issuecomment-984739170_

sosnovsky:
This PR fixes EKM setup with multiple keys

close #1146

- Tests added or updated - updated `CheckMessageProcessingErrors` test

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Setup/SetupEKMKeyViewController.swift
+++ b/FlowCrypt/Controllers/Setup/SetupEKMKeyViewController.swift
@@ -82,7 +82,7 @@ extension SetupEKMKeyViewController {
     private func setupAccountWithKeysFetchedFromEkm(with passPhrase: String) async throws {
         self.showSpinner()
         try await self.validateAndConfirmNewPassPhraseOrReject(passPhrase: passPhrase)
-        var allFingerprints: [[String]] = []
+        var allFingerprintsOfAllKeys: [[String]] = []
         for keyDetail in self.keys {
             guard let privateKey = keyDetail.private else {
                 throw CreatePassphraseWithExistingKeyError.noPrivateKey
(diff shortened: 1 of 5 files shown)


FlowCrypt/Controllers/Setup/SetupEKMKeyViewController.swift
```diff
@@ -98,11 +98,14 @@ extension SetupEKMKeyViewController {
...
         }
         if self.storageMethod == .memory {
-            let passPhrase = PassPhrase(value: passPhrase, fingerprintsOfAssociatedKey: allFingerprints.unique())
-            try appContext.passPhraseService.savePassPhrase(with: passPhrase, storageMethod: self.storageMethod)
+            try allFingerprints
+                    .map { PassPhrase(value: passPhrase, fingerprintsOfAssociatedKey: $0) }
+                    .forEach {
+                        try appContext.passPhraseService.savePassPhrase(with: $0, storageMethod: storageMethod)
+                    }
         }
```
tomholub, line 109: I'd like to ask that we standardize on a `for thing in things {` loop and always avoid `.forEach {`. It's more direct: you can return from it, you can try/throw/catch without having to try in front of the loop, and why it "looks less functional" I'd say it's more idiomatic. Either way, it's one of those small things I'll have to insist on 🤷 

So in this case, 


```swift
for allFingerprintsOfOneKey in allFingerprintsOfAllKeys {
    try appContext.passPhraseService.savePassPhrase(
        with: PassPhrase(value: passPhrase, fingerprintsOfAssociatedKey: allFingerprintsOfOneKey) },
        storageMethod: storageMethod
    )
}
```

It will be clearer.

tomholub (commented):
Please a style change - else looks good 👍

tomholub (approved):
Thanks!

PR #1191, opened 2021-12-07
Primary key for saving private keys should be email+fingerprint

### LINKED ISSUE #1150: primary key for saving private keys should be email+fingerprint
Realm has following limitation:
> You can define only one primary key per object schema.

So we need to think how to get that index with fingerprint and email.

_Originally posted by @ivan-ushakov in https://github.com/FlowCrypt/flowcrypt-ios/pull/1133#discussion_r761394522_

ivan-ushakov:
close #1150

### To be filled by reviewers


FlowCrypt/Models/Common/Keypair.swift
```diff
@@ -8,7 +8,7 @@
 
 import Foundation
 
-struct KeyInfo {
+struct Keypair {
```
ivan-ushakov, line 11: Let's change this one too for consistency

tomholub: very good

FlowCrypt/Functionality/DataManager/Encrypted Storage/Version5SchemaMigration.swift
```diff
@@ -254,4 +260,38 @@ final class Version5SchemaMigration {
...
+    private func setKeypairObjectPrimaryKey(_ object: MigrationObject) throws {
+        guard
+            let userObject = object[Properties.Keypair.user.rawValue] as? Object,
+            let email = userObject[Properties.User.email.rawValue] as? String,
+            let primaryFingerprint = object[Properties.Keypair.primaryFingerprint.rawValue] as? String
+        else {
+            throw AppErr.unexpected("KeypairRealmObject primary key")
+        }
+
+        object[Properties.Keypair.primaryKey.rawValue] = primaryFingerprint + email
```
tomholub, line 273: this is string concatenation, not a composed index, right? If string concat, then `primaryFingerprint + "," + base64(email)`.

The comma is necessary so that id of `abcd,efgh@example.com` and `abcdef,gh@example.com` (or their base64 representations) don't conflict

ivan-ushakov: Sure, it is still in progress so any ideas are welcome. I like idea with base64 encoding, let's use it.

FlowCrypt/Models/Realm Models/KeypairRealmObject.swift
```diff
@@ -61,17 +63,28 @@ extension KeyInfoRealmObject {
...
     /// associated user email
     var account: String {
         guard let email = user?.email else { fatalError() }
         return email
     }
 }
+
+extension KeypairRealmObject {
+    static func createPrimaryKey(primaryFingerprint: String, email: String) -> String {
+        var hash = SHA256()
```
ivan-ushakov, line 84: @tomholub
Let's try to use hash function, like you suggested before. BTW since we switched to iOS 15 we could use new `CryptoKit` instead of `CommonCrypto`.

FlowCrypt/Functionality/DataManager/Encrypted Storage/SchemaMigration.swift
```diff
@@ -0,0 +1,127 @@
...
+            return self[property.propertyKey]
+        }
+        set {
+            self[property.propertyKey] = newValue
+        }
+    }
+}
+
+extension SchemaMigration.Properties {
+    enum User: String, RealmProperty {
```
ivan-ushakov, line 33: I did this to avoid string duplication. Now we keep all Realm property names statically typed.

tomholub: I don't really understand this - I think @sosnovsky may better be able to have a look.

FlowCrypt/Models/Realm Models/KeypairRealmObject.swift
tomholub: Forgot a comma again. Else hashes may collide.

ivan-ushakov: Not sure about this because we update has with two values, not use hash function on concatenated string. Maybe we need a unit test to check this.

tomholub: Updating hash twice or concatenating the string first should amount to the same thing. You can update it a third time in between the two with a comma. Unit test is not needed as long as we do one thing and stick to it.

ivan-ushakov:
```
let primaryFingerprint = "ABC"
let email = "test@example.com"

var h = SHA256()
h.update(data: primaryFingerprint.data(using: .utf8)!)
h.update(data: email.data(using: .utf8)!)
print(h.finalize())

var h1 = SHA256()
h1.update(data: (email + primaryFingerprint).data(using: .utf8)!)
print(h1.finalize())
```

output:
```
SHA256 digest: c709f433890955e1b8a818859b0c52b46acb414b2f069f80e3123e5270d5c548
SHA256 digest: 8fd18525b74f59bb0dd80376ec8a4d918e6e6e46f55c53f2ed2f35d18bca9e3e
```

FlowCrypt/Functionality/DataManager/Encrypted Storage/SchemaMigration.swift
sosnovsky: It's a good idea to have enums for all database properties - it'll help to avoid typos when typing object properties in migrations. Also property names will have autocompletion, which is useful too.

I see a couple possible improvements here:
- what about using `rawValue` instead of custom `propertyKey` property? So we won't need `var propertyKey: String { ...` in every `RealmProperty` and protocol will look like this
```swift
protocol RealmProperty {
     var rawValue: String { get }
}
```

- also this code looks a bit messy
```swift
enum SchemaMigration {}

 extension SchemaMigration {
     enum Properties {}
 }

 extension SchemaMigration.Properties {
...
```
Can we use just simple struct like this?
```swift
struct SchemaMigration {
    struct Properties {
        enum User: String, RealmProperty {
...
```

FlowCrypt/Models/Realm Models/KeypairRealmObject.swift
```diff
@@ -83,6 +83,7 @@ extension KeypairRealmObject {
     static func createPrimaryKey(primaryFingerprint: String, email: String) -> String {
         var hash = SHA256()
         hash.update(data: primaryFingerprint.data())
+        hash.update(data: Data(count: 1))
```
tomholub, line 86: Is this a null character? At least should be explicit. Let's do `",".data()`

tomholub (changes requested):

FlowCrypt/Models/Realm Models/KeypairRealmObject.swift
ivan-ushakov: Yes, one zero byte.

sosnovsky (approved):
Looks good now 👍

FlowCrypt/Models/Realm Models/KeypairRealmObject.swift
ivan-ushakov: Are you sure that `,` could not be injected with `primaryFingerprint` or `email`?

tomholub: You're right. I'm more sure that a null byte wouldn't get through than a comma. I've reverted it back to what you did.

tomholub:
@ivan-ushakov please also look at the conflict. Else all good

tomholub (approved):

PR #1207, opened 2021-12-11
issue 414 decide by bundle how to treat FES error

### LINKED ISSUE #414: decide based on consumer / enterprise build when trying FES

tomholub:
This PR adds functionality that requires FES to be running on enterprise bundle, and allows to skip FES integration (if not running) on consumer bundle.

Also adding an `--enterprise` sysarg for debug bundle which will make it behave like enterprise app.

close #414

future tests

- [ ] #1208
- [ ] #1209
- [ ] #1210

### To be filled by reviewers


FlowCrypt/Functionality/Services/Account Server Services/EnterpriseServerApi.swift
```diff
@@ -162,15 +158,48 @@ class EnterpriseServerApi: EnterpriseServerApiType {
...
+    
+    private func shouldTolerateWhenCallingOpportunistically(_ error: Error) async -> Bool {
+        if Bundle.isEnterprise() {
+            return false // FlowCrypt Enterprise Server (FES) required on enterprise bundle
+        }
+        // on consumer release, FES is called opportunistically - if it's there, it will be used
+        // guards first - don't tolerate unknown / other errors. Only interested in network errors.
+        guard let apiError = error as? ApiError else { return false }
+        guard let nsError = apiError.internalError as NSError? else { return false }
+        guard Constants.getToleratedNSErrorCodes.contains(nsError.code) else { return false }
```
sosnovsky, line 176: For removing duplicated code and making it a bit more readable we can use single `guard` like this:
```swift
guard let apiError = error as? ApiError,
      let nsError = apiError.internalError as NSError?,
      Constants.getToleratedNSErrorCodes.contains(nsError.code)
else { return false }
```

FlowCrypt/Functionality/Services/Account Server Services/EnterpriseServerApi.swift
```diff
@@ -162,15 +158,48 @@ class EnterpriseServerApi: EnterpriseServerApiType {
...
-    private func shouldTolerateWhenCallingOpportunistically(_ error: Error) -> Bool {
-        guard
-            let apiError = error as? ApiError,
-            let nsError = apiError.internalError as NSError?,
-            Constants.getToleratedNSErrorCodes.contains(nsError.code)
-        else {
+    private func isExpectedFesServiceResponse(responseData: Data) -> Bool {
+        // "try?" because unsure what server is running there, want to test without failing
+        guard let responseDictionary = try? responseData.toDict() else { return false }
+        guard let service = responseDictionary["service"] as? String else { return false }
```
sosnovsky, line 164: Single `guard` can be used here too:
```swift
guard let responseDictionary = try? responseData.toDict(),
      let service = responseDictionary["service"] as? String
else { return false }
```

sosnovsky (approved):

FlowCrypt/Functionality/Services/Account Server Services/EnterpriseServerApi.swift
tomholub: I know. I just seem to prefer the readability of separate guards. Since you find a combined guard more readable, I'll combine them back.

tomholub:
I'd be surprised if this passes tests, I think I'll still be fixing it up.

tomholub:
Interesting, it does pass. That leaves us with writing appium tests for it: 

- #1210 
- #1209 
- #1208

FlowCrypt/Functionality/Services/Account Server Services/EnterpriseServerApi.swift
sosnovsky: Yes, it looks a bit easier to read when you have just conditions on each line, without `guard` (I think most Swift style guides use this format too)

PR #1220, opened 2021-12-13
Bugfix/issue 1074 send button

### LINKED ISSUE #1074: appium doesn't see send button until scrolled down
This is interesting - the create email button always hovers in the same space. Why is it needed to scroll down before we can tap it? (in actual app, I don't need to scroll to see/tap the button)

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1065#discussion_r755169995_

sosnovsky:
This PR updates test's `clickCreateEmail` helper function.

close #1074

### To be filled by reviewers


FlowCryptUI/Nodes/ComposeButtonNode.swift
```diff
@@ -15,7 +15,8 @@ public final class ComposeButtonNode: ASButtonNode {
         super.init()
         onTap = action
         backgroundColor = .main
-        setTitle("+", with: UIFont.boldSystemFont(ofSize: 30), with: .white, for: .normal)
+        accessibilityIdentifier = "composeMessageButton"
```
tomholub, line 18: We now have a set format for accessibility identifiers: always start with `aid` (means accessibility id) and separate with dashes. So here let's do `aid-compose-message-button`

appium/package.json
```diff
@@ -31,7 +31,7 @@
     "@wdio/jasmine-framework": "^7.16.6",
     "@wdio/local-runner": "^7.16.10",
     "@wdio/spec-reporter": "^7.16.9",
-    "appium": "1.22.1",
+    "appium": "^1.22.1",
```
tomholub, line 34: Better not use carets, so that dependency versions are determinate.

appium/tests/helpers/ElementHelper.ts
```diff
@@ -49,10 +49,12 @@ class ElementHelper {
...
-    await element.waitForDisplayed();
+  static waitAndClick = async (element: WebdriverIO.Element, delayMs?: number) => {
+    await this.waitElementVisible(element);
     // stability fix to make sure element is ready for interaction
-    await browser.pause(delayMs);
+    if (delayMs !== undefined) {
+      await browser.pause(delayMs);
+    }
     await element.click();
   }
```
tomholub, line 59: This will likely make tests more brittle, I believe waiting 50ms by default was helping them run more smoothly.

If you want to have an option to skip this, I recommend to use `if (delayMs) {`. Then the argument doesn't have to be optional, and you can set `waitAndClick(xxx, 0)` for the one case you explicitly don't want to wait.

sosnovsky: During my testing it worked well - probably `waitElementVisible(element)` is enough for element waiting.
I'll add `50ms` delay back to be sure it won't fail in the future.

sosnovsky: `50ms` should be ok - it adds just 2 seconds for total tests duration

PR #1225, opened 2021-12-14
#1091 add google id tokens refresh

### LINKED ISSUE #1091: more generalized in-memory storage + remember id_token
for #955 we will need a way to retain id_token longer.

Currently, we retrieve it during setup and then use it to authenticate with EKM, then forget it.

I'd like it instead be remembered, same like in-memory pass phrase (but always only in memory - regardless how pass phrase is remembered).

You can use the same mechanism/even same class and object that is used for in memory pass phrase storage, as long as you give it a more general name like `InMemoryStorage`. Or you can have a separate instance of that same class to be used for id tokens, or you can have two different classes, based on what you think makes sense.

Either way, whenever we wipe in-memory pass phrases (eg logout), id tokens should get wiped too.

This is a preparation for #955 where soon, we'll have to be calling FlowCrypt Enterprise Server (FES) and authenticate ourselves while doing that.

sosnovsky:
This PR adds auto-refresh for Google id tokens.

close #1091

- Difficult to test (explain why) - it's `GTMAppAuth` built-in functionality and we're using mock `GoogleUserService` in tests.

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Setup/SetupInitialViewController.swift
+++ b/FlowCrypt/Controllers/Setup/SetupInitialViewController.swift
@@ -157,10 +157,20 @@ extension SetupInitialViewController {
         }
     }
 
+    private func getIdToken() async throws -> String {
+        let googleService = GoogleUserService(
+            currentUserEmail: user.email,
+            appDelegateGoogleSessionContainer: nil
+        )
+
+        return try await googleService.getCachedOrRefreshedIdToken()
+    }
+
     private func fetchKeysFromEKM() {
         Task {
             do {
-                let result = try await emailKeyManagerApi.getPrivateKeys(currentUserEmail: user.email)
+                let idToken = try await getIdToken()
+                let result = try await emailKeyManagerApi.getPrivateKeys(idToken: idToken)
                 switch result {
                 case .success(keys: let keys):
                     proceedToSetupWithEKMKeys(keys: keys)
@@ -183,7 +193,7 @@ extension SetupInitialViewController {
                 if case .noPrivateKeysUrlString = error as? EmailKeyManagerApiError {
                     return
                 }
-                showAlert(message: error.localizedDescription, onOk: { [weak self] in
+                showAlert(message: error.errorMessage, onOk: { [weak self] in
                     self?.state = .decidingIfEKMshouldBeUsed
                 })
             }
--- a/FlowCrypt/Functionality/Services/GoogleUserService.swift
+++ b/FlowCrypt/Functionality/Services/GoogleUserService.swift
@@ -52,8 +52,17 @@ extension IdToken {
     }
 }
 
-enum TokenError: Error {
+enum IdTokenError: Error, CustomStringConvertible {
     case missingToken, invalidJWTFormat, invalidBase64EncodedData
+
+    var description: String {
+        switch self {
+        case .missingToken:
+            return "id_token_missing_error_description".localized
+        case .invalidJWTFormat, .invalidBase64EncodedData:
+            return "id_token_invalid_error_description".localized
+        }
+    }
 }
 
 protocol GoogleUserServiceType {
@@ -95,7 +104,7 @@ final class GoogleUserService: NSObject, GoogleUserServiceType {
         tokenResponse?.idToken
     }
 
-    var userToken: String? {
+    var accessToken: String? {
         tokenResponse?.accessToken
     }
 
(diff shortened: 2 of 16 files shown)


FlowCrypt/Functionality/Services/GoogleUserService.swift
```diff
@@ -42,6 +42,20 @@ struct GoogleUser: Codable {
...
+    let exp: Int
+}
+
+extension IdToken {
+    var isExpired: Bool {
+        Double(exp) < Date().timeIntervalSince1970
+    }
+}
+
+enum TokenError: Error {
```
tomholub, line 55: I guess `IdTokenError`

FlowCrypt/Functionality/Services/GoogleUserService.swift
```diff
@@ -73,16 +87,16 @@ final class GoogleUserService: NSObject, GoogleUserServiceType {
...
 
-    var idToken: String? {
-        authorization?.authState
-            .lastTokenResponse?
-            .idToken
+    private var idToken: String? {
+        tokenResponse?.idToken
+    }
+
+    var userToken: String? {
```
tomholub, line 98: should be called `accessToken`

FlowCrypt/Functionality/Services/Remote Private Key Services/EmailKeyManagerApi.swift
```diff
@@ -56,12 +56,13 @@ actor EmailKeyManagerApi: EmailKeyManagerApiType {
...
+        let googleService = GoogleUserService(
             currentUserEmail: currentUserEmail,
             appDelegateGoogleSessionContainer: nil // only needed when signing in/out
-        ).idToken else {
-            throw EmailKeyManagerApiError.noGoogleIdToken
-        }
+        )
+
+        guard let idToken = try? await googleService.getIdToken()
+        else { throw EmailKeyManagerApiError.noGoogleIdToken }
```
tomholub, line 65: Please refactor this method: `idToken` should be supplied into `getPrivateKeys` instead of `currentUserEmail`. Then `GoogleUserService` doesn't have to be here at all.

FlowCryptAppTests/Functionality/Mail Provider/GmailServiceTest.swift
```diff
@@ -46,7 +46,7 @@ class GmailServiceTest: XCTestCase {
 class GoogleUserServiceMock: GoogleUserServiceType {
     var authorization: GTMAppAuthFetcherAuthorization?
     func renewSession() async throws {
-        await Task.sleep(1_000_000_000)
+        try await Task.sleep(nanoseconds: 1_000_000_000)
```
tomholub, line 49: I was earlier wondering what is the difference between the two

FlowCrypt/Functionality/Services/GoogleUserService.swift
```diff
@@ -229,6 +243,55 @@ extension GoogleUserService {
...
+            authorization?.authState.performAction { accessToken, idToken, error in
+                guard let accessToken = accessToken, let idToken = idToken else {
+                    let tokenError = error ?? AppErr.unexpected("Shouldn't happen because received nil error and nil token")
+                    return continuation.resume(throwing: tokenError)
+                }
+                let result = (accessToken, idToken)
+                return continuation.resume(with: .success(result))
+            }
+        }
+    }
```
tomholub, line 292: Does this work without user interaction?

FlowCrypt/Functionality/Services/GoogleUserService.swift
```diff
@@ -229,6 +243,55 @@ extension GoogleUserService {
     }
 }
 
+// MARK: - Tokens
+extension GoogleUserService {
+    func getIdToken() async throws -> String {
```
tomholub, line 248: Let's call it `getCachedOrRefreshedIdToken` to make it clear an external call may happen ?

tomholub (changes requested):

FlowCrypt/Functionality/Services/GoogleUserService.swift
sosnovsky: Yes, I tested it - token is automatically refreshed in background, without user interaction.

FlowCryptAppTests/Functionality/Mail Provider/GmailServiceTest.swift
sosnovsky: Xcode showed deprecation warning for `await Task.sleep(1_000_000_000)`, so it's the same, only named parameter `nanoseconds` was added

FlowCrypt/Controllers/Setup/SetupInitialViewController.swift
```diff
@@ -157,10 +157,20 @@ extension SetupInitialViewController {
         }
     }
 
+    private func getIdToken() async throws -> String {
+        let googleService = GoogleUserService(
+            currentUserEmail: user.email,
+            appDelegateGoogleSessionContainer: nil
+        )
```
tomholub, line 164: We probably could/should dependency-inject this through AppContext. I'll file another issue.

tomholub (approved):
Thanks!

Issue #955, opened 2021-11-08
send password-protected messages through web portal


tomholub:
will add details, for now filing sub-issues:
 - [x] #1091 
 - [x] #1221
 - [x] #1222
 - [x] #1231
 - [x] #1254
 - [x] #1306
 - [x] #1253

sosnovsky:
@tomholub should we close this one too, as all subtasks are completed now?

PR #1249, opened 2021-12-23
#1221 UI for message passwords

### LINKED ISSUE #1221: UI for password encrypted message functionality
part of #955

sosnovsky:
This PR adds UI for setting message passwords

close #1221

- Tests added or updated - updated `SendEmailToRecipientWithoutPublicKey` test with new password functionality

### To be filled by reviewers


sosnovsky:
@tomholub password UI is ready, but I think we should merge it into the `master` only after finishing other tasks from #955, as currently it's only UI without functionality.

tomholub:
> @tomholub password UI is ready, but I think we should merge it into the `master` only after finishing other tasks from #955, as currently it's only UI without functionality.

We can merge it, this will take care of it:

> let domainsWithPasswordSupport = ["flowcrypt.com"]

There are no customers on flowcrypt.com domain 😃 so we can safely merge it without affecting anyone.

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -634,6 +642,17 @@ extension ComposeViewController {
         }
     }
 
+    private func passwordNode() -> ASCellNode {
```
tomholub, line 645: let's use the term `messagePassword` so that we always know which one this is. So that would be `messagePasswordNode`

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -634,6 +642,17 @@ extension ComposeViewController {
...
+        let input = contextToSend.hasPassword
```
tomholub, line 646: `hasMessagePassword`

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -996,6 +1019,38 @@ extension ComposeViewController {
             }
         }
     }
+
+    private func setMessagePassword() {
+        Task {
+            contextToSend.password = await enterMessagePassword()
```
tomholub, line 1025: messagePassword

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -996,6 +1019,38 @@ extension ComposeViewController {
...
+            node.reloadSections([Section.password.rawValue], with: .automatic)
+        }
+    }
+
+    private func enterMessagePassword() async -> String? {
```
tomholub, line 1030: I like modals as async methods - it really cleans up the logic of wherever it ends up used.

FlowCrypt/Controllers/Compose/ComposeViewDecorator.swift
```diff
@@ -98,10 +98,32 @@ struct ComposeViewDecorator {
         return (text + message).attributed(.regular(17))
     }
 
+    func styledEmptyPasswordInput() -> MessagePasswordCellNode.Input {
```
tomholub, line 101: styledEmptyMessagePasswordInput

FlowCrypt/Controllers/Compose/ComposeViewDecorator.swift
```diff
@@ -98,10 +98,32 @@ struct ComposeViewDecorator {
...
+            text: "compose_password_placeholder".localized.attributed(
+                .regular(14),
+                color: .warningColor
+            ),
+            color: .warningColor,
+            image: UIImage(systemName: "lock")?.tinted(.warningColor)
+        )
+    }
+
+    func styledFilledPasswordInput() -> MessagePasswordCellNode.Input {
```
tomholub, line 112: styledFilledMessagePasswordInput

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageContext.swift
```diff
@@ -0,0 +1,38 @@
...
+    var subject: String?
+    var password: String?
+    var attachments: [MessageAttachment] = []
+}
+
+extension ComposeMessageContext {
+    var hasPassword: Bool {
+        guard let password = password else { return false }
+        return password.isNotEmpty
+    }
```
tomholub, line 23: In that case maybe it should also have a `getPassword` that would return a nil on empty string. And make password not externally readable? So that we don't just retrieve `.password` which could be empty string - who knows what kind of vulnerability that may lead to. (encrypting message for a password which is an empty string. OpenPGP.js is probably clever to handle it, but don't want to risk it)

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageContext.swift
```diff
@@ -0,0 +1,38 @@
...
+        recipients
+            .filter {
+                if case .keyNotFound = $0.state { return true }
+                return false
+            }
+            .first(where: {
+                guard let domain = $0.email.recipientDomain else { return !withPasswordSupport }
+                let domainsWithPasswordSupport = ["flowcrypt.com"]
+                let supportsPassword = domainsWithPasswordSupport.contains(domain)
+                return withPasswordSupport == supportsPassword
```
tomholub, line 35: Regarding support, we don't care about domain of recipient, only of sender. So this can be removed.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageError.swift
```diff
@@ -31,6 +32,8 @@ enum MessageValidationError: Error, CustomStringConvertible, Equatable {
             return "compose_no_pub_sender".localized
         case .noPubRecipients:
             return "compose_recipient_no_pub".localized
+        case .needsMessagePassword:
+            return ""
```
tomholub, line 36: Not sure if intentional. If so, please add a code comment.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -122,36 +110,57 @@ final class ComposeMessageService {
...
         }
+
+        func hasRecipientsWithoutPubKey(withPasswordSupport: Bool) -> Bool {
+            recipients
+                .filter { $0.keyState == .empty }
+                .first(where: {
+                    guard let domain = $0.email.recipientDomain else { return !withPasswordSupport }
+                    let supportsPassword = domainsWithPasswordSupport.contains(domain)
+                    return withPasswordSupport == supportsPassword
+                }) != nil
```
tomholub, line 141: Same here - can be removed. We only care that recipient has no pubkey, and sender is on `flowcrypt.com` domain - then we allow the functionality.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -122,36 +110,57 @@ final class ComposeMessageService {
...
+        }
+
         logger.logDebug("validate recipients: \(recipients)")
-        logger.logDebug("validate recipient keyStates: \(recipients.map { $0.keyState })")
-        guard !contains(keyState: .empty) else {
+        logger.logDebug("validate recipient keyStates: \(recipients.map(\.keyState))")
+
+        let domainsWithPasswordSupport = ["flowcrypt.com"]
```
tomholub, line 147: Let's call it `senderDomainsWithPasswordSupport`

appium/tests/data/index.ts
```diff
@@ -75,7 +75,12 @@ export const CommonData = {
     signatureBadgeText: 'not signed'
   },
   recipientWithoutPublicKey: {
-    email: 'no.publickey@flowcrypt.com'
+    emailWithoutPasswordSupport: 'some@gmail.com',
+    emailWithPasswordSupport: 'no.publickey@flowcrypt.com',
```
tomholub, line 79: And here - only one recipient needs to be tested. It's the sender that matters.

tomholub (changes requested):
Didn't run app yet - please see code comments.

FlowCrypt/Controllers/Compose/ComposeViewController.swift
sosnovsky: Yeah, `async/await` is a great addition to Swift, makes code much more readable.
Especially in such cases as modals where previously we needed several action handlers.

FlowCryptCommon/Extensions/StringExtensions.swift
```diff
@@ -61,27 +61,14 @@ public extension NSAttributedString {
 
 // MARK: Email parsing
 public extension String {
-    var userAndRecipientDomain: (user: String, domain: String)? {
+    var email: (username: String?, domain: String?) {
```
tomholub, line 64: It would be more ergonomic to return `EmailParts?` where `EmailParts` is `struct { username, domain }`. Then you only have to do one nil check. Also rename `var email` to `var emailParts`?

FlowCrypt/Functionality/Services/Account Server Services/EnterpriseServerApi.swift
```diff
@@ -61,7 +61,7 @@ class EnterpriseServerApi: EnterpriseServerApiType {
 
     func getActiveFesUrl(for email: String) async throws -> String? {
         do {
-            guard let userDomain = email.recipientDomain,
+            guard let userDomain = email.emailDomain,
```
tomholub, line 64: Much better 👍

FlowCryptCommon/Extensions/StringExtensions.swift
sosnovsky: Agree, having one `nil` will be easier to check.

tomholub:
Please make sure to also resolve the hidden conversations (github hides them by default when too many)

image

tomholub (changes requested):
Almost perfect, except:

 - setting empty string on this modal should not be allowed. I could use cancel for that. So either I enter an actual string, or it should tell me to enter one. Soon there will also be more complex evaluation of password strength, so some way to indicate the input is invalid is needed anyway.

image

- crash when trying to remove recipient

image

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageContext.swift
sosnovsky: I implemented this using custom getter for `messagePassword` - https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-1221-message-password/FlowCrypt/Functionality/Services/Compose%20Message%20Service/ComposeMessageContext.swift#L18

sosnovsky:
> * crash when trying to remove recipient

@tomholub are you able to reproduce it with the latest code? As recipients deletion works well for me - no crash occurred

tomholub (commented):
Changes look good. Will retry the deletion

tomholub:
Still reproduce, it's actually when tapping the recipient. This is on simulator.

1) start typing email of someone who doesn't have pubkey in recipient field. Choose suggestion from offered contacts.
2) now the contact is gray and it prompts me to add password
3) I tap the gray recipient bubble
4) crash:

image

tomholub (approved):
This fixed the issue 👍

PR #1311, opened 2022-01-16
Handle error if key was not submit to Attester

### LINKED ISSUE #1303: broken procedure that generates a key during setup
Ok, so there are three issues when creating a key during setup:
 - missing / unclear details of that error message. It should be showing clearly the URL and http code like other API errors do. This will need an appium test as well. The Attester mock will be set to return a 400 on submission with some error message, and the above modal must show that error message (we had a mechanism for this that was supposed to work across modals)
 - the fact that there is an error submitting to attester
 - per Dmitry's account, whenever there isn't the above error, backup doesn't get sent to inbox

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/484#issuecomment-1011990874_

### LINKED ISSUE #1304: The key is not saved when user set key and relogin
**Steps:**

1. Login to outlook accoun with key
2. Create key
3. Logout from outlook account
4. Try to login again 
**Actual:** User needs to setup new key, old key is not saved
**Expected:** User should be able to login with pass phrase

Kharchevskyi:
This PR:
* Fixes error handling if key was not submit to Attester
* Fix storing backup key in inbox

close #1304

- Difficult to test (explain why)
- Not worth testing
- Tests added or updated

### To be filled by reviewers


FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
```diff
@@ -119,10 +113,9 @@ private actor Service {
             try appContext.passPhraseService.savePassPhrase(with: passPhrase, storageMethod: .memory)
         }
 
-        await submitKeyToAttesterAndShowAlertOnFailure(
+        try await submitKeyToAttester(
```
Kharchevskyi, line 116: In case this operation is failed pass phrase and key are already saved in app. 
Should we change order of this operation during generating backup?

Kharchevskyi: @tomholub @sosnovsky

Kharchevskyi: please check implementation for 
```
func setupAccount(
        passPhrase: String,
        storageMethod: StorageMethod,
        viewController: ViewController
) 
```

FlowCrypt/Functionality/Services/ApiCall.swift
```diff
@@ -63,15 +63,23 @@ struct ApiError: LocalizedError {
...
                 internalError: httpError.error
             )
         }
+        
+        guard let object = try? JSONDecoder().decode(HttpError.self, from: data) else {
+            let errorDescription = httpError.error?.localizedDescription
+                ?? String(data: data, encoding: .utf8)
+                ?? "missed description"
+            
+            return ApiError(
```
Kharchevskyi, line 78: return error description or message from received data in case it is not possible to decode json object

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
Kharchevskyi: @tomholub can you please help with this error?

Kharchevskyi: @sosnovsky Do you aware of what this error message means?

sosnovsky: Not sure, but looks like something wrong with checking your existing public key. As error says you already have `previously recorded key`, but app tries to submit another one.
In this case app should open `SetupManuallyImportKeyViewController` instead of `SetupGenerateKeyViewController`.
Does it work well for you on `master` branch or it fails there too?

FlowCrypt/Functionality/Mail Provider/Message Gateway/Imap+send.swift
```diff
@@ -6,9 +6,9 @@ import Foundation
 
 extension Imap: MessageGateway {
     func sendMail(input: MessageGatewayInput, progressHandler: ((Float) -> Void)?) async throws {
-        try await withCheckedThrowingContinuation { [weak smtpSess] (continuation: CheckedContinuation<Void, Error>) in
-            guard let session = smtpSess else {
-                return continuation.resume(returning: ())
```
Kharchevskyi, line 11: @sosnovsky on master there is an error with submitting key to attester with IMAP usage so it would not reproduce

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
Kharchevskyi: https://github.com/FlowCrypt/flowcrypt-ios/pull/1311#discussion_r787679362 
Looks like due to this issue we failed to backup key in inbox but it was submitted to attester

FlowCrypt/Functionality/Mail Provider/Message Gateway/Imap+send.swift
sosnovsky: @Kharchevskyi I think for testing new public key submit you need to delete your existing public key at https://flowcrypt.com/attester/
But not sure if it'll affect your previous emails which were using your current public key.

Kharchevskyi: If i remove this pub key from attester manually as you suggest flow works. 
I can create a new backup in inbox and submit it. 

But if I remove this backup from web interface and try to setup a new backup I receive the same error.
Do we need to handle it somehow?

sosnovsky: Yes, it should be better to have alert with such actions for key error
- `Replace public key` - opens  https://flowcrypt.com/attester/update-public-keys
- `Delete public key` - opens https://flowcrypt.com/attester/update-public-keys?intention=delete
- `Cancel` - redirects user back to sign in screen

@tomholub does such error handling looks good for you?

Kharchevskyi:
@tomholub can you please check comments in this pr

FlowCrypt/Functionality/Error Handling/KeyServiceErrorHandler.swift
```diff
@@ -10,14 +10,37 @@ import UIKit
 
 enum CreateKeyError: Error {
     case weakPassPhrase(_ strength: CoreRes.ZxcvbnStrengthBar)
-    // Missing user email
+    /// Missing user email
     case missedUserEmail
```
tomholub, line 14: All instances of `missed ...` should be all `missing ...`

`missedUserEmail` looks like broken English to me

I'll file another issue for that.

FlowCrypt/Functionality/Mail Provider/Message Gateway/Imap+send.swift
tomholub: @sosnovsky thank you for the suggestion - it's well meaning and a good idea, but in this case unneeded because this is such a fringe situation.

We have nearly zero users using SMTP/IMAP. In the future we may have more, but it will never be nearly enough to to amount of users that we have for Oauth2 due to the difference in convenience. I'll file the suggestion as a separate issue for sometime later in the future.

FlowCrypt/Functionality/Services/ApiCall.swift
```diff
@@ -63,15 +63,23 @@ struct ApiError: LocalizedError {
...
             return ApiError(
                 errorDescription: httpError.error?.localizedDescription ?? "",
                 internalError: httpError.error
             )
         }
+        
+        guard let object = try? JSONDecoder().decode(HttpError.self, from: data) else {
+            let errorDescription = httpError.error?.localizedDescription
+                ?? String(data: data, encoding: .utf8)
+                ?? "missed description"
```
tomholub, line 76: MISSSING, not MISSED :-) applies to everywhere

tomholub (approved):
Looks ok to me overall

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
```diff
@@ -119,33 +115,35 @@ private actor Service {
...
                 email: email,
                 pubkey: publicKey,
                 token: appContext.dataService.token
             )
         } catch {
-            let message = "Failed to submit Public Key"
-            await viewController.showAlert(error: error, message: message)
+            throw CreateKeyError.submitKey(error)
         }
     }
```
tomholub, line 148: This is consistent with our general error handling approach, but now you can't implement my request `we have to add a modal with retry/skip option` easily. For that, you would need local handling of the error like before.

Nevermind, since IMAP is a fringe usecase, we can leave it as is for now for simplicity.

tomholub (dismissed):

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
tomholub: The error rendering is clear now, and for your manual testing, using https://flowcrypt.com/attester/update-public-keys?intention=delete will help you as @sosnovsky mentioned.

As far as user experience is concerned, when this error happens, they are screwed because they won't know what to do, but that's ok for now since few people if anyone uses IMAP/SMTP.

tomholub:
@Kharchevskyi please check the tests later - still not passing

sosnovsky (approved):
All good 👍

PR #1323, opened 2022-01-20
#1233 Decrypt attachments on tap

### LINKED ISSUE #1233: don't decrypt attachments before tapping + show modal on error
also mostly fixed #1143. The error was that while the text could be decrypted, the attachments were throwing decrypt errors. I've made it recognize decrypt errors and download the original file, but it needs another issue to actually render the decrypt error before downloading, make sure that's what the user wanted.

image

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1219#issuecomment-997013348_

sosnovsky:
This PR implements attachments decrypt on tap, so they're not decrypted on message open.

close #1233

- Tests added or updated - updated `user is able to view message processing errors` and `user is able to view encrypted email with attachment` tests

### To be filled by reviewers


tomholub (dismissed):
This looks good. That made me think: there are three types of attachments:
 - plain / not encrypted
 - encrypted as a separate file
 - encrypted with pgp/mime (a mime message with attachments, encrypted whole)

Could you double check that the third case works after this change? Please try email with subject `[enigmail] encrypted+signed+file PGP/MIME` on `flowcrypt.compatibiltiy` account.

I supose the TS core will auto-decrypt it and return back as decrypted (plain), which should work. Worth checking though.

sosnovsky:
Here is how it works in this branch

https://user-images.githubusercontent.com/1062093/150418281-556155d2-334b-4b85-a12a-e95aff8de920.mp4

Signature can't be verified, but I get the same error on `master` branch, is it okay?

sosnovsky:
The same signature error is in browser extension, so should be good

tomholub:
Looks perfect :+1:

tomholub:
Just the tests to fix else looks great to me.

tomholub (approved):

tomholub (approved):

PR #1324, opened 2022-01-20
Fix for collection view layout

### LINKED ISSUE #1315: iPad - fix layout for recipients
Screenshot 2022-01-13 at 14 51 41

Kharchevskyi:
This PR fixes layout for recipients input

close #1315

- Does not need tests (ui changes)

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Compose/ComposeViewController.swift
+++ b/FlowCrypt/Controllers/Compose/ComposeViewController.swift
@@ -22,7 +22,6 @@ private struct ComposedDraft: Equatable {
  * - Or from *ThreadDetailsViewController* controller by tapping on *reply* or *forward*
  **/
 final class ComposeViewController: TableNodeViewController {
-    private let minRecipientsPartHeight: CGFloat = 44
     var calculatedRecipientsPartHeight: CGFloat? {
         didSet {
             node.reloadRows(at: [recipientsIndexPath], with: .fade)
@@ -31,6 +30,7 @@ final class ComposeViewController: TableNodeViewController {
 
     private enum Constants {
         static let endTypingCharacters = [",", " ", "\n", ";"]
+        static let minRecipientsPartHeight: CGFloat = 44
     }
 
     enum State {
@@ -721,18 +721,13 @@ extension ComposeViewController {
     private func recipientsNode() -> RecipientEmailsCellNode {
         return RecipientEmailsCellNode(
             recipients: recipients.map(RecipientEmailsCellNode.Input.init),
-            height: calculatedRecipientsPartHeight ?? minRecipientsPartHeight
+            height: calculatedRecipientsPartHeight ?? Constants.minRecipientsPartHeight
         )
             .onLayoutHeightChanged { [weak self] layoutHeight in
-                guard let self = self, self.calculatedRecipientsPartHeight != layoutHeight else {
+                guard self?.calculatedRecipientsPartHeight != layoutHeight, layoutHeight > 0 else {
                     return
                 }
-
-                if layoutHeight > self.minRecipientsPartHeight {
-                    self.calculatedRecipientsPartHeight = layoutHeight
-                } else if layoutHeight > 0 {
-                    self.calculatedRecipientsPartHeight = layoutHeight
-                }
+                self?.calculatedRecipientsPartHeight = layoutHeight
             }
             .onItemSelect { [weak self] (action: RecipientEmailsCellNode.RecipientEmailTapAction) in
                 switch action {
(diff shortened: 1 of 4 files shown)


tomholub:
Shouldn't the recipient bubbles flow like text? Eg how the browser extension does it, or more or less how it was before. After this change it's one recipient per line - why so?

Kharchevskyi:
Can change it to layout one to another like text flow

tomholub:
What was the issue in this screenshot?

image

Kharchevskyi:
Wrong calculation for collection view frame height. Each new recipient adds extra space to the bottom of collection view. 

Will just fix height calculation.

tomholub:
ah - understood

> Will just fix height calculation.

:+1:

Kharchevskyi:
@tomholub please check attached sreen shots.

FlowCryptUI/Cell Nodes/RecipientEmailsCellNode.swift
```diff
@@ -55,15 +61,8 @@ final public class RecipientEmailsCellNode: CellNode {
...
-        let recipientNodeInset = RecipientEmailNode.Constants.layoutInsets.height
-            + RecipientEmailNode.Constants.titleInsets.height
 
-        let textSize: CGSize = recipients.first?.email.size() ?? .zero
-        let recipientsHeight = (textSize.height + recipientNodeInset) * CGFloat(recipients.count)
-        let insets = Constants.minimumLineSpacing * CGFloat(recipients.count - 1)
-        let height = recipientsHeight + insets + Constants.sectionInset.height
-
-        collectionNode.style.preferredSize.height = height
+        collectionNode.style.preferredSize.height = collectionLayoutHeight
```
Kharchevskyi, line 65: @sosnovsky note for reviewing:
Calculation for flow layout height happens after we set height for collectionNode, so it is not possible to calculate height here( 

Thats why we need to observe calculated height from collection layout in controller to be able to react on height changes and update `RecipientEmailsCellNode` with updated height value

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -713,11 +719,27 @@ extension ComposeViewController {
     }
 
     private func recipientsNode() -> RecipientEmailsCellNode {
-        RecipientEmailsCellNode(recipients: recipients.map(RecipientEmailsCellNode.Input.init))
+        return RecipientEmailsCellNode(
+            recipients: recipients.map(RecipientEmailsCellNode.Input.init),
+            height: calculatedRecipientsPartHeight ?? minRecipientsPartHeight
+        )
+            .onLayoutHeightChanged { [weak self] layoutHeight in
```
Kharchevskyi, line 726: observe changes from flow layout

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -22,6 +22,12 @@ private struct ComposedDraft: Equatable {
  * - Or from *ThreadDetailsViewController* controller by tapping on *reply* or *forward*
  **/
 final class ComposeViewController: TableNodeViewController {
+    private let minRecipientsPartHeight: CGFloat = 44
+    var calculatedRecipientsPartHeight: CGFloat? {
+        didSet {
+            node.reloadRows(at: [recipientsIndexPath], with: .fade)
```
Kharchevskyi, line 28: update height of collection node

FlowCryptUI/Cell Nodes/RecipientEmailsCellNode.swift
tomholub: If it's generally surprising for the person reading the code, it may be appropriate to add a comment in the code.

Kharchevskyi: Have just added comment's for reviewers, don't think it worth commenting it.

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -713,11 +719,27 @@ extension ComposeViewController {
...
+            .onLayoutHeightChanged { [weak self] layoutHeight in
+                guard let self = self, self.calculatedRecipientsPartHeight != layoutHeight else {
+                    return
+                }
+
+                if layoutHeight > self.minRecipientsPartHeight {
+                    self.calculatedRecipientsPartHeight = layoutHeight
+                } else if layoutHeight > 0 {
+                    self.calculatedRecipientsPartHeight = layoutHeight
+                }
```
sosnovsky, line 735: Do we need both checks? We can leave only `if layoutHeight > 0` as `minRecipientsPartHeight` is bigger than 0 too.

sosnovsky (commented):
Looks good to me, just one small comment

sosnovsky (approved):

PR #1330, opened 2022-01-24
#1306 Add message password policy

### LINKED ISSUE #1306: password policy for message passwords
same as https://github.com/FlowCrypt/flowcrypt-android/issues/1618

browser implementation https://github.com/FlowCrypt/flowcrypt-browser/pull/4252

sosnovsky:
This PR adds message password policy for checking password strength

close #1306 

- Tests added or updated - updated `SendEmailToRecipientWithoutPublicKey` test

### To be filled by reviewers


FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -114,6 +117,23 @@ final class ComposeMessageService {
...
 
+        if let password = contextToSend.messagePassword, password.isNotEmpty {
+            if subject.lowercased().contains(password.lowercased()) {
+                throw MessageValidationError.subjectContainsPassword
+            }
+
+            let storedPassphrases = passPhraseService.getPassPhrases().map(\.value)
+            if storedPassphrases.contains(password) {
+                throw MessageValidationError.notUniquePassword
+            }
```
tomholub, line 128: Ere these stored pass phrases only, or a combination of stored pass phrases + whatever is currently available in memory for in-memory pass phrases?

sosnovsky:
@tomholub here is a video of current implementation:

https://user-images.githubusercontent.com/1062093/150786879-ab46ecbd-6ceb-433a-8d7c-817cf9ecf418.mp4

What about notifying users about weak password before tapping `Send` - for example, we can show warning instead of green `Web portal password added` message. Do we need it or current implementation is enough for now?

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -338,4 +358,26 @@ extension ComposeMessageService {
...
         return SendableMsgBody(text: text, html: html)
     }
+
+    private func isMessagePasswordStrong(pwd: String, isFesUsed: Bool) -> Bool {
+        let minLength = 8
+
+        guard isFesUsed else {
+            // consumers - just 8 chars requirement
+            return pwd.count >= minLength
+        }
```
tomholub, line 368: We don't support this scenario on iOS yet - the only way to send a password-protected message is with FES. It's ok to add this code but it should come with a note.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
sosnovsky: `getPassPhrases` returns `encryptedStorage.getPassPhrases() + inMemoryStorage.getPassPhrases()` (https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/Functionality/Services/Local%20Private%20Key%20Services/PassPhraseService.swift#L113), so we check against all current user passphrases

tomholub:
> What about notifying users about weak password before tapping Send - for example, we can show warning instead of green Web portal password added message. Do we need it or current implementation is enough for now?

Agree, actually best would be still on the modal where the user inputs it.

Best would be:
 - the modal should describe the requirements for the password
 - there should be visual indication to show if the password is sufficient. Maybe color of the input and the set button
 - user should not be allowed to set it unless it fits the requirements

I don't know if we are trying to do more than what is reasonable to expect from a modal - do you think this needs its own view? Or should modal be enough for this?

tomholub:
On browser extension we color it in red while it's non-conforming, plus it shows the requirements above:

image

On Android it evaluates them one by one which is an overkill, I don't want to do that on iOS. But again there is indication if it conforms + list of requirements.

image

sosnovsky:
I think it's possible to do such check using `UIAlertViewController` - `Set` button will be disabled until all requirements are met. And I'll also add all requirements to `add password` modal

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
tomholub: In that case the variable you test should be renamed to `allAvailablePassPhrases`

sosnovsky:
I updated implementation to check password strength when user types it in modal, here is video:

https://user-images.githubusercontent.com/1062093/150991741-e513cf07-d0fe-4aeb-93de-fd0c51b80a35.mp4

tomholub:
Yup - simple, effective, clear.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -101,7 +104,7 @@ final class ComposeMessageService {
 
         let subject = contextToSend.subject ?? "(no subject)"
 
-        guard let myPubKey = storage.getKeypairs(by: email).map(\.public).first else {
+        guard let myPubKey = storage.getKeypairs(by: sender).map(\.public).first else {
             throw MessageValidationError.missedPublicKey
```
tomholub, line 108: `missingPublicKey`. I'm filing an issue to rename all similar situations.

tomholub (approved):
Code looks good :+1:

tomholub (dismissed):

tomholub (approved):

.semaphore/semaphore.yml
```diff
@@ -5,7 +5,7 @@ agent:
     type: a1-standard-4
     os_image: macos-xcode13
 execution_time_limit:
-  minutes: 90
+  minutes: 100
```
tomholub, line 8: If we are so close to the limit, may be worth to do 120 here and 115 below. 

Eventually we'll optimize / parallelize this.

sosnovsky: Done ✅

tomholub (dismissed):

tomholub (approved):
:+1:

PR #1331, opened 2022-01-24
Current user should be dependency-injected

### LINKED ISSUE #863: current user should be dependency-injected

ivan-ushakov:
close #863

### To be filled by reviewers


FlowCryptAppTests/ExtensionTests.swift
```diff
@@ -94,7 +94,7 @@ extension ExtensionTests {
         let otherYearDate = Date(timeIntervalSince1970: 1579883652)
 
         XCTAssertTrue(dateFormatter.date(from: DateFormatter().formatDate(sameDayDate)) != nil)
-        XCTAssertEqual(DateFormatter().formatDate(sameYearDate), "Jan 24")
+        XCTAssertEqual(DateFormatter().formatDate(sameYearDate), "00:00")
```
ivan-ushakov, line 97: This case was wrong because for the same year and month we format only time

FlowCrypt/Functionality/Services/AppStartup.swift
```diff
@@ -68,8 +68,13 @@ struct AppStartup {
 
         switch entryPoint {
         case .mainFlow:
-            let contentViewController = InboxViewContainerController(appContext: appContext)
-            viewController = SideMenuNavigationController(appContext: appContext, contentViewController: contentViewController)
+            guard let user = appContext.dataService.currentUser else { return }
```
ivan-ushakov, line 71: Should not be possible, so I just return. Other ideas are welcome

FlowCryptAppTests/ExtensionTests.swift
ivan-ushakov: OK, let's wait for your fix

sosnovsky: This test was fixed by another PR - https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCryptAppTests/ExtensionTests.swift#L100, so it should pass after merging `master`

FlowCrypt/Functionality/Services/AppStartup.swift
tomholub: Could it be refactored so that user is somehow an attribute of mainFlow? Like case .mainFlow(let user). That would be best. That way a main flow truly cannot be initiated without user.

FlowCrypt/Controllers/Inbox/InboxViewController+Factory.swift
```diff
@@ -11,10 +11,9 @@ import UIKit
 
 class InboxViewControllerFactory {
     @MainActor
-    static func make(appContext: AppContext, with viewModel: InboxViewModel) -> InboxViewController {
+    static func make(appContext: AppContext, user: User, viewModel: InboxViewModel) -> InboxViewController {
         guard
-            let currentAuthType = appContext.dataService.currentAuthType,
-            let user = appContext.dataService.currentUser
+            let currentAuthType = appContext.dataService.currentAuthType
```
tomholub, line 16: This is another thing that should be dependency-injected. (separate issue, I suppose)

Also, maybe we should include `user` in the `AppContext`? That should minimize the changes, and clean things up.

sosnovsky: Agree - it seems we pass `user` param along with `appContext` almost each time, it'll be simpler and cleaner to have `user` property included in `AppContext`.

tomholub: And then, for situations that don't need `AppContext`, we'll pass the user alone

tomholub: Although - if we do this, I suppose the user in AppContext would have to be optional, so we are back where we started. Then each controller would have to still double check the user is there.

Could we then have two separate types - `AppContext` and `UserContext` ? The first one would stay as is, and `UserContext` would be required in all VCs that need it?

ivan-ushakov: I replaced several `user` with `UserContext` but it does not look better for me. Also it is hard to use two contexts because we don't have strict flow now like "here we 100% have session, user and auth type", and here not.

FlowCrypt/App/AppContext.swift
```diff
@@ -78,16 +78,25 @@ class AppContext {
...
-    func withSession(_ session: SessionType?) -> AppContext {
-        return AppContext(
-            encryptedStorage: self.encryptedStorage,
+    func withSession(_ session: SessionType?) -> UserContext {
+        guard
+            let authType = dataService.currentAuthType,
+            let user = dataService.currentUser
+        else {
+            fatalError()
+        }
```
tomholub, line 87: These should be dependency injected also, I think?. They should be retrieved only once, somewhere all the way up in router or similar place.

FlowCrypt/App/AppContext.swift
```diff
@@ -131,3 +140,35 @@ class AppContext {
         )
     }
 }
+
+class UserContext: AppContext {
```
tomholub, line 144: This is not what I meant, but it's not bad. Why not call it `AppContextWithUser`. Then inside controllers, you put `appContext: AppContextWithUser` to make the intention clear. I think it's a good compromise - the variable will be called `appContext` everywhere for consitency, and the type `AppContext` vs `AppContextWithUser` will define if there's user or not. So far so good, to my eye.

tomholub (commented):
Actually it doesn't look bad to me. See comments.

FlowCrypt/Controllers/Inbox/InboxViewController+Factory.swift
tomholub: See https://github.com/FlowCrypt/flowcrypt-ios/pull/1331#pullrequestreview-865505229

> don't have strict flow now like "here we 100% have session, user and auth type"

That will be the next step, another PR, another issue - all of that has to be refactored to have exactly that :+1:

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -99,34 +99,31 @@ final class ComposeViewController: TableNodeViewController {
...
             )
         )
         self.cloudContactProvider = cloudContactProvider
         self.composeMessageService = composeMessageService ?? ComposeMessageService(
             clientConfiguration: clientConfiguration,
-            encryptedStorage: appContext.encryptedStorage,
-            messageGateway: appContext.getRequiredMailProvider().messageSender,
-            passPhraseService: appContext.passPhraseService,
-            sender: email
+            encryptedStorage: userContext.encryptedStorage,
```
tomholub, line 123: let's always call it `appContext` for consistency, even if type is either `AppContext` or `AppContextWithUser` (usually)

FlowCrypt/App/AppContext.swift
sosnovsky: I like this idea of `AppContext` subclass for authorised users 👍

So just need to rename it to `AppContextWithUser` and `userContext` variable to `appContext` in view controllers as proposed by @tomholub. Otherwise it looks good to me too.

FlowCrypt/App/AppContext.swift
ivan-ushakov: We have 5 places where we use it: 4 in GlobalRouter and 1 in AppStartup. Also we use `dataService` and the only object that have it is context itself.

tomholub: Sure, but I don't want to have this hidden in here. I'd rather that we do this in `GlobalRouter` and then we refactor it there in the future once we get rid of `dataService`.

It's an eye sore, so putting it in GlobalRouter where it's more visible will make us refactor it later :smirk:

sosnovsky (approved):
Looks great ✅

PR #1336, opened 2022-01-25
Crash when moving email to trash

### LINKED ISSUE #1076: crash when moving email to trash
image

ivan-ushakov:
close #1076

### To be filled by reviewers


FlowCrypt/ObjC/ObjException.m
```diff
@@ -0,0 +1,26 @@
...
+//  ObjException.m
+//  FlowCrypt
+//
+//  Created by  Ivan Ushakov on 25.01.2022
+//  Copyright © 2017-present FlowCrypt a. s. All rights reserved.
+//
+
+#import "ObjException.h"
+
+@implementation ObjException
```
sosnovsky, line 11: It'll be better to call it `ObjcException`, as `ObjException` doesn't make it clear that we're catching exceptions from Objective C.

FlowCrypt/ObjC/ObjException.m
```diff
@@ -0,0 +1,26 @@
...
+
++ (BOOL)catchException:(void(^)(void))tryBlock error:(__autoreleasing NSError **)error {
+    @try
+    {
+        tryBlock();
+        return YES;
+    }
+    @catch (NSException *exception)
+    {
+        *error = [[NSError alloc] initWithDomain:exception.name code:0 userInfo:exception.userInfo];
```
sosnovsky, line 21: There is updated implementation on stackoverflow, which makes error message more  detailed - https://stackoverflow.com/a/61825487/3091318

FlowCrypt/Controllers/Inbox/InboxViewController.swift
```diff
@@ -452,10 +452,22 @@ extension InboxViewController: MsgListViewController {
                 tableNode.reloadData()
             } else {
                 state = .fetched(.byNumber(total: newTotalNumber))
-                tableNode.deleteRows(at: [IndexPath(row: index, section: 0)], with: .left)
+                do {
+                    try ObjException.catch {
+                        self.tableNode.deleteRows(at: [IndexPath(row: index, section: 0)], with: .left)
+                    }
+                } catch {
+                    logger.logError("Failed to remove message at \(index) in fetched state")
```
sosnovsky, line 460: Let's also add catched error message here, so it'll be `"Failed to remove message at \(index) in fetched state: \(error)"`

sosnovsky (changes requested):

FlowCrypt/Controllers/Inbox/InboxViewController.swift
```diff
@@ -452,10 +452,22 @@ extension InboxViewController: MsgListViewController {
                 tableNode.reloadData()
             } else {
                 state = .fetched(.byNumber(total: newTotalNumber))
-                tableNode.deleteRows(at: [IndexPath(row: index, section: 0)], with: .left)
+                do {
+                    try ObjcException.catch {
```
Kharchevskyi, line 456: Can you please explain how is this fix related to the crash fix?

Kharchevskyi: It looks like there is an issue related to the index of the row that we try to update

FlowCrypt/Controllers/Inbox/InboxViewController.swift
```diff
@@ -452,10 +452,22 @@ extension InboxViewController: MsgListViewController {
...
+                        self.tableNode.deleteRows(at: [IndexPath(row: index, section: 0)], with: .left)
```
Kharchevskyi, line 457: Data source for table node should be updated with correct amount of `inboxInput` then reload should be performed.

Kharchevskyi (changes requested):

FlowCrypt/Controllers/Inbox/InboxViewController.swift
ivan-ushakov: As @tomholub said, we want to avoid having crash here. Even if issue is still here, we should not crash.

Kharchevskyi: Can we have an issue to solve the root cause of the crash?

tomholub: absolutely. Let's solve it in this PR as well.

FlowCrypt/Controllers/Inbox/InboxViewController.swift
ivan-ushakov: Data source is updated above:
```
inboxInput.remove(at: index)
```
We could reload table node with `reloadData()`, but crash means that we are in invalid state. We could move from index to some kind of message ID. 

But I guess that this PR is more about handling Objc exceptions in Swift.

FlowCrypt/Controllers/Inbox/InboxViewController.swift
ivan-ushakov: Since all UI is async with Texture, I guess that using message index is unsafe. We should move to message ID, as this is the only way to know that message we want to delete is the same message.

Kharchevskyi: Would suggest to investigate root cause of the issue. 

Also if there was only one message before moving to trash we need to update whole state of the controller.

FlowCrypt/Controllers/Inbox/InboxViewController.swift
tomholub: While originally it was indeed primary about handling of ObjC exceptions, that was in part because we didn't know any better.

If the actual root cause of this exception can be fixed, that is even better and more appropriate. Both should be done.

sosnovsky: I think it's something more specific than just updating data source, as this crash happens very rarely.
Implementing ObjC exception handling should help us to find root cause of this issue.

May be switching from index to message id (as @ivan-ushakov proposed) will fix this crash, but we can't be 100% sure as currently we're not able to reproduce this issue.

tomholub: If the goal is to collect more information about the crash, I suppose we should have it show a modal similar to other crash scenarios? Then some user may take a screenshot and send it to us for debugging.

As currently, it will just be logged then forgotten.

sosnovsky (approved):
I think it's a good solution for now - app won't crash if this issue occurs again and users will get alert with error message.

ivan-ushakov:
@Kharchevskyi are you agree with current approach? We will try to track down the cause of the crash with that alerts and also prevent unhandled exception to crash app.

Kharchevskyi:
@ivan-ushakov sure, no objections 👍

Kharchevskyi (approved):

sosnovsky (approved):

PR #1340, opened 2022-01-26
Focus on recipients when forwarding

### LINKED ISSUE #1085: focus on recipients when forwarding
Functionality works well. One improvement: when forwarding (unlike reply) the recipients field should be in automatic focus when the compose view renders. When replying, the focus should be on message body as it does now.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1066#pullrequestreview-815218082_

Kharchevskyi:
This PR apply focus when forwardinf

close #1085

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Compose/ComposeViewControllerInput.swift
+++ b/FlowCrypt/Controllers/Compose/ComposeViewControllerInput.swift
@@ -25,18 +25,52 @@ struct ComposeMessageInput: Equatable {
 
     enum InputType: Equatable {
         case idle
-        case quote(MessageQuoteInfo)
+        case reply(MessageQuoteInfo)
+        case forward(MessageQuoteInfo)
     }
 
     let type: InputType
 
-    var isQuote: Bool {
+    var quoteRecipients: [String] {
+        guard case .reply(let info) = type else {
+            return []
+        }
+        return info.recipients
+    }
+
+    var subject: String? {
+        type.info?.subject
+    }
+
+    var replyToMime: Data? {
+        type.info?.mime
+    }
+
+    var threadId: String? {
+        type.info?.threadId
+    }
+
+    var attachments: [MessageAttachment] {
+        type.info?.attachments ?? []
+    }
+}
+
+extension ComposeMessageInput {
+    var successfullySentToast: String {
         switch type {
         case .idle:
-            return false
-        case .quote:
-            return true
+            return "compose_encrypted_sent".localized
+        case .forward:
+            return "compose_forward_successful".localized
+        case .reply:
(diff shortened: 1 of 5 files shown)


FlowCrypt/Controllers/Compose/ComposeViewControllerInput.swift
```diff
@@ -31,8 +32,42 @@ struct ComposeMessageInput: Equatable {
...
+        switch type {
+        case .idle:
+            return false
+        case .quote(let messageQuoteInfo):
+            switch messageQuoteInfo.quoteType {
+            case .reply:
+                return false
+            case .forward:
+                return true
+            }
```
sosnovsky, line 70: Current implementation with 3 different properties `isIdle`, `isReply` and `isForward` looks a bit complicated.

I think updating `InputType` to look like this will make code more clear
```swift
enum InputType: Equatable {
    case idle
    case reply(MessageQuoteInfo)
    case forward(MessageQuoteInfo)
}
```
Then `isQuote` will be just
```swift
var isQuote: Bool {
    return type != .idle
 }
```
and we won't need complex `switch` for checking other types.

sosnovsky (changes requested):

FlowCrypt/Controllers/Compose/ComposeViewControllerInput.swift
Kharchevskyi: Completely agree with solution 
```
enum InputType: Equatable {
    case idle
    case reply(MessageQuoteInfo)
    case forward(MessageQuoteInfo)
}
```
I thought it was done intentionally. Will apply this changes

sosnovsky: Previously we didn't need to separate `forward` and `reply` for this screen, so `quote` type was enough.
But for this functionality and for possible future features it'll be better to have separate `InputType` cases for each type.

FlowCrypt/Controllers/Compose/ComposeViewControllerInput.swift
```diff
@@ -12,6 +12,7 @@ struct ComposeMessageInput: Equatable {
     static let empty = ComposeMessageInput(type: .idle)
 
     struct MessageQuoteInfo: Equatable {
+        let quoteType: MessageQuoteType
```
sosnovsky, line 15: With updated `InputType` enum we don't need `quoteType` property here

sosnovsky (changes requested):

sosnovsky (approved):
✅

Issue #1327, opened 2022-01-21
inconsistency between contact trash icon and public key trash icon result


tomholub:
This one is supposed to delete whole contact + pubkey, as you say:

![image](https://user-images.githubusercontent.com/6306961/150513550-d77a969f-02a6-460d-8f0b-5d9f8de896de.png)

This one is supposed to only delete pubkey but leave the contact in:

![image](https://user-images.githubusercontent.com/6306961/150513628-b04ebce3-55e6-41a7-9bca-2307914bee01.png)

You can tap either of the two. The result is that if you go back to compose screen, it should refetch the key and be green.

Did I understand it correctly that you are seeing it go gray in stage 4, when using `delete only pubkey` icon?

But it shows green when using `delete whole contact` icon?

That would be a bug, I think. Please confirm. The two are supposed to work the same when you go to compose screen.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1301#issuecomment-1018392512_

fcvakintos:
@tomholub i checked using my account and after deleting only pubkey - the account is displayed grey in compose and new pub key is present in contact details

https://user-images.githubusercontent.com/91205853/150751087-6a50e965-8f7a-4ba4-a5ae-abe487e025b5.mov

tomholub:
Thank you for the video, this is a bug for sure. I'm glad you mentioned it.

tomholub:
Interestingly, after you go back to Settings -> contacts, it seems that it did save the new public key. But it didn't render it. I think, if you went back to compose one more time and added that email one more time, it would likely be green.

The bug that needs to be fixed is that when the email is entered after deleting the public key, it's supposed to be gray at first but then it should turn green as soon as the public keys are re-fetched.

Similarly, if we only had an orange (expired) public key and we search for it again, it should be orange at first then turn gray once the pubkey is refetched.

tomholub:
@sosnovsky once you're done with 0.7.0 issues, please look at this. You can test manually, and @fcvakintos will write an automated test for it.

Issue #1120, opened 2021-11-30
wrong ordering of items in inbox


tomholub:
visible on e2e account

@tomholub check this pls
<img width="403" alt="Screenshot 2021-11-30 at 13 30 03" src="https://user-images.githubusercontent.com/91205853/144039725-2d2ff265-fd19-4504-a00c-97ef5920b1b1.png">

_Originally posted by @fcvakintos in https://github.com/FlowCrypt/flowcrypt-ios/issues/959#issuecomment-982549799_

should be ordered by date descending

sosnovsky:
As I found - ordering here is right (it's the same as in web-version of Gmail), but Gmail API returns wrong value for `internalDate` of message object. I checked `GTLRGmail_Message` from API response, but didn't find right date in any property, so it'll need further investigation.

<img width="784" alt="Screenshot 2022-01-28 at 17 22 52" src="https://user-images.githubusercontent.com/1062093/151574002-84316175-2075-43e8-827d-736d215d59b1.png">

<img width="315" alt="Screenshot 2022-01-28 at 17 23 27" src="https://user-images.githubusercontent.com/1062093/151573987-2bc18717-c357-4a77-a078-972ae91bd074.png">

tomholub:
> As I found - ordering here is right (it's the same as in web-version of Gmail), but Gmail API returns wrong value for internalDate of message object. I checked GTLRGmail_Message from API response, but didn't find right date in any property, so it'll need further investigation.

So currently the order matches the order on webmail? But date renders wrong.

See what "encrypted inbox" in browser extension does, where we render it ourselves from API (browser action button -> "encrypted inbox"). If that one is correct, see what typescript does to get it correct.

sosnovsky:
> So currently the order matches the order on webmail? But date renders wrong.

Yes, order is right, but date is wrong for some messages.

`encrypted inbox` in browser extension has the same issue:

<img width="706" alt="Screenshot 2022-01-28 at 22 32 57" src="https://user-images.githubusercontent.com/1062093/151617416-ae65d0f9-0985-4d5e-b461-bda5d904c66e.png">

tomholub:
got it. A possible Gmail API bug.

Issue #954, opened 2021-11-08
adding cc and bcc recipients


tomholub:
currently missing on compose screen

Kharchevskyi:
@tomholub can I pick up this issue? or can you assign some issues for me?

tomholub:
I'll assign some other issues - one sec.

PR #1342, opened 2022-01-27
#1253 Disable password-protected messages when FES is not used

### LINKED ISSUE #1253: disable password-protected messages if FES is not used

sosnovsky:
This PR enables password-protected messages only when FES is used

close #1253

- Tests added or updated

### To be filled by reviewers


sosnovsky:
I think that after finishing and merging https://github.com/FlowCrypt/flowcrypt-ios/pull/1331 we should also update `EnterpriseServerApi` to have `email` property (which will be passed from `AppContextWithUser`), so we don't need to pass it manually for every method.
Then we'll be able to fetch `fesUrl` only once and store result locally, while currently it's fetched for each request.

sosnovsky:
@tomholub I have 2 remaining tasks - https://github.com/FlowCrypt/flowcrypt-ios/issues/1327 from `0.8.0` milestone and https://github.com/FlowCrypt/flowcrypt-ios/issues/1120 from `0.9.0`
Should I also take some other tasks from `0.8.0` like https://github.com/FlowCrypt/flowcrypt-ios/issues/953 or https://github.com/FlowCrypt/flowcrypt-ios/issues/954 ?

FlowCrypt/Functionality/Services/Client Configuration Service/ClientConfigurationService.swift
```diff
@@ -34,7 +34,8 @@ extension ClientConfigurationService: ClientConfigurationServiceType {
     func fetch(for user: User) async throws -> ClientConfiguration {
         do {
             let raw = try await server.getClientConfiguration(for: user.email)
-            try local.save(for: user, raw: raw)
+            let fesUrl = try await server.getActiveFesUrl(for: user.email)
+            try local.save(for: user, raw: raw, fesUrl: fesUrl)
             return ClientConfiguration(raw: raw)
```
tomholub, line 39: from my memory, `getClientConfiguration` also calls `getActiveFesUrl` separately, so we are calling twice. Can we optimize that? 

`getActiveFesUrl` is to figure out if FES is running there or not (once discovered at least once as running, we should save that and always expect it going forward)

`getClientConfiguration` is to fetch the actual config, but it should have to call `getActiveFesUrl` each time if we already saved it as existing at least once.

sosnovsky: Yes, it'll be fixed by https://github.com/FlowCrypt/flowcrypt-ios/issues/1348 - `fesUrl` will be fetched only once and then reused.

tomholub: ok

tomholub (approved):

PR #1345, opened 2022-01-27
Show file name extension when saving

### LINKED ISSUE #901: show attachment name extension when downloading - popup
The name of the image was `image.png` but the popup only shows `image`. Could it show `image.png` ?

Screenshot 2021-11-03 at 12 31 33

Kharchevskyi:
This PR add file name extension to `UIDocumentPickerViewController` when saving

close #901 

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Common UI/AttachmentManager.swift
+++ b/FlowCrypt/Common UI/AttachmentManager.swift
@@ -46,6 +46,7 @@ final class AttachmentManager: NSObject {
     @MainActor
     private func openDocumentsController(from url: URL) {
         let documentController = UIDocumentPickerViewController(forExporting: [url])
+        documentController.shouldShowFileExtensions = true
         documentController.delegate = self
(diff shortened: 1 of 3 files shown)


sosnovsky (commented):
It's a bit hacky way to show file extensions - we're just duplicating extension to make it visible in iOS Files app. Then if we'll check file details - it's extension will be duplicated:

Simulator Screen Shot - iPhone 13 - 2022-01-27 at 23 09 53

I checked other apps (including Gmail) - they use default iOS behavior and not showing file extensions when saving attachments.

@tomholub may be we should leave it as it is, without visible file extensions?

tomholub:
> I checked other apps (including Gmail) - they use default iOS behavior and not showing file extensions when saving attachments.

> @tomholub may be we should leave it as it is, without visible file extensions?

Agree, if it's the default and common behavior on iOS, no point trying to go against it.

PR #1347, opened 2022-01-28
#1327 Fix public key refetch

### LINKED ISSUE #1327: inconsistency between contact trash icon and public key trash icon result

sosnovsky:
This PR fixes recipients keys refetch after manual deletion in Settings -> Contacts

close #1327

### To be filled by reviewers


FlowCrypt/Functionality/Services/Local Pub Key Services/ContactsService.swift
```diff
@@ -43,15 +43,8 @@ struct ContactsService: ContactsServiceType {
...
-            let recipient = try await pubLookup.lookup(email: email)
-            try localContactsProvider.save(recipient: recipient)
-            return recipient
-        }
-
-        let recipient = try await pubLookup.lookup(email: email)
-        try localContactsProvider.updateKeys(for: recipient)
+        let contact = try await pubLookup.lookup(email: email)
+        try localContactsProvider.updateKeys(for: contact)
         return contact
```
tomholub, line 48: This way, we are never actually using the local contacts, always only saving them. Also, wouldn't the verification of signed messages now be forcibly delayed even if we already have the right public keys?

I'd need a different fix that is more complicated but more performant, as follows (from the issue)

> The bug that needs to be fixed is that when the email is entered after deleting the public key, it's supposed to be gray at first but then it should turn green as soon as the public keys are re-fetched.

> Similarly, if we only had an orange (expired) public key and we search for it again, it should be orange at first then turn gray once the pubkey is refetched.

When we already have a contact for a recipient, the app should render the result from the local contacts first, and then fire off a background method to re-fetch them from remote. The full sequence would be:

1) user enters recipient email
2) pull from local contact. Is there any available?
  - there is no contact available:
      3. fetch remote contact, update local contacts, render result. done
  - there is contact already available:
      3. render result based on existing contact right away
      4. fetch updated public keys in background, update local contacts, render result again, done

The effect is that in the scenario you were fixing, the recipient should turn immediately gray and than _change_ to green.

-------------------

Another problem with this code, already present on master, is that this is called `Local Pub Key Services` but they also fetch from remote. That's confusing. I guess we need three categories `Local Pub KeyServices`, `Remote Pub Key Services` and `Combined Pub Key Services` if we need this kind of method somewhere here for reuse.

To make the distinctions clear.

tomholub (commented):
see below

FlowCrypt/Functionality/Services/Local Pub Key Services/ContactsService.swift
sosnovsky: > This way, we are never actually using the local contacts, always only saving them. Also, wouldn't the verification of signed messages now be forcibly delayed even if we already have the right public keys?

It seems current `master` works the same way - app refreshes contact keys even when we have stored ones (https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/Functionality/Services/Local%20Pub%20Key%20Services/ContactsService.swift#L53). So we shouldn't see any additional delays with current fix.

But your explanation looks like more complete solution and will make contacts and messages verifications faster 👍

> Another problem with this code, already present on master, is that this is called Local Pub Key Services but they also fetch from remote. That's confusing. I guess we need three categories Local Pub KeyServices, Remote Pub Key Services and Combined Pub Key Services if we need this kind of method somewhere here for reuse.

What about such structure as on screenshot below? I renamed `Local Pub KeyServices` to `Local Contacts Service` as we have `LocalContactsProvider` there (it provides not only pub keys, but contacts as well). And I also added `Contacts Service` group which includes models used by both (local and remote) providers and `ContactsService` which fetches local and remote contacts.

Screenshot 2022-01-28 at 15 36 22

tomholub: It's not 100% there but it's better than now. Maybe in the future we'll come up with something as we refactor over time.

sosnovsky:
I updated code to use new logic - recipient UI is rendered after retrieving local contact and it's updated if new keys were fetched in background:

https://user-images.githubusercontent.com/1062093/151822415-06e09140-71fa-4ed2-8d82-1cbc1c171cd7.mp4

tomholub (approved):
:+1:

PR #1354, opened 2022-02-01
added new spec for checking user color after removing pubkey for it

### LINKED ISSUE #1327: inconsistency between contact trash icon and public key trash icon result

### LINKED ISSUE #1353: test deleting public without deleting contact, then composing
@sosnovsky once you're done with 0.7.0 issues, please look at this. You can test manually, and @fcvakintos will write an automated test for it.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1327#issuecomment-1020006720_

fcvakintos:
This PR contains new spec for checking user color on compose email after removing pubkey for it, and checking pubkey after compose for contact with deleted pubkey

close #1353 

**Tests** _
- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/appium/tests/specs/live/composeEmail/CheckRecipientColorAfterRemovingPubKey.spec.ts
+++ b/appium/tests/specs/live/composeEmail/CheckRecipientColorAfterRemovingPubKey.spec.ts
@@ -1,0 +1,89 @@
+import {
+  SplashScreen,
+  SetupKeyScreen,
+  MailFolderScreen,
+  NewMessageScreen,
+  ContactScreen,
+  ContactPublicKeyScreen,
+  SettingsScreen,
+  MenuBarScreen,
+  PublicKeyDetailsScreen
+} from '../../../screenobjects/all-screens';
+
+import { CommonData } from '../../../data';
+
+describe('COMPOSE EMAIL: ', () => {
+
+  it('check recipient color after removing public key from settings', async () => {
+
+    const contactEmail = CommonData.contact.email;
+    const contactName = CommonData.contact.name;
+
+    await SplashScreen.login();
+    await SetupKeyScreen.setPassPhrase();
+    await MailFolderScreen.checkInboxScreen();
+
+    // Add first contact
+    await MailFolderScreen.clickCreateEmail();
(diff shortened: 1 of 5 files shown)


appium/tests/specs/live/composeEmail/CheckRecipientColorAfterRemovingPubKey.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    await ContactScreen.checkContactScreen();
+    await ContactScreen.checkContact(contactEmail);
+
+    await ContactScreen.clickOnContact(contactEmail);
+    await ContactPublicKeyScreen.checkPgpUserId(contactEmail);
+    await ContactPublicKeyScreen.checkPublicKeyDetailsNotEmpty();
+    await ContactPublicKeyScreen.clickOnFingerPrint();
+
+    await PublicKeyDetailsScreen.checkPublicKeyDetailsScreen();
+    await PublicKeyDetailsScreen.checkPublicKeyNotEmpty();
```
sosnovsky, line 87: Lines `27-49` and `65-87` are the same, can we move them to separate helper method? (for example `addRecipientAndCheckFetchedKey(name, email)`)

sosnovsky (changes requested):

appium/tests/specs/live/composeEmail/CheckRecipientColorAfterRemovingPubKey.spec.ts
fcvakintos: done

sosnovsky (approved):
👍

Issue #1346, opened 2022-01-28
Missing private key packets validation when importing/retrieving private key


martgil:
FlowCrypt recently received a security report wherein a bad private key is not getting blocked when importing/retrieving the private key. eg retrieving the key from backup (inbox), importing a private key file.

Sample bad private key:
[corrupted-rsa-key.asc.txt](https://github.com/FlowCrypt/flowcrypt-ios/files/7955539/corrupted-rsa-key.asc.txt)

Steps to reproduce:
1. The easiest way to reproduce it is by importing the private key file from the app.

error message from gpg:
![image](https://user-images.githubusercontent.com/46025304/151485210-f97bb05e-b401-4d8f-8a81-270b0fd68a0b.png)

For reference to how this issue gets fixed on the browser extension, please see FlowCrypt/flowcrypt-browser#4271.

Impact:
As stated from the original email:
> In particular, given that encrypting the private key only encrypts (and authenticates) the private key parameters, but not the public key parameters, an attacker could overwrite the public key. This can lead to various vulnerabilities, and we realized some of these vulnerabilities might be applicable to FlowCrypt.

tomholub:
@martgil could you please write a (failing) test case for this here? https://github.com/FlowCrypt/flowcrypt-ios/blob/master/Core/source/test.ts

martgil:
sir @tomholub sorry for the late reply. I will try to, sir. I'm currently having a minor issue on FlowCrypt/flowcrypt-ios/pull/1357 but will try to settle the errors and hopefully achieve the same result as `FlowCrypt/flowcrypt-browser#4271`

PR #1357, opened 2022-02-02
add private key validation test

martgil:
This PR will add a private key validation test and update PgpKey class

close https://github.com/FlowCrypt/flowcrypt-ios/issues/1346 // if this PR closes an issue

Todo:
- [ ] validate private key in `readMany()` of pgp-keys.ts
- [ ] eliminate errors `TypeError { message: 'prvPacket.validate is not a function' }`
- [ ] complete the test for `unencrypted private key`, `encrypted private key` validation.  

reference: https://github.com/FlowCrypt/flowcrypt-browser/pull/4271

- Tests added or updated

### To be filled by reviewers


martgil:
@IvanPizhenko sorry for any disturbance, sir. I would like to ask for any hints?

I'm trying to achieve the same way that sir, Roman did on `FlowCrypt/flowcrypt-browser#4271` when validating the key. 
I keep getting the error of `prvPacket.validate is not a function` even though I've added the `.validate(): Promise<any>;` on `openpgp.d.ts` type definition file. I feel like I'm still missing something but not sure exactly what it is.

I also tried to find any references for the `.validate()` but I only get results from the `openpgp.d.ts` and `openpgp-keys.ts(browser extension) / pgpkey.ts(ios-repo)`

Any feedback would be much appreciated 🙇🏻‍♂️ Thanks!

IvanPizhenko:
@martgil `.d.ts` file is just type definitions for the Typescript compiler, it does not provide function implementation. I think browser extension may use modified version of openpgp.js with such custom function added. You better ask Roman about this.

tomholub:
Likely, flowcrypt-ios uses an older version of OpenPGP.js than browser extension does.

@IvanPizhenko we could update the iOS OpenPGP.js to v5 (instead of FES, initially), what do you think? It won't be an easy task.

IvanPizhenko:
@tomholub yes, we can try to do it.

PR #1360, opened 2022-02-03
Move strings to localiztion files

### LINKED ISSUE #956: all texts to be in separate strings files
currently most are, but I think there were some alerts that still had hardcoded texts

we may need to do a language translation, this will help

Kharchevskyi:
This PR moves texts to separate strings files.
* Add missed localizations
* `@inline(__always) func localizeWithArguments(_ arguments: String...) -> String` changed to accept only strings arguments to prevent crashes in runtime
* Enable showNonLocalizedStrings for debug scheme

close #956

### To be filled by reviewers


sosnovsky:
@Kharchevskyi many ui tests are failing after update - https://flowcrypt.semaphoreci.com/jobs/3d7e754c-290c-46fd-a54f-1c16eb60eb66/summary?test-result-id=4ae71336-e44b-39bf-b9d2-752e234818a5

Kharchevskyi:
@sosnovsky thanks, already on it. Checking

FlowCrypt/Controllers/Setup/SetupManuallyImportKeyViewController.swift
```diff
@@ -171,7 +171,10 @@ extension SetupManuallyImportKeyViewController {
         if privateKey.isEmpty {
             userInfoMessage = "import_no_backups_clipboard".localized + user
         } else {
-            userInfoMessage = "Found \(privateKey.count) key\(privateKey.count > 1 ? "s" : "")"
+            let key = privateKey.count > 1
```
sosnovsky, line 174: For such cases it's better to use `stringsdict` with pluralisation support - it'll automatically use appropriate localization for provided parameter (`privateKey.count`) - https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/Resources/Localizable.stringsdict

FlowCrypt/Controllers/Setup/SetupViewDecorator.swift
```diff
@@ -74,9 +74,10 @@ struct SetupViewDecorator {
 
         switch subtitleType {
         case let .fetchedKeys(count):
-            subtitle = "Found \(count) key backup\(count > 1 ? "s" : "")"
+            let key = count > 1 ? "setup_fetched_keys" : "setup_fetched_key"
```
sosnovsky, line 77: `stringsdict` should be used here too

sosnovsky (changes requested):
Looks good, just need to fix ui tests and use `stringsdict` for pluralisation support

FlowCrypt/Controllers/Setup/SetupManuallyImportKeyViewController.swift
Kharchevskyi: Why I do not prefer this solution because `NSLocalizedString` accepts array of CVarArg as arguments and in case we pass string instead of int we will get a crash in runtime

As you see to prevent that I have changed implementation for localizeWithArguments to accept only stings in code and all parameters in `localization.strings` accepts only `%@` as String parameter
```
@inline(__always) func localizeWithArguments(_ arguments: String...)
``` 

With `stringsdict` solution we need to explicitly pass Int as a variable to be able to distinguish is it variable plural. 
Do we really want to have this possibility?

Kharchevskyi: @sosnovsky 
What do you think if I leave only Strings as parameters here in `localizeWithArguments` implementation? And use only strings in localizable files 

And will add a separate `func localizePluralWithArguments` to accept `Int` and use stringsdict?

sosnovsky: > @sosnovsky What do you think if I leave only Strings as parameters here in `localizeWithArguments` implementation? And we will use only string as a parameter in localizable files
> 
> And will add a separate `func localizePluralWithArguments` to accept `Int` and use stringsdict?

Sounds good for me, let's do it this way 👍

sosnovsky (approved):

PR #1371, opened 2022-02-08
0.6.0 testflight feedback crashes

### LINKED ISSUE #1365: 0.6.0 testflight feedback crashes
These happened on 0.6.0 release

[testflight_feedback.zip](https://github.com/FlowCrypt/flowcrypt-ios/files/8006564/testflight_feedback.zip)

[testflight_feedback2.zip](https://github.com/FlowCrypt/flowcrypt-ios/files/8006565/testflight_feedback2.zip)

[testflight_feedback_3.zip](https://github.com/FlowCrypt/flowcrypt-ios/files/8006563/testflight_feedback_3.zip)

ivan-ushakov:
close #1365

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/App/AppContext.swift
+++ b/FlowCrypt/App/AppContext.swift
@@ -48,7 +48,7 @@ class AppContext {
         let keyService = KeyService(
             storage: encryptedStorage,
             passPhraseService: passPhraseService,
-            currentUserEmail: { encryptedStorage.activeUser?.email }
+            currentUserEmail: { try? encryptedStorage.activeUser?.email }
         )
         let clientConfigurationService = ClientConfigurationService(
             local: LocalClientConfiguration(
@@ -58,10 +58,10 @@ class AppContext {
         return AppContext(
             encryptedStorage: encryptedStorage,
             session: nil, // will be set later. But would be nice to already set here, if available
-            userAccountService: SessionService(
+            userAccountService: try SessionService(
                 encryptedStorage: encryptedStorage,
                 googleService: GoogleUserService(
-                    currentUserEmail: encryptedStorage.activeUser?.email,
+                    currentUserEmail: try encryptedStorage.activeUser?.email,
                     appDelegateGoogleSessionContainer: UIApplication.shared.delegate as? AppDelegate
                 )
             ),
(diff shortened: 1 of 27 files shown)


ivan-ushakov:
@tomholub 
Are you OK with having such amount of `throw` and `try` in code?

sosnovsky:
I think this implementation won't fix initial issue - currently `var storage: Realm` is a computed property (https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/Functionality/DataManager/Encrypted%20Storage/EncryptedStorage.swift#L71) and it's re-initialized each time when app uses `encryptedStorage.storage`.
So possibly we have crashes when app tries to perform multiple Realm requests simultaneously.

My proposal here is to initialize `let storage: Realm` only once in the `init` of `EncryptedStorage` and then check Testflight crash logs to see if these crashes are gone or still there.

tomholub:
> My proposal here is to initialize `let storage: Realm` only once in the `init` of `EncryptedStorage` and then check Testflight crash logs to see if these crashes are gone or still there.

:+1: I've additionally still been seeing the "red button screen of death" (storage corrupted) view and had to reset storage just today. This may be one of the reasons. We should only initialize it once.

sosnovsky:
> It is possible but since Realm attaches to thread and we have different places where we use Realm, we have use one thread. So it should be either actor or queue and async facade, like we did for Google libraries. This could affect performance.

Yeah, you're right - it'll be more complex than just re-initializing Realm on the same thread as current code.

But maybe crashes from #1365 are already fixed by some other changes - I checked Testflight logs from Xcode and the last recorded crash was on February 1. Then on February 2 we merged #1331 to `master` and no crashes with `EncryptedStorage.storage` happened since then.

Also these crashes were related to `DataService.currentUser` getter, but we removed `DataService` in yesterday's PR, it can help to avoid this Realm issue too.

However I noticed new crashes which appeared in `0.7.0` (attaching crashlog), I think we should find what causes them and check if they're related to Realm initialisation too.

[testflight_feedback.zip](https://github.com/FlowCrypt/flowcrypt-ios/files/8036438/testflight_feedback.zip)

tomholub:
> I checked Testflight logs from Xcode and the last recorded crash was on February 1. Then on February 2 we merged https://github.com/FlowCrypt/flowcrypt-ios/pull/1331 to master and no crashes with EncryptedStorage.storage happened since then.

Testflight crashes are not affected by what gets pushed to master, only by what is sent to apple for review (and what users end up using). I only upload new versions to apple around the time I make a versioned commit on master.

Therefore, the lack of reports is not caused by any merged PR around that time. More likely, users that got crashes walked away and didn't try again, or the crash self-recovered on next app startup - hard to say.

tomholub:
ok

sosnovsky (approved):

PR #1374, opened 2022-02-09
issue #1362 Upgrade to openpgp.js v5

### LINKED ISSUE #1362: update OpenPGP.js to v5
Likely, flowcrypt-ios uses an older version of OpenPGP.js than browser extension does.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1357#issuecomment-1028299363_

IvanPizhenko:
This PR upgrades OpenPGP.js to v 5.1.0

close #1362

- Tests added or updated

### To be filled by reviewers


IvanPizhenko:
This is a bit surprising, but still the fact - all CI tests passed.

IvanPizhenko:
@tomholub Ready for review

IvanPizhenko:
Converted back to draft, since there are some discussions on openpgp.

tomholub (commented):
Still need to do a proper review first. Would it be possible to drop `lib/openpgpjs-v5` directory and instead rely directly on imports from NodeJS? Just like any other library. It will make updates easier. We could still have our own `.d.ts` file if necessary (before v5 definitions get fixed)

Core/source/core/pgp-hash.ts
```diff
@@ -2,17 +2,17 @@
...
+import { util, Sha1, Sha256 } from 'openpgp';
 import { Buf } from './buf';
-import { openpgp } from './pgp';
 
 export class PgpHash {
 
   public static sha1UtfStr = async (string: string): Promise<string> => {
-    return openpgp.util.Uint8Array_to_hex(await openpgp.crypto.hash.digest(openpgp.enums.hash.sha1, Buf.fromUtfStr(string)));
+    return util.uint8ArrayToHex((new Sha1()).process(Buf.fromUtfStr(string)).finish().result);
   }
```
tomholub, line 12: is `sha1UtfStr` used anywhere in the library? I see `doubleSha1Upper`. Is that one used? If not, both can be removed.

Core/source/core/pgp-key.ts
```diff
@@ -172,21 +174,23 @@ export class PgpKey {
       throw new Error(`No private key packets in key to encrypt. Is this a private key?`);
     }
     if (encryptedPacketCount) {
-      throw new Error(`Cannot encrypt a key that has ${encryptedPacketCount} of ${secretPackets.length} private packets still encrypted`);
+      throw new Error(`Cannot encrypt a key that has ${encryptedPacketCount} of ` +
+        `${secretPackets.length} private packets still encrypted`);
     }
-    await prv.encrypt(passphrase);
+    await ((prv as unknown) as BaseSecretKeyPacket).encrypt(passphrase);
```
tomholub, line 180: This looks wrong - `prv` here should be of a `Key` type. `BaseSecretKeyPacket` is not of `Key` type, `PrivateKey` would be appropriate.

Core/source/core/pgp-key.ts
```diff
@@ -297,9 +302,9 @@ export class PgpKey {
     return Boolean(await key.getEncryptionKey(undefined, oneSecondBeforeExpiration));
   }
 
-  public static dateBeforeExpiration = async (key: OpenPGP.key.Key | string): Promise<Date | undefined> => {
+  public static dateBeforeExpiration = async (key: Key | string): Promise<Date | undefined> => {
     const openPgpKey = typeof key === 'string' ? await PgpKey.read(key) : key;
-    const expires = await openPgpKey.getExpirationTime('encrypt');
+    const expires = await openPgpKey.getExpirationTime();
```
tomholub, line 307: Is this no longer supported?

Core/source/core/pgp-key.ts
```diff
@@ -311,12 +316,13 @@ export class PgpKey {
...
 
-  public static details = async (k: OpenPGP.key.Key): Promise<KeyDetails> => {
+  public static details = async (k: Key): Promise<KeyDetails> => {
     const keys = k.getKeys();
-    const algoInfo = k.primaryKey.getAlgorithmInfo();
-    const algo = { algorithm: algoInfo.algorithm, bits: algoInfo.bits, curve: (algoInfo as any).curve, algorithmId: openpgp.enums.publicKey[algoInfo.algorithm] };
-    const created = k.primaryKey.created.getTime() / 1000;
-    const exp = await k.getExpirationTime('encrypt');
+    const algoInfo = k.keyPacket.getAlgorithmInfo();
+    const algo = { algorithm: algoInfo.algorithm, bits: algoInfo.bits, curve: (algoInfo as any).curve,
```
tomholub, line 322: `(algoInfo as any)` could have been due to lousy types before. Maybe this can be improved now? Could be a separate PR, or here.

Core/source/core/pgp-key.ts
```diff
@@ -311,12 +316,13 @@ export class PgpKey {
...
+      algorithmId: enums.publicKey[algoInfo.algorithm] };
+    const created = k.keyPacket.created.getTime() / 1000;
+    const exp = await k.getExpirationTime();
```
tomholub, line 325: Similarly expiration here - to track until discussed

Core/source/core/pgp-key.ts
```diff
@@ -349,32 +355,36 @@ export class PgpKey {
...
     }
-    allSignatures.sort((a, b) => b.created.getTime() - a.created.getTime());
+    allSignatures.sort((a, b) => (b.created ? b.created.getTime() : 0) - (a.created ? a.created.getTime() : 0));
     const newestSig = allSignatures.find(sig => sig.verified === true);
     if (newestSig) {
-      return newestSig.created.getTime();
+      return newestSig.created ? newestSig.created.getTime() : 0;
     }
     throw new Error('No valid signature found in key');
   }
```
tomholub, line 375: Highlighting to keep track

tomholub (commented):
Only got through these files so far:

image

Core/source/core/pgp-key.ts
IvanPizhenko: fixed it

Core/source/core/pgp-hash.ts
IvanPizhenko: removed it

Core/source/core/pgp-key.ts
IvanPizhenko: fixed

Core/source/core/pgp-key.ts
IvanPizhenko: Need to ask maintainers

IvanPizhenko: Asked question https://github.com/openpgpjs/openpgpjs/discussions/1493

Core/source/core/pgp-key.ts
IvanPizhenko: @tomholub Please read carefully this one https://github.com/openpgpjs/openpgpjs/discussions/1491#discussioncomment-2149242
What do you think of special cases mentioned there?

IvanPizhenko: I've implemented new version of the `lastSig()`. Asked for review from openpgp maintainers.

Core/source/core/pgp-key.ts
IvanPizhenko: At least there is no more such parameter.

IvanPizhenko: @tomholub I've backported old function from openpgp v4

Core/source/core/pgp-key.ts
IvanPizhenko: @tomholub I've backported old function from openpgp v4

IvanPizhenko:
@tomholub I've made number of fixes to address points you've indicated so far, so I think you can continue review of it.

tomholub:
> UPDATE: I've implemented new version of the lastSig() and changed the way we patch openpgp.js. Although it is not exactly using OOB one, we don't need to store our own version of the library code in the our repo, since patch can be applied automatically by the build script. Still we need to distribute somehow modified version, because it is LGPL.

Since the patch is out in the open here on public repo, I believe that counts as distributing and therefore compatible with LGPL.

If we only did it on a closed repo, then we'd have to think about it more.

Core/source/mobile-interface/validate-input.ts
```diff
@@ -178,12 +178,7 @@ const hasProp = (v: Obj, name: string, type: 'string[]' | 'string[]?' | 'object'
...
 export const readArmoredKeyOrThrow = async (armored: string) => {
-  const { keys: [key], err } = await openpgp.key.readArmored(armored);
-  if (err && err.length && err[0] instanceof Error) {
-    throw err[0];
-  }
-  if (!key) {
-    throw new Error('No key found');
-  }
+  const key = await readKey({armoredKey: armored});
+  if (!key) throw new Error('No key found');
```
tomholub, line 182: Please use `{ }` instead of implied blocks

Core/source/platform/require.ts
```diff
@@ -2,20 +2,6 @@
...
-  if (typeof openpgp !== 'undefined') {
-    // @ts-ignore;
-    return openpgp; // self-contained node-mobile
-  }
-  // running tests on a desktop os node instance
-  // making the require semi-dynamic to surpress Webpack warnings/errors. This line does not rely on webpack at all
-  // if this was webpack, then the `openpgp` variable would be already set, and it would never get down here
-  return require(`${'../../../source/lib/openpgp'}`); // points to flowcrypt-mobile-core/source/lib/openpgp.js
-};
-
```
tomholub, line 18: We may still need to do some magic later for it to work within iOS. Hopefully, with the new import mechanism, less magic than before - that is yet to be tested by building this into the iOS assets.

Core/source/platform/util.ts
```diff
@@ -30,3 +31,80 @@ export const setGlobals = () => {
 export const iso2022jpToUtf = (content: Buf) => {
   return dereq_encoding_japanese.convert(content, { to: 'UTF8', from: 'JIS', type: 'string' });
 }
+
+/**
+ * Create hex string from a binary
+ * @param {String} str String to convert
+ * @returns {String} String containing the hexadecimal values
+ */
+export const str_to_hex = (str: string): string => {
```
tomholub, line 40: Is this from OpenPGP.js v4? Should mark each method that is largely from elsewhere with a comment and a link to where can the original be found on GitHub.

Core/source/platform/util.ts
```diff
@@ -30,3 +31,80 @@ export const setGlobals = () => {
...
+}
+
+const getSubkeyExpirationTime = (subkey: Subkey): number | Date => {
+  const bindingCreated = maxDate(subkey.bindingSignatures.map(b => b.created));
+  const binding = subkey.bindingSignatures.filter(b => b.created === bindingCreated)[0];
+  return binding.getExpirationTime();
+}
+
+// Attempt to backport from openpgp.js v4
+export const getKeyExpirationTimeForCapabilities = async (
```
tomholub, line 75: Thank you. I think it would be appropriate to review our code and ensure we need this functionality. I think we do, to make sure the key we encrypt for is not expired. It depends on how OpenPGP.js v5 handles this by default:
 - if their `getExpirationTime` looks at longes expiration regardless of usage, then that's not good - it will blow up when actually trying to encrypt
 - if their algo looks at intersection, that may work for us, but I think it's unlikely they designed it that way

This would be for another issue

tomholub: I haven't inspected this `getKeyExpirationTimeForCapabilities` method in detail yet

Core/source/test.ts
```diff
@@ -335,12 +337,13 @@ ava.default('isEmailValid - false', async t => {
...
...pqPENaTB/0faBFR\n2k3RM7P427HyZOsZtqEPxuynsLUqmsAAup6LtPhir4CAsb5DSvgYrzC8pbrf\njCaodoB7hMXc8RxTbSh+vQc5Su4QwY8sqy7hyMXOGGWsRxnuZ8t8BeEJBIHy\nPguXIR+wYvo1eveC+NMxHhTtjoSIn/E4vW0W9j5OlFeTK7HTNCuidIE0Hk2k\nXnEEoNO7ztxPPxsHz9g56uMhyAhf3mqKfvUFo/FLLRBOpxLO0kk64yAMcAHm\nc6ZI5Fz10y48+hHEv/RFOwfub9asF5NWHltanqyiZ+kHeoaieYJFc6t7Mt3j\ng8qxMKTUKAEeCfHt1UJCjp/aIgJRU4JRXgYXzsBNBFwBWOEBB/9nclmx98vf\noSpPUccBczvuZxmqk+jY6Id+vBhBFoEhtdTSpaw/JNstf0dTXN8RCFjB0lHt\na51llTjSobqcFwAU54/HKDOW3qMVbvadaGILpuCMCxdMgLWlpZdYY7BApv1N\n9zpN+iQ2tIrvnUQ312xKOXF/W83NUJ1nTObQYNpsUZLLG2N3kz11HuBS3E9F\ngEOYYy1tLT53hs5btqvQ5Jp4Iw5cBoBoTAmv+dPMDKYBroBPwuFeNRIokwLT\nrVcxrXajxlXaGXmmGS3PZ00HXq2g7vKIqWliMLLIWFl+LlVb6O8bMeXOT1l0\nXSO9GlLOSMDEc7pY26vkmAjbWv7iUWHNABEBAAHCwGkEGAEIABMFAlwBWOEJ\nEDow9MwKmo8QAhsMAAoJEDow9MwKmo8QjTcH/1pYXyXW/rpBrDg7w/dXJCfT\n8+RVYlhW3kqMxbid7EB8zgGVTDr3us/ki99hc2HjsKbxUqrGBxeh3Mmui7OD\nCI8XFeYl7lSDbgU6mZ5J4iXzdR8LNqIib4Horlx/Y24dOuvikSUNpDtFAYfa\nbZwxyKa/ihZT1rS1GO3V7tdAB9BJagJqVRssF5g5GBUAX3sxQ2p62HoUxPlJ\nOOr4AaCc1na92xScBJL8dtBBRQ5pUZWOjb2UHp9L5QdPaBX8T9ZAieOiTlSt\nQxoUfCk7RU0/TnsM3KqFnDFoCzkGxKAmU4LmGtP48qV+v2Jzvl+qcmqYuKtw\nH6FWd+EZH07MfdEIiTI=\n=wXbX\n-----END PGP PUBLIC KEY BLOCK-----\n",
```
tomholub, line 346: I'm worried seeing this type of change here. If we change both the implementation and test assets in one PR, how can we be really sure we didn't break something?

Core/source/test.ts
```diff
@@ -349,28 +352,29 @@ ava.default('parseKeys', async t => {
           { "fingerprint": "9EF2F8F36A841C0D5FAB8B0F0BAB9C018B265D22", "longid": "0BAB9C018B265D22", "shortid": "8B265D22", "keywords": "ARM FRIEND ABOUT BIND GRAPE CATTLE" }
         ],
         "algo": {
-          "algorithm": "rsa_encrypt_sign",
-          "bits": 2048,
+          "algorithm": "rsaEncryptSign",
```
tomholub, line 355: This will be a breaking change for the iOS app (maybe). We'll see later if it breaks the test.

Core/source/test.ts
```diff
@@ -349,28 +352,29 @@ ava.default('parseKeys', async t => {
...
+          "bits": 2047,
```
tomholub, line 356: Also worrying, probably related to the key change above, which should be reverted.

Core/source/test.ts
```diff
@@ -349,28 +352,29 @@ ava.default('parseKeys', async t => {
...
...MJawNCLui6Fx4Ba1ipPTaqlJPybLoe6z/WVZwAA9CmbjkCIk683pp\nmGQ3GXv7f8Sdk7DqhEhfZ7JtAK/Uw2VZqqIryNrrB0WV3EUHsENCOlq0YJod\nLqtkqgl83lCNDIkeoQwq4IyrgC8wsPgF7YMpxxQLONJvChZxSdCDjnfX3kvO\nZsLYFiKnNlX6wyrKAQxWnxxYhglMf0GDDyh0AJ+vOQHJ9m+oeBnA1tJ5AZU5\naQHvRtyWBKkYaEhljhyWr3eu1JjK4mn7/W6Rszveso33987wtIoQ66GpGcX2\nmh7y217y/uXz4D3X5PUEBXIbhvAPty71bnTOwE0EXw9x1QEIALdJgAsQ0Jnv\nLXwAKoOammWlUQmracK89v1Yc4mFnImtHDHS3pGsbx3DbNGuiz5BhXCdoPDf\ngMxlGmJgShy9JAhrhWFXkvsjW/7aO4bM1wU486VPKXb7Av/dcrfHH0ASj4zj\n/TYAeubNoxQtxHgyb13LVCW1kh4Oe6s0ac/hKtxogwEvNFY3x+4yfloHH0Ik\n9sbLGk0gS03bPABDHMpYk346406f5TuP6UDzb9M90i2cFxbq26svyBzBZ0vY\nzfMRuNsm6an0+B/wS6NLYBqsRyxwwCTdrhYS512yBzCHDYJJX0o3OJNe85/0\nTqEBO1prgkh3QMfw13/Oxq8PuMsyJpUAEQEAAcLAfAQYAQgADwUCXw9x1QUJ\nAAAAAQIbDAAhCRC+46QtmpyKyRYhBG0+CYZ1RO5ify6Sj77jpC2anIrJARgH\n/1KV7JBOS2ZEtO95FrLYnIqI45rRpvT1XArpBPrYLuHtDBwgMcmpiMhhKIZC\nFlZkR1W88ENdSkr8Nx81nW+f9JWRR6HuSyom7kOfS2Gdbfwo3bgp48DWr7K8\nKV/HHGuqLqd8UfPyDpsBGNx0w7tRo+8vqUbhskquLAIahYCbhEIE8zgy0fBV\nhXKFe1FjuFUoW29iEm0tZWX0k2PT5r1owEgDe0g/X1AXgSQyfPRFVDwE3QNJ\n1np/Rmygq1C+DIW2cohJOc7tO4gbl11XolsfQ+FU+HewYXy8aAEbrTSRfsff\nMvK6tgT9BZ3kzjOxT5ou2SdvTa0eUk8k+zv8OnJJfXA=\n=LPeQ\n-----END PGP PUBLIC KEY BLOCK-----\n",
```
tomholub, line 377: This applies for all test assets above and below - should not change together with changing implementation.

Core/source/test.ts
```diff
@@ -745,18 +761,22 @@ ava.default('verify plain-text signed message by providing it wrong key (fail: c
...
-  const { json: decryptJson, data: decryptData } = parseResponse(await endpoints.parseDecryptMsg({ keys, isEmail: true, verificationPubkeys: pubKeys }, [await getCompatAsset('mime-email-plain-signed-edited')]));
+  const { json: decryptJson, data: decryptData } = parseResponse(
+    await endpoints.parseDecryptMsg({ keys, isEmail: true, verificationPubkeys: pubKeys },
+      [await getCompatAsset('mime-email-plain-signed-edited')]));
   expect(decryptJson.replyType).equals('plain');
   expect(decryptJson.subject).equals('mime email plain signed');
   const parsedDecryptData = JSON.parse(decryptData.toString());
   expect(!!parsedDecryptData.verifyRes).equals(true);
-  expect(parsedDecryptData.verifyRes.match).equals(false);
+  expect(parsedDecryptData.verifyRes.match).is.null;
```
tomholub, line 771: Unsure if this is a breaking change for iOS app or not - will find out later as well.

Core/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+clear
+npm run test
```
tomholub, line 3: Please gitignore, I'd prefer that there is just one obvious way to run tests, which here is `npm run test`. I'm ok to add `clear` there if it doesn't break anything.

Core/tooling/build.sh
```diff
@@ -2,15 +2,50 @@
...
 rm -rf build/ts build/bundles build/final/*
 mkdir -p build/final
 
 # build our source with typescript
-./node_modules/.bin/tsc --project tsconfig.json
+node_modules/.bin/tsc --project tsconfig.json
 
 # build raw/ with webpack
-./node_modules/.bin/webpack --config webpack.bare.config.js
+node_modules/.bin/webpack --config webpack.bare.config.js
```
tomholub, line 48: Did this give you trouble? `./` will ensure that it's run even if not marked as executable - I think. In either case, it shouldn't hurt and is more explicit.

Core/tooling/fix-bundles.js
```diff
@@ -70,9 +70,10 @@ const replace = (libSrc, regex, replacement) => {
   return libSrc.replace(regex, replacement);
 }
 
-let openpgpLib = fs.readFileSync('source/lib/openpgp.js').toString();
+let openpgpLib = fs.readFileSync('./node_modules/openpgp/dist/node/openpgp.js').toString();
 const openpgpLibNodeDev = openpgpLib; // dev node runs without any host, no modifications needed
 
+/*
```
tomholub, line 76: This will break the version of this code that builds for the iOS app (likely make it very slow).

The purpose is to offload intensive operations to Swift/ObC/C code that does it faster.

Can do in another PR.

Core/tooling/fix-bundles.js
```diff
@@ -85,9 +86,11 @@ openpgpLib = replace( // rsa verify on host
   const computed = await coreHost.verifyRsaModPow(m.toString(10), e.toString(10), n.toString(10)); // returns empty str if not supported: js fallback below
   const EM = computed ? new _bn2.default(computed, 10).toArrayLike(Uint8Array, 'be', n.byteLength()) : await _public_key2.default.rsa.verify(m, n, e);`
 );
+*/
 
 let openpgpLibBare = openpgpLib; // further modify bare code below
 
+/*
 openpgpLibBare = replace( // bare - produce s2k (decrypt key) on host (because JS sha256 implementation is too slow)
   openpgpLibBare,
```
tomholub, line 95: Same here ok for a quick fix to get build working, but will need to be fixed after.

tomholub (changes requested):
Looks good, with comments below. Especially the updated tests are worrying - please have a look.

Core/tooling/build.sh
```diff
@@ -2,15 +2,50 @@
...
+# fix openpgp in node_modules
+echo "Patching openpgp.js v5..."
+for f in openpgp.min.js openpgp.min.js.map openpgp.min.mjs openpgp.min.mjs.map openpgp.mjs;
+do
+  if [ -f node_modules/openpgp/dist/$f ]; then rm -f node_modules/openpgp/dist/$f ; fi
+  if [ -f node_modules/openpgp/dist/node/$f ]; then rm -f node_modules/openpgp/dist/node/$f ; fi
+done
+extra_exports="
+// -----BEGIN ADDED BY FLOWCRYPT----
+exports.readToEnd = readToEnd;
```
tomholub, line 14: This solution works but is too hacky if the other alternative is to add type definitions as suggested in https://github.com/openpgpjs/openpgpjs/discussions/1478#discussioncomment-2149306 and https://github.com/openpgpjs/web-stream-tools/issues/10

You don't initially even need to make any PR there. You can simply `import readToEnd as untypedReadToEnd fron 'openpgp/web-stream-tools'` and then `const readToEnd = untypedReadToEnd as WhatEverTypeYouWant`

That is much cleaner than approach here.

Core/tooling/build.sh
```diff
@@ -2,15 +2,50 @@
...
+exports.util = util;
```
tomholub, line 15: Similarly here, except we'll have to re-implement. Or copy with a link to the source.

It's the following methods, if I'm seeing correctly: 
 - in native-crypto.js `Uint8Array_to_str` and `b64_to_Uint8Array`
 - in actual code `uint8ArrayToString`

I see only one use which is here:

```
  public static random = () => { // eg TDW6-DU5M-TANI-LJXY
    return base64encode(util.uint8ArrayToString(secureRandomBytes(128)))
      .toUpperCase().replace(/[^A-Z0-9]|0|O|1/g, '').replace(/(.{4})/g, '$1-').substr(0, 19);
  }
```

And that is not even used, so let's remove it. That way, can also remove this `build.sh` update. (let me know if there are other usages I overlooked)

We may still later need that implementation for `native-crypto.js` but for this PR here, we can do without.

Core/source/core/pgp-msg.ts
```diff
@@ -184,32 +212,50 @@ export class PgpMsg {
...
+          success: false,
+          error: { type: DecryptErrTypes.usePassword, message: 'Use message password' },
+          longids,
+          isEncrypted
+        };
       }
       const passwords = msgPwd ? [msgPwd] : undefined;
-      const privateKeys = keys.prvForDecryptDecrypted.map(ki => ki.decrypted!);
-      const decrypted = await (prepared.message as OpenPGP.message.Message).decrypt(privateKeys, passwords, undefined, false);
+      const privateKeys = keys.prvForDecryptDecrypted.map(ki => ki.decrypted!) as PrivateKey[];
```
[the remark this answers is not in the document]
IvanPizhenko: There's also 
```typescript
longids.chosen = keys.prvForDecryptDecrypted.map(ki => ki.longid);
```
So I'd not touch it now. It is not significant for the overall upgrade.

Core/source/core/pgp-msg.ts
```diff
@@ -220,34 +266,51 @@ export class PgpMsg {
...
+        message,
+        date,
+        encryptionKeys: await (async () => {
+          const encryptionKeys = [];
+          for (const armoredPubkey of pubkeys) {
+            const publicKeys = await readKeys({armoredKeys: armoredPubkey});
+            encryptionKeys.push(...publicKeys);
+          }
+          return encryptionKeys;
+        })(),
```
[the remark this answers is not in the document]
IvanPizhenko: It doesn't work, that's why I had to write this `if`. Error:
```
No overload matches this call.
  Overload 1 of 3, '(options: EncryptOptions & { message: Message<Uint8Array>; format?: "armored" | undefined; }): Promise<string>', gave the following error.
    Type '"armored" | "binary"' is not assignable to type '"armored" | undefined'.
      Type '"binary"' is not assignable to type '"armored" | undefined'.
  Overload 2 of 3, '(options: EncryptOptions & { message: Message<MaybeStream<Data>>; format: "binary"; }): Promise<Uint8Array | WebStream<Uint8Array> | NodeStream<...>>', gave the following error.
```

Core/source/platform/require.ts
IvanPizhenko: CI tests, even those are in the Swift part - are passing, so I assume it just works.

Core/source/platform/util.ts
IvanPizhenko: Yes, it is from openpgp v4. I'll mark it.

Core/tooling/fix-bundles.js
IvanPizhenko: Since Swift based tests are passing - I assume it doesn't break. But it can slowdown.
But because library internals have significantly changed, this code was failing. So I've commented it out.
Also I've read somewhere (not sure now where exactly) that openpgp v5 automatically uses some native stuff if it is available.

Core/tooling/build.sh
IvanPizhenko: ok, will revert

Core/tooling/build.sh
IvanPizhenko: I'm not a big specialist in writing typescript definitions, so I unlikely will do it, and they unlikely will do in the neartest time. So I see this as easiest and the most efficient way. This way we don't need to deal with `web-stream-tools` separately. I assume they are just included into final JavaScript source of the library and they don't plan to change this, although they don't export that explicitly.

Core/tooling/build.sh
IvanPizhenko: Removed `random()`. So after this we really only have one extra export, which sounds not a big hack.

Core/source/test.ts
IvanPizhenko: They have changed those names, so I had to change them too.

Core/source/test.ts
IvanPizhenko: The differences here are following: 
- newline style is `\n`, not `\r\n`
- comments like `Version: FlowCrypt [....] ...` are stripped

Core/test.sh
IvanPizhenko: Removed from repo

IvanPizhenko:
> > UPDATE: I've implemented new version of the lastSig() and changed the way we patch openpgp.js. Although it is not exactly using OOB one, we don't need to store our own version of the library code in the our repo, since patch can be applied automatically by the build script. Still we need to distribute somehow modified version, because it is LGPL.
> 
> Since the patch is out in the open here on public repo, I believe that counts as distributing and therefore compatible with LGPL.
> 
> If we only did it on a closed repo, then we'd have to think about it more.

It is not only distributing, you must allow freely changing it. And here you say: "limitations apply".

IvanPizhenko:
@tomholub I've applied some fixes for the most of your comments. Maybe except export of `readToEnd`, which I think still the easiest and consistent way for us. So please recheck these changes.

Core/source/core/pgp-msg.ts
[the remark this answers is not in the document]
tomholub: That one should be also `PrivateKey[]`. I'm ok to file another issue for these.

Core/source/core/pgp-msg.ts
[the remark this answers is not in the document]
tomholub: I suspected so. Let's keep the if since I'm unsure of a cleaner workaround, and extract the encryptionKeys code above to prepare a variable.

Core/source/core/pgp-msg.ts
```diff
@@ -220,34 +266,51 @@ export class PgpMsg {
...
-    }
-    if (pwd) {
-      options.passwords = [pwd];
-    }
     if (!pubkeys && !pwd) {
       throw new Error('no-pubkeys-no-challenge');
     }
-    if (signingPrv && typeof signingPrv.isPrivate !== 'undefined' && signingPrv.isPrivate()) { // tslint:disable-line:no-unbound-method - only testing if exists
-      options.privateKeys = [signingPrv];
+    const message = await createMessage({binary: data, filename, date});
```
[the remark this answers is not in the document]
tomholub: Sure, it's well readable. I was just thinking for consistency. Let's file another issue to look into formatting across the TS code later.

Core/source/core/pgp-key.ts
```diff
@@ -356,20 +363,34 @@ export class PgpKey {
...
+          allSignatures.push(selfCertification);
+        } catch (e) {
+          console.log(e);
+        }
+      }
     }
     for (const subKey of key.subkeys) {
-      allSignatures.push(...subKey.bindingSignatures);
+      try {
+        const latestValidSig = await subKey.verify();
```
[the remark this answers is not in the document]
tomholub: I see, thank you. How about `latestValidKeyBindingSig` ?

Core/tooling/build.sh
tomholub: This really has to go - fixing this is as easy as adding the separate dependency and defining the type for one method. There is no excuse for us to be adding shell scripts and patching things in node_modules if the same can be resolved through standard methods.

If there was no other way, that would be a different story.

Core/tooling/fix-bundles.js
tomholub: There is no browser where we run this code. There is `JavaScriptCore` which is a pure JS engine without any bells or whistles - no window, no node, no setTimeout or setInterval, no ... etc. We'll still need to supply our own crypto methods, unfortunately. It would be great if we didn't have to :-)

Core/source/platform/require.ts
tomholub: They pass because they use already built javascript from before - your changes haven't been updated to the app yet. The built product it's using: https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/Resources/flowcrypt-ios-prod.js.txt

You can run `npm run-script deploy` which will rebuild the code and replace that file. If you commit after that, Swift/Appium tests will certainly break.

But within this PR, we don't have to do that yet. Let's focus on getting TypeScript pass its own tests + cleaning up the PR as is, and then look into making it work for the app in another PR.

Core/source/test.ts
tomholub: Got it. Just highlighting that someone may have to update Swift code after the code is deployed to the app, possibly.

Core/tooling/fix-bundles.js
tomholub: > Since Swift based tests are passing - I assume it doesn't break. But it can slowdown.

They don't take this new code into account yet - still using previously built version, until we replace that manually, which we'll do in another PR. It will be an effort on its own.

tomholub:
> > > UPDATE: I've implemented new version of the lastSig() and changed the way we patch openpgp.js. Although it is not exactly using OOB one, we don't need to store our own version of the library code in the our repo, since patch can be applied automatically by the build script. Still we need to distribute somehow modified version, because it is LGPL.
> > 
> > 
> > Since the patch is out in the open here on public repo, I believe that counts as distributing and therefore compatible with LGPL.
> > If we only did it on a closed repo, then we'd have to think about it more.
> 
> It is not only distributing, you must allow freely changing it, reusing it, etc. And here you say: "limitations apply".

We should be able to fulfill those requirements by adding `Except when stated otherwise` to the main license file.

Then for the one method that we need and we brought over, we add a comment `This method, brought from OpenPGP.js project is, unlike the rest of the codebase, licensed under LGPL. See original license file: (OpenPGP.js repo license url)`

Core/source/core/pgp-msg.ts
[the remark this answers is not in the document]
IvanPizhenko: filed #1376

Core/source/platform/require.ts
IvanPizhenko: ok

Core/source/test.ts
IvanPizhenko: The only changes are newline style and removal of comments (like "Version: ...") to match what is returned from openpgp.js.

Core/tooling/fix-bundles.js
IvanPizhenko: ok

Core/source/core/pgp-key.ts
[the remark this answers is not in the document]
IvanPizhenko: ok, will rename

Core/source/core/pgp-msg.ts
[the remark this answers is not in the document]
IvanPizhenko: ok

IvanPizhenko:
> > > > UPDATE: I've implemented new version of the lastSig() and changed the way we patch openpgp.js. Although it is not exactly using OOB one, we don't need to store our own version of the library code in the our repo, since patch can be applied automatically by the build script. Still we need to distribute somehow modified version, because it is LGPL.
> > > 
> > > 
> > > Since the patch is out in the open here on public repo, I believe that counts as distributing and therefore compatible with LGPL.
> > > If we only did it on a closed repo, then we'd have to think about it more.
> > 
> > 
> > It is not only distributing, you must allow freely changing it, reusing it, etc. And here you say: "limitations apply".
> 
> We should be able to fulfill those requirements by adding `Except when stated otherwise` to the main license file.
> 
> Then for the one method that we need and we brought over, we add a comment `This method, brought from OpenPGP.js project is, unlike the rest of the codebase, licensed under LGPL. See original license file: (OpenPGP.js repo license url)`

I've put such notice on that function. I'll leave to you updating the LICENSE text.

tomholub:
> I'll leave to you updating the LICENSE text.

updated on master

Core/tooling/build.sh
IvanPizhenko: @tomholub Well, I have removed this patch, I've added `@openpgp/web-stream-tools` into package.json, I've tried to import `readToEnd()` from it. And now for this code:
```typescript
import { readToEnd } from '@openpgp/web-stream-tools';
```
I'm getting following error:
```
Could not find a declaration file for module '@openpgp/web-stream-tools'. '/home/ivan/pro/flowcrypt/flowcrypt-ios/Core/node_modules/@openpgp/web-stream-tools/lib/streams.js' implicitly has an 'any' type.
```
What I've tried:
https://javascript.plainenglish.io/how-to-write-typescript-types-for-a-javascript-library-e598b9eb8be7
But in `ambients.d.ts` for the code
```typescript
declare module "@openpgp/web-stream-tools" {
```
I'm getting error:
```
Invalid module name in augmentation. Module '@openpgp/web-stream-tools' resolves to an untyped module at '/home/ivan/pro/flowcrypt/flowcrypt-ios/Core/node_modules/@openpgp/web-stream-tools/lib/streams.js', which cannot be augmented.ts(2665)
```
Any ideas how to solve this?

IvanPizhenko: Basically I think my questions are:
- where to put file with type definitions? (exact directory)
- what should be file name?
- how to solve above error if it will happen if I create properly named file in the proper directory.

IvanPizhenko: Removed `ambients.d.ts`, placed npm-installed-like type defintion in the `./source/core/types/`. No effect. So I really don't know how to solve this.

IvanPizhenko: This is why I tell patching openpgp.js the way I did was significantly easier.

IvanPizhenko: One more update - tried to copy that into `node_modules/@types`. So I have `node_modules/@types/@openpgp/web-stream-tools`. But - this also didn't work. No more ideas.

IvanPizhenko: Updated build script with web-stream-tools patch.

IvanPizhenko: But as result ... unfortunately, it just doesn't work:
```
> flowcrypt-mobile-core@0.0.1 test
> ./node_modules/.bin/ava build/ts/test.js --verbose


  Uncaught exception in build/ts/test.js

  /home/ivan/pro/flowcrypt/flowcrypt-ios/Core/node_modules/@openpgp/web-stream-tools/lib/streams.js:1
  import { isNode, isStream, isArrayStream, isUint8Array, concatUint8Array } from './util';
  ^^^^^^

  SyntaxError: Cannot use import statement outside a module

  › SyntaxError: Cannot use import statement outside a module
  › Object.<anonymous> (build/ts/core/pgp-key.js:14:28)

  ✖ build/ts/test.js exited with a non-zero exit code: 1
```

My overall impression - web-stream-tools are in the too early stage. I strongly suggest to give up on them for now and just patch openpgp.js - which was already proven to work. Later on, when they get more mature - we can try them again.

IvanPizhenko: @tomholub - what do you think of above?

IvanPizhenko: Posted them PR with initial type definitions https://github.com/openpgpjs/web-stream-tools/pull/11

IvanPizhenko:
@tomholub So, as long as `web-stream-tools` doesn't work,  I'm reverting to patch for openpgp.js. Later on, when they resolve it, we can try `web-stream-tools` again.

tomholub:
> @tomholub So, as long as `web-stream-tools` doesn't work, I'm reverting to patch for openpgp.js. Later on, when they resolve it, we can try `web-stream-tools` again.

You didn't actually my suggestion from above, that I could see. You don't need a d.ts file. You don't need the library to have types at all.

> This solution works but is too hacky if the other alternative is to add type definitions as suggested in https://github.com/openpgpjs/openpgpjs/discussions/1478#discussioncomment-2149306 and https://github.com/openpgpjs/web-stream-tools/issues/10
> 
> You don't initially even need to make any PR there. You can simply `import readToEnd as untypedReadToEnd fron 'openpgp/web-stream-tools'` and then `const readToEnd = untypedReadToEnd as WhatEverTypeYouWant`
> 
> That is much cleaner than approach here.

You should be able to:

```ts
import { readToEnd as readToEndUntyped } from '@openpgpjs/web-stream-tools';
// some other imports...

type ReadToEndFn = (x: SomeInputType) => Promise<SomeOutputType>;
const readToEnd = readToEndUntyped as ReadToEndFn;

// some other code ...

let result = await readToEnd(stream)l

There is no need for any struggle with d.ts files, and unlikely any need to patch anything in node_modules. I will not merge any PR that does that as long as there is any chance of any other solution.

**Comment by tomholub** [comment]
Actually, better, try using a conventional import. `const webStreamToolsModule = require(''@openpgp/web-stream-tools'');`

**Comment by IvanPizhenko** [comment]
> Try adding `// @ts-ignore` just above the import

@tomholub well, `@ts-ignore` lets it compile, but then when I run it with `npm run test` I'm getting the error that I've already faced before:
```
> flowcrypt-mobile-core@0.0.1 test
> ./node_modules/.bin/ava build/ts/test.js --verbose


  Uncaught exception in build/ts/test.js

  /home/ivan/pro/flowcrypt/flowcrypt-ios/Core/node_modules/@openpgp/web-stream-tools/lib/streams.js:1
  import { isNode, isStream, isArrayStream, isUint8Array, concatUint8Array } from './util';
  ^^^^^^

  SyntaxError: Cannot use import statement outside a module

  › SyntaxError: Cannot use import statement outside a module
  › Object.<anonymous> (build/ts/core/pgp-key.js:17:28)

  ✖ build/ts/test.js exited with a non-zero exit code: 1
```
And this seems to be the problem in the `web-stream-tools` themselves.

**Comment by tomholub** [comment]
A `require` may work, in that case.

**Comment by IvanPizhenko** [comment]
So all sorts of "import" for web-stream-tools fail the same way. I still suggest to give up on it for now, and add it later if they fix it.

**Comment by tomholub** [comment]
I'll take a look.

**Comment by tomholub** [comment]
On a first look, it seems that OpenPGP.js library uses the streams module through another build step - rollup. My guess is that the library is optimized for that. But it doesn't look like something that we would be unable to use. Still looking.

File: Core/source/core/pgp-key.ts [hunk 26]
```diff
@@ -349,32 +364,53 @@ export class PgpKey {
...
     }
     throw new Error('No valid signature found in key');
   }
 
-  public static revoke = async (key: OpenPGP.key.Key): Promise<string | undefined> => {
-    if (! await key.isRevoked()) {
-      key = await key.revoke({});
+  public static revoke = async (key: Key): Promise<string | undefined> => {
+    // todo - quick hack to pass tests
+    if (!key.isRevoked(undefined as any as SignaturePacket)) {
```
tomholub, line 404: @IvanPizhenko I added this while debugging the imports - it's wrong as is - please have a look

Core/source/core/pgp-key.ts
```diff
@@ -349,32 +364,53 @@ export class PgpKey {
...
+    // "await key.getExpirationTime()" no longer works, need some alternate solution
+    // discussion is in progress: https://github.com/openpgpjs/openpgpjs/discussions/1491
+    const allSignatures: SignaturePacket[] = [];
     for (const user of key.users) {
-      allSignatures.push(...user.selfCertifications);
+      const data = { userID: user.userID, userAttribute: user.userAttribute, key: key };
+      for (const selfCert of user.selfCertifications) {
+        try {
+          // todo - add proper type in openpgp.d.ts
+          await selfCert.verify(key.keyPacket, enums.signature.certGeneric, data as any); // quick hack
```
tomholub, line 376: @IvanPizhenko I added this wile debugging imports, it's wrong as is, please have a look

tomholub (commented):
It works now, without modifying node_modules which should never be done.

I had to tell webpack to transpile the dependency from ES6 to ES5, and then it's imported with `require`, with some types added.

It's annoying to have to transpile it, but it's less bad than modifying node_modules.

Core/source/core/pgp-key.ts
IvanPizhenko: Because you removed copying modified `openpgp.d.ts` in the build script into node_modules, it doesn't pick up rest of my changes. These changes are submitted as PRs to openpgp.js and some are even merged. However what you've done here also works, and it must stay here until openpgp.js releases new version with my edits to `openpgp.d.ts`.

Core/source/core/pgp-key.ts
IvanPizhenko: see above - need to have openpgp.js to release new version with my edits.

IvanPizhenko:
> It works now, without modifying node_modules which should never be done.
> 
> I had to use webpack to transpile the dependency from ES6 to ES5, and then it's imported with `require`, with some types added.
> 
> It's annoying to have to transpile it, but it's less bad than modifying node_modules.

That's obviously over my experience with JavaScript, so I could not even imagine such solution myself.

IvanPizhenko:
@tomholub I've reviewed your recent changes and made couple more small fixes, but I think I have nothing else to add here, and it is ready for the next stage when it needs to be made running on iOS.

tomholub:
> > It works now, without modifying node_modules which should never be done.
> > I had to use webpack to transpile the dependency from ES6 to ES5, and then it's imported with `require`, with some types added.
> > It's annoying to have to transpile it, but it's less bad than modifying node_modules.
> 
> That's obviously over my experience with JavaScript, so I could not even imagine such solution myself.

The number one rule to learn - do not modify node_modules :-) it's unexpected and undesired, and will likely blow something else up when someone else touches it.

Don't shy away from learning about the javascript ecosystem. I myself had to literally look up "what is webpack" this evening because I long forgot it all. It's what we as developers need to do to get the job done, even if particular stack is not exactly our cup of tea.

Core/source/core/pgp-key.ts
tomholub: https://github.com/FlowCrypt/flowcrypt-ios/issues/1377

Core/source/core/pgp-key.ts
tomholub: https://github.com/FlowCrypt/flowcrypt-ios/issues/1377

IvanPizhenko:
I understand this, but if you at least knew what to lookup, I had no idea completely what can be done else. So I would (and I already said this before) stick for some time with such slightly patched openpgp.js and would wait until they fix web-stream-tools so that it does not require any extra steps to use it and then update to them.

Core/source/core/pgp-key.ts
tomholub: ok for now

Core/source/core/pgp-key.ts
tomholub: ok for now

Core/source/core/pgp-msg.ts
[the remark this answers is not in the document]
tomholub: https://github.com/FlowCrypt/flowcrypt-ios/issues/1380

Core/source/test.ts
tomholub: https://github.com/FlowCrypt/flowcrypt-ios/issues/1379

Core/source/test.ts
tomholub: https://github.com/FlowCrypt/flowcrypt-ios/issues/1379

Core/tooling/fix-bundles.js
tomholub: https://github.com/FlowCrypt/flowcrypt-ios/issues/1379

Core/source/core/pgp-key.ts
```diff
@@ -356,20 +363,34 @@ export class PgpKey {
    * This is used to figure out how recently was key updated, and if one key is newer than other.
    */
   public static lastSig = async (key: Key): Promise<number> => {
```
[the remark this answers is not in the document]
tomholub: We indeed have at least some tests for it, like this one `ava.default('parseKeys', async t => {`

Core/tooling/fix-bundles.js
tomholub: https://github.com/FlowCrypt/flowcrypt-ios/issues/1379

tomholub (changes requested):
Code looks good except the tests as below.

I've closed all conversations that can be closed, and left some open to resolve (you need to click a few times to show them all since there is so many).

Here is a screenshot - this is still unresolved, I cannot merge such large changes together with test asset changes, regardless how small the test asset changes are. Tests must be exactly testing same inputs and same outputs. You can change the inputs in another PR, but not in this one - please revert.

image

tomholub:
Meanwhile @rrrooommmaaa please have a look at this. We'll soon have to update browser-extension to OpenPGP v5 as well - your input on this PR will be greatly appreciated 👍

IvanPizhenko:
@tomholub Please read this feedback from openpgp.js developers regarding `capabilities` - https://github.com/openpgpjs/openpgpjs/discussions/1493#discussioncomment-2174939
Anyway, what I did is tried to port original code of `getExpirationTime()` with `capabilities`.

IvanPizhenko:
> Code looks good except the tests as below.
> 
> I've closed all conversations that can be closed, and left some open to resolve (you need to click a few times to show them all since there is so many).
> 
> Here is a screenshot - this is still unresolved, I cannot merge such large changes together with test asset changes, regardless how small the test asset changes are. Tests must be exactly testing same inputs and same outputs. You can change the inputs in another PR, but not in this one - please revert.
> 
> image

@tomholub In this case these tests will be failing and I will have to mark them with `skip` to have CI passing. Is it ok?

IvanPizhenko:
> > @tomholub In this case these tests will be failing and I will have to mark them with skip to have CI passing. Is it ok?
> 
> Why would they be failing? If it's just removed comments and `\r\n` changed to `\n`? I was exactly worried about this - that there is something more going on.
> 
> What will they be failing on and why?

These changes to test string are done because strings returned by openpgp.js v5 are different. So I had to fix these strings to match what openpgp.js returns now.

Core/source/core/pgp-key.ts
```diff
@@ -349,32 +364,51 @@ export class PgpKey {
...
+        console.log(`PgpKey.lastSig: Skipping subkey ${subKey.getKeyID().toHex()} ` +
+          `because there is no valid binding signature: ${String(e)}`);
+      }
     }
-    allSignatures.sort((a, b) => b.created.getTime() - a.created.getTime());
-    const newestSig = allSignatures.find(sig => sig.verified === true);
-    if (newestSig) {
-      return newestSig.created.getTime();
+    if (allSignatures.length > 0) {
+      allSignatures.sort((a, b) => {
```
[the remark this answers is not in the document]
IvanPizhenko: ok, I've applied this

tomholub:
> These changes to test string are done because strings returned by openpgp.js v5 are different. So I had to fix these strings to match what openpgp.js returns now. Because test matches strings exactly (is.deep.equal(...)).

My bad, I thought these were test inputs. Now I see these are outputs. Looks good.

tomholub:
PR Looks good from my end. @rrrooommmaaa please have a look

tomholub (dismissed):

tomholub:
Once tests are passing, I'm good to merge it, assuming Roman is.

tomholub:
@IvanPizhenko well done to update to v5, not an easy task.

IvanPizhenko:
@tomholub I can see in CI that Appium UI tests fail. What's that?

tomholub:
Likely unrelated. Try updating the branch to master, let's see if that passes.

rrrooommmaaa (approved):

IvanPizhenko:
@tomholub Still failing, even after multiple times merging updates from master

tomholub:
@IvanPizhenko let's have @sosnovsky finalize this, assuming the issue is unrelated to the code (which it should be, we haven't actually changed the app yet)

sosnovsky (approved):

PR #1390, opened 2022-02-21
Add private key decrypt validation

martgil:
This PR adds a private key validation test.

close https://github.com/FlowCrypt/flowcrypt-ios/issues/1346 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


tomholub:
On `normalize` you can let the error be swallowed and test for it returning no keys.

When you add to `parse`, does it have the same issue?

Core/source/core/pgp-key.ts
```diff
@@ -135,6 +135,13 @@ export class PgpKey {
     return p instanceof SecretKeyPacket || p instanceof SecretSubkeyPacket;
   }
 
+  public static validateAllDecryptedPackets = async (key: Key): Promise<void> => {
+    const packets = key.toPacketList() as PacketList<SecretKeyPacket>;
+    for (const prvPacket of packets.filter(PgpKey.isPacketPrivate).filter(packet => packet.isDecrypted())) {
+      await (prvPacket as SecretKeyPacket).validate(); // gnu-dummy never raises an exception, invalid keys raise exceptions
```
IvanPizhenko, line 141: should be type cast to `BaseSecretKeyPacket`

Core/source/core/pgp-key.ts
```diff
@@ -135,6 +135,13 @@ export class PgpKey {
     return p instanceof SecretKeyPacket || p instanceof SecretSubkeyPacket;
   }
 
+  public static validateAllDecryptedPackets = async (key: Key): Promise<void> => {
+    const packets = key.toPacketList() as PacketList<SecretKeyPacket>;
```
IvanPizhenko, line 139: Please remove type cast to `PacketList<SecretKeyPacket>`, not all packets will be of type `SecretKeyPacket`, return type of the `toPacketList()` is `PacketList<AllowedKeyPackets>`, that's what we need to operate on.

IvanPizhenko (commented):
@martgil please see my remarks

IvanPizhenko:
> I also think of adding the validation to `PgpKey.parse()` through `PgpKey.normalize()` but when the throw triggered on `.validate()` function of the Openpgp.js the return from `PgpKey.normalization` kinda overrides the error.

You can extend return type of `normalize()` with additional optional field called `exception`, and store exception there if it has occurred and then inspect it in the test.

Core/source/core/pgp-key.ts
martgil: update, thanks!

Core/source/core/pgp-key.ts
martgil: thank you, sir, I revert the change to use `PacketList<AllowedKeyPackets>` and check if the instance of `prvPacket` is SecretKeyPacket and if the prvPacket is already decrypted because I don't have much idea of how I could make the `.filter(PgpKey.isPacketPrivate)` to work.

Core/source/core/pgp-key.ts
martgil: sir, I made a recent change that would work without further typecasting but by checking if the prvPacket is an instance of SecretKeyPacket. would that be ok?

martgil:
> @martgil You can extend return type of `normalize()` with additional optional field called like `exception` or `error`, and store exception there if it has occurred and then inspect it in the test. Regarding `parse()` - it actually calls `normalize()`, so you can do the same - propagate error from `normalize()` to the object returned by `parse()`.

Thank you sir, it helps a lot. I tried to apply it the better way I know if - please correct me if any of the changes look awful. Thanks again!

martgil:
> On `normalize` you can let the error be swallowed and test for it returning no keys.
> 
> When you add to `parse`, does it have the same issue?

Thank you, sir, I also find this helpful with help of the additional field `exception` on `normalize()` and `parse()` - the test looks much clearer.

Core/source/core/pgp-key.ts
```diff
@@ -135,6 +135,14 @@ export class PgpKey {
     return p instanceof SecretKeyPacket || p instanceof SecretSubkeyPacket;
   }
 
+  public static validateAllDecryptedPackets = async (key: Key): Promise<void> => {
+    for (const prvPacket of key.toPacketList()) {
+      if (prvPacket instanceof SecretKeyPacket && prvPacket.isDecrypted()) {
```
IvanPizhenko, line 140: What about `SecretSubkeyPacket`? Why not just use `isPacketPrivate()` for filtering? Something like:
```typescript
for (const packet of key.toPacketList().filter(PgpKey.isPacketPrivate)) {
  await (packet as BaseSecretKeyPacket).validate();
}
```

Core/source/core/pgp-key.ts
```diff
@@ -182,7 +191,7 @@ export class PgpKey {
     await encryptKey({ privateKey: (prv as PrivateKey), passphrase });
   }
 
-  public static normalize = async (armored: string): Promise<{ normalized: string, keys: Key[] }> => {
+  public static normalize = async (armored: string): Promise<{ normalized: string, keys: Key[], exception: string }> => {
```
IvanPizhenko, line 194: should be optional, i.e. `exception?: string`, i.e. it can be undefined.

Core/source/core/pgp-key.ts
```diff
@@ -196,13 +205,14 @@ export class PgpKey {
       }
       for (const k of keys) {
         for (const u of k.users) {
+          await PgpKey.validateAllDecryptedPackets(k);
           u.otherCertifications = []; // prevent key bloat
         }
       }
-      return { normalized: keys.map(k => k.armor()).join('\n'), keys };
+      return { normalized: keys.map(k => k.armor()).join('\n'), keys, exception: '' };
```
IvanPizhenko, line 212: no need to return `exception` here if it is declared optional

Core/source/core/pgp-key.ts
```diff
@@ -315,9 +325,9 @@ export class PgpKey {
     return undefined;
   }
 
-  public static parse = async (armored: string): Promise<{ original: string, normalized: string, keys: KeyDetails[] }> => {
-    const { normalized, keys } = await PgpKey.normalize(armored);
-    return { original: armored, normalized, keys: await Promise.all(keys.map(PgpKey.details)) };
+  public static parse = async (armored: string): Promise<{ original: string, normalized: string, keys: KeyDetails[], exception: string }> => {
```
IvanPizhenko, line 328: same: `exception` should be optional

Core/source/core/pgp-key.ts
IvanPizhenko: name `exception` is a bit misleading. It could be `exception` if you would return here the whole exception object, but only error message is returned, as I can see later in the code. So I suggest to rename it into `error`, (and moreover, this will be more aligned with what we already have in this codebase).

IvanPizhenko:
@martgil Some more comments, please check.

Core/source/core/pgp-key.ts
IvanPizhenko: What's the problem with `.filter(PgpKey.isPacketPrivate)`? Why doesn't it work?

martgil: I'm not sure but i'm getting the following type error when doing so:
```
No overload matches this call.
  Overload 1 of 2, '(predicate: (value: AllowedKeyPackets, index: number, array: AllowedKeyPackets[]) => value is PrvPacket, thisArg?: any): PrvPacket[]', gave the following error.
    Argument of type '(p: AnyKeyPacket) => p is PrvPacket' is not assignable to parameter of type '(value: AllowedKeyPackets, index: number, array: AllowedKeyPackets[]) => value is PrvPacket'.
      Types of parameters 'p' and 'value' are incompatible.
        Type 'AllowedKeyPackets' is not assignable to type 'BasePublicKeyPacket'.
          Type 'SignaturePacket' is missing the following properties from type 'BasePublicKeyPacket': algorithm, getAlgorithmInfo, getFingerprint, getFingerprintBytes, and 6 more.
  Overload 2 of 2, '(predicate: (value: AllowedKeyPackets, index: number, array: AllowedKeyPackets[]) => unknown, thisArg?: any): AllowedKeyPackets[]', gave the following error.
    Argument of type '(p: AnyKeyPacket) => p is PrvPacket' is not assignable to parameter of type '(value: AllowedKeyPackets, index: number, array: AllowedKeyPackets[]) => unknown'.
      Types of parameters 'p' and 'value' are incompatible.
        Type 'AllowedKeyPackets' is not assignable to type 'BasePublicKeyPacket'.ts(2769)
```

Core/source/core/pgp-key.ts
martgil: oh, okay sir. I've rename it appropriately to `error` field, making it optional and removing the error from the returning data. thanks!

Core/source/core/pgp-key.ts
martgil: updated, thanks!

Core/source/core/pgp-key.ts
martgil: updated, thanks!

IvanPizhenko:
@martgil Generally looks good, I will try later today to solve this issue with `isPrvPacket()`.

Core/source/core/pgp-key.ts
IvanPizhenko: Maybe just change parameter type of `isPrvPacket()` to `AllowedKeyPackets`?

IvanPizhenko: @martgil I've changed parameter type of the `isPacketPrivate()` to `AllowedKeyPackets` and it worked. I've pushed this change.

sosnovsky (approved):

PR #1398, opened 2022-02-23
#1337 Include names for from/to/cc/bcc emails

### LINKED ISSUE #1337: include names in `to/cc/bcc` fields when calling FES
client app is stripping out the names of the Google contacts - and sending just the email address.

```json
{"associateReplyToken":"e24d308e-6595-42f0-965f-282038d9247d","from":"sender@example.com","to":["recipient@example.com"],"cc":[],"bcc":[]}
```

FES can deal with the name and email address for all the from/to fields like `Recipient Name <recipient@example.com>` - please update client to send it

_Originally posted by @Jerbell in https://github.com/FlowCrypt/flowcrypt-enterprise-on-prem/issues/2153#issuecomment-1018210504_

related issues https://github.com/FlowCrypt/flowcrypt-android/issues/1662 https://github.com/FlowCrypt/flowcrypt-ios/issues/1337 https://github.com/FlowCrypt/flowcrypt-browser/issues/4268

### LINKED ISSUE #1437: app crash on reply all of certain message
@tomholub I found issue
app crash when user click on "Reply all" this email "Message with cc and multiple recipients and attachment"

https://user-images.githubusercontent.com/91205853/159701293-6b0baff9-5e0e-4302-be73-ed3f10168495.mov

_Originally posted by @fcvakintos in https://github.com/FlowCrypt/flowcrypt-ios/issues/1432#issuecomment-1076331162_

sosnovsky:
This PR adds names to sender and recipients fields

close #1337
close #1437

- Tests added or updated

### To be filled by reviewers


tomholub:
@ioanmo226 please have a look at bringing this up to date to master, and then fixing / finalizing. The goal was to include names of recipients in a particular API (that until now only included email addresses), but it seems it turned out a little larger change than expected.

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -207,6 +207,7 @@ final class ComposeViewController: TableNodeViewController {
         self.contextToSend.recipients = [
             ComposeMessageRecipient(
                 email: "tom@flowcrypt.com",
+                name: "Tom",
                 type: .to,
                 state: decorator.recipientIdleState
             )
```
tomholub, line 213: I don't understand, why is there my email, or any email, in production code? This is a VC right? This has no business being here?

when is this method called?

FlowCrypt/Functionality/Services/Contacts Service/Models/RecipientWithSortedPubKeys.swift
```diff
@@ -26,21 +26,12 @@ struct RecipientWithSortedPubKeys {
 extension RecipientWithSortedPubKeys {
     init(_ recipient: Recipient, keyDetails: [KeyDetails] = []) {
         self.email = recipient.email
-        self.name = recipient.name
+        self.name = recipient.name ?? keyDetails.first?.users.first ?? email
```
tomholub, line 29: That's a good idea, use name from pubkey when otherwise missing. But I'm not sure what the `?? email` will do in practice, I hope it will not send an email message with a recipient like `email@example.com <email@example.com>`. And I also hope that `email@example.com <email@example.com>` will not be submitted to the FES upload API. Can you please verify that?

FlowCrypt/Models/Common/Recipient.swift
```diff
@@ -22,4 +23,67 @@ extension Recipient {
...
+            return
+        }
+
+        self.email = email.filter { !["<", ">"].contains($0) }
+        let name = string
+            .replacingOccurrences(of: email, with: "")
+            .replacingOccurrences(of: "\"", with: "")
+            .trimmingCharacters(in: .whitespaces)
+        self.name = name == self.email ? nil : name
+    }
```
tomholub, line 42: This is way too dangerous to do by hand. You need to use an email parser that follows the RFC properly. You can't implement that in a few lines of code, there is a gazillion edge cases when parsing email strings.

tomholub (changes requested):
This looks pretty good, but has an issue + some other comments - see below.

FlowCrypt/Models/Common/Recipient.swift
ioanmo226: Which  library do you mean?
> Our MIME email library probably has a method for this, use that.

FlowCrypt/Functionality/Services/Contacts Service/Models/RecipientWithSortedPubKeys.swift
tomholub: Yes - in that case, should send without name, as it was before.

tomholub (dismissed):

tomholub:
@ioanmo226 please have a look at tests

tomholub (dismissed):

tomholub (approved):

PR #1407, opened 2022-03-09
#1405 Fix cc recipients not respected issue

### LINKED ISSUE #1405: Cc/Bcc recipients are not respected on reply-all
to reproduce:

1) open a message that has cc or bcc recipients in it. In my case, it was `to:me,cc:many people`. It could also be `from:me,to:somebody,cc:others,bcc:the rest` for example.
2) tap reply all
3) the message I'm composing now is `to:original sender` but it's missing all the cc.

ioanmo226:
This PR fixed cc recipients not respected issue when user choose to `Reply All`.

close #1405 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Compose/ComposeViewController.swift
+++ b/FlowCrypt/Controllers/Compose/ComposeViewController.swift
@@ -334,6 +334,16 @@ extension ComposeViewController {
             contextToSend.add(recipient: recipient)
             evaluate(recipient: recipient)
         }
+
+        input.quoteCCRecipients.forEach { email in
+            let recipient = ComposeMessageRecipient(email: email, type: .cc, state: decorator.recipientIdleState)
+            contextToSend.add(recipient: recipient)
+            evaluate(recipient: recipient)
+        }
+
+        if input.quoteCCRecipients.isNotEmpty {
+            shouldShowAllRecipientTypes.toggle()
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -334,6 +334,16 @@ extension ComposeViewController {
             contextToSend.add(recipient: recipient)
             evaluate(recipient: recipient)
         }
+
+        input.quoteCCRecipients.forEach { email in
+            let recipient = ComposeMessageRecipient(email: email, type: .cc, state: decorator.recipientIdleState)
+            contextToSend.add(recipient: recipient)
+            evaluate(recipient: recipient)
+        }
```
tomholub, line 342: Looking at the code, please try the following scenario (in Gmail app for reference, and then also in FlowCrypt app)

1) send and email `to:A` `bcc:B`
2) open sent email. It should show `to:A` `bcc:B` (I think? since you are the sender, you should see this)
3) reply to that email

expected result (in Gmail - please confirm, in which case also in FlowCrypt): the reply will be initiated as `to:A`, `bcc:B`

ioanmo226: Well, IMO bcc should not appear in reply.
image

As you can see from below, `reply all` button is not even there when there are only bcc recipients 
image

tomholub: thank you

tomholub (approved):
👍

tomholub:
Filing a separate issue to write a test for this issue #1405 / PR #1407 . The UI test should:

1) navigate to a message that has both `to:` and `cc:` (some already existing, don't add one during test. Or add one manually to the test account if there isn't any already present, then during test expect it to be already present)
2) tap reply all
3) verify cc included

ioanmo226:
Ok, got it

PR #1410, opened 2022-03-09
#1279 Added logout button in retry dialog on startup error

### LINKED ISSUE #1279: cannot log out if error happens during app startup
> Also it looks that it's impossible to log out if any error occurs during app startup. Please check #1267 There is no other way than delete app if it happens

I'll track that as a separate issue

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1276#issuecomment-1005995217_

ioanmo226:
This PR adds logout button in retry dialog when error happens during app start.

close #1279 // if this PR closes an issue

- Not worth testing

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Services/AppStartup.swift
+++ b/FlowCrypt/Functionality/Services/AppStartup.swift
@@ -152,7 +152,18 @@ struct AppStartup {
         ) { _ in
             self.initializeApp(window: window)
         }
+        let logout = UIAlertAction(
+            title: "log_out".localized,
+            style: .default
+        ) { _ in
+            do {
+                try appContext.globalRouter.signOut(appContext: appContext)
+            } catch let logoutError {
+                Logger.logError("Logout failed due to \(logoutError.localizedDescription)")
+            }
+        }
         alert.addAction(retry)
+        alert.addAction(logout)
         window.rootViewController?.present(alert, animated: true, completion: nil)
     }
 


FlowCrypt/Functionality/Services/AppStartup.swift
```diff
@@ -152,7 +152,18 @@ struct AppStartup {
...
             self.initializeApp(window: window)
         }
+        let logout = UIAlertAction(
+            title: "log_out".localized,
+            style: .default
+        ) { _ in
+            do {
+                try appContext.globalRouter.signOut(appContext: appContext)
+            } catch let logoutError {
+                Logger.logError("Logout failed due to \(logoutError.localizedDescription)")
```
tomholub, line 162: Better here would be to make another alert - let the user know what happened. It doesn't have to have any buttons anymore. Alternatively, we have a "fatal error" VC that you could send the user with the error message.

Because from the perspective of the user, the logout button would appear to do nothing.

However, it's an unlikely situation within an already unlikely situation, so can leave it as is.

tomholub (approved):
.

FlowCrypt/Functionality/Services/AppStartup.swift
ioanmo226: Got it

PR #1419, opened 2022-03-14
#853 Use local contact if public key lookup failed

### LINKED ISSUE #853: failed public key lookup for recipient can be ignored if valid pubkey already in contacts
eg network down (offline) or server slow or timeout etc. If we already have the public key and it's valid, we're only making external http calls to update it. The call is not required to be successful since we already have good keys to use.

ioanmo226:
This PR uses local contact to display compose view recipient state if public key lookup failed.

close #853 // if this PR closes an issue

- Difficult to test (This issue needs offline status or server down status to be reproduced)

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Compose/ComposeViewController.swift
+++ b/FlowCrypt/Controllers/Compose/ComposeViewController.swift
@@ -1049,15 +1049,17 @@ extension ComposeViewController {
         }
 
         Task {
+            var localContact: RecipientWithSortedPubKeys?
             do {
                 if let contact = try await service.findLocalContact(with: recipient.email) {
+                    localContact = contact
                     handleEvaluation(for: contact)
                 }
 
                 let contactWithFetchedKeys = try await service.fetchContact(with: recipient.email)
                 handleEvaluation(for: contactWithFetchedKeys)
             } catch {
-                handleEvaluation(error: error, with: recipient.email)
+                handleEvaluation(error: error, with: recipient.email, localContact: localContact)
             }
         }
     }
@@ -1085,8 +1087,11 @@ extension ComposeViewController {
         }
     }
 
-    private func handleEvaluation(error: Error, with email: String) {
+    private func handleEvaluation(error: Error, with email: String, localContact: RecipientWithSortedPubKeys?) {
         let recipientState: RecipientState = {
+            if let _localContact = localContact {
+                return getRecipientState(from: _localContact)
+            }
             switch error {
             case ContactsError.keyMissing:
                 return self.decorator.recipientKeyNotFoundState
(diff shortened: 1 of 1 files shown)


FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -1085,8 +1087,11 @@ extension ComposeViewController {
         }
     }
 
-    private func handleEvaluation(error: Error, with email: String) {
+    private func handleEvaluation(error: Error, with email: String, localContact: RecipientWithSortedPubKeys?) {
         let recipientState: RecipientState = {
+            if let _localContact = localContact {
+                return getRecipientState(from: _localContact)
+            }
```
tomholub, line 1094: Let's call it simply `contact` instead of `_localContact`.

There's nothing wrong with the underscore but since we don't use that in other parts of the codebase, let's avoid it for consistency.

Thanks!

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -1085,8 +1087,11 @@ extension ComposeViewController {
         }
     }
 
-    private func handleEvaluation(error: Error, with email: String) {
+    private func handleEvaluation(error: Error, with email: String, localContact: RecipientWithSortedPubKeys?) {
         let recipientState: RecipientState = {
+            if let _localContact = localContact {
```
tomholub, line 1092: Here let's only do this if the local contact has a valid state. There could be other states like expired, revoked.

If the local contact is expired, and the lookup fails, I want the result of the lookup to fail visibly, so that the user knows to retry (since the public key that is remote could be not-expired - it's worth retrying).

If the local pubkey is already valid, then there is no need for retrying the remote fetch and that's why we can ignore the fetch error and render it.

tomholub (changes requested):

tomholub:
This will be a test case for @fcvakintos to write after this PR is merged:

1. open compose screen, search `robot@flowcrypt.com`, wait for it to load. Verify renders as valid. 
2. Search for `expired@flowcrypt.com` and wait for that to load. Verify renders as expired.
3. search for `revoked@flowcrypt.com` and wait to load. Verify renders as revoked.
4. Remove all recipients.
5. go offline
6. load `robot@flowcrypt.com` and verify it shows valid while offline (as if you were online)
7. load `expired@flowcrypt.com` and verify it shows an error while offline (could not load while offline)
8. load `revoked@flowcrypt.com` and verify it shows an error while offline (could not load while offline)

tomholub:
@ioanmo226 please go through the test case steps manually, to make sure it has the expected behavior. Later on, @fcvakintos will write an automated test for it.

tomholub (approved):
I forgot you can do `if let contact = contact`. Yes - that's better.

PR #1423, opened 2022-03-15
#1391 Improve to/cc/bcc ui layout

### LINKED ISSUE #1391: recipient fields cc/bcc improvements
I'll approve and merge this but it needs improvement.

First: Cc/Bcc recipients are not respected on reply-all. To reproduce:

1) open a message that has cc or bcc recipients in it. In my case, it was `to:me,cc:many people`. It could also be `from:me,to:somebody,cc:others,bcc:the rest` for example.
2) tap reply all
3) the message I'm composing now is `to:original sender` but it's missing all the cc.

Second: The UI is disorderly. Let's follow what Gmail app does more closely - they do it well:
 - instead of `Add Recipient` placeholder, let's put `To` label to the left of the input that stays there
 - Similarly for cc and bcc when we expand it. Add `cc` and `bcc` to the left as labels
 - Let's auto-collapse all of this, only showing the email bubbles, when switches to subject or body (lost focus)
 - Let's re-expand the recipient area when user taps on the collapsed rendering

This way, we no longer need to have `Add Recipient`, `Add cc Recipient`, `Add bcc recipient` placeholders which take up a lot of space, each occupying own line. When user adds a few recipients, it quickly overwhelms my small screen (12 mini) whereas on Gmail app, it's comfortable.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1381#pullrequestreview-888771544_

ioanmo226:
This PR improved to/cc/bcc layout

close #1391  // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


ioanmo226:
@tomholub MVP version is now ready. Please check if functionality is working as expected.
I'll start writing tests and improving code quality after you confirm.

tomholub:
Definitely an improvement, could you do the following changes:

 - in collapsed view, use red text if the bubble was to be rendered in red, green for green, orange for orange. Gray color bubble should result in black text when collapsed. Please use the existing colors from the app palette.
 - In expanded view, please make line height smaller so that the recipient section doesn't take up so much space when expanded. Adding two `To` and one `Cc` causes half of the screen to be taken up by recipients - I wish it to be more vertically compact

Thanks!

tomholub:
Also, I suppose each section doesn't need an empty row below it? The caret could be to the right of the last recipient in expanded view? Like this view:

image

would become this:

image

ioanmo226:
Currently we have entered recipient list + recipient input field.
I also tried to implement like that but we have to refactor many things and it takes time.
It's like re-design recipients view.

Let me know if you still want me to implement this way and I'll implement it.
> Also, I suppose each section doesn't need an empty row below it? The caret could be to the right of the last recipient in expanded view? Like this view:

tomholub:
I know it's an input field. But it looks a lot like an empty row :-)

We can leave that change for another time. It would be good to do it, but we don't have to do it now.

Let's still do changes mentioned in https://github.com/FlowCrypt/flowcrypt-ios/pull/1423#issuecomment-1068268708 in this PR, and I'll file another issue for the text input update.

ioanmo226:
> I know it's an input field. But it looks a lot like an empty row :-)
> 
> We can leave that change for another time. It would be good to do it, but we don't have to do it now.
> 
> Let's still do changes mentioned in [#1423 (comment)](https://github.com/FlowCrypt/flowcrypt-ios/pull/1423#issuecomment-1068268708) in this PR, and I'll file another issue for the text input update.

ok

tomholub:
The `To` and cc/bcc has different color (or maybe text weight) then the other two. Can you align all of these to be the original gray color?

tomholub:
Otherwise this looks good! Thanks

ioanmo226:
@tomholub Fixed, Now let me work on tests & code improvements.

tomholub (approved):
Looks good :+1:

PR #1430, opened 2022-03-18
#669 Threads in search view

### LINKED ISSUE #669: threads in search view
Can we create a separate ticket to support threads view in search controller?

_Originally posted by @Kharchevskyi in https://github.com/FlowCrypt/flowcrypt-ios/issues/608#issuecomment-942442660_

At the moment the search results are rendered using a custom view, but instead we should use the same view that we use for all other folders.

ioanmo226:
This PR implemented threads in search view

close #669 // if this PR closes an issue

- Does not need tests (Test is already implemented in #959)

### To be filled by reviewers


ioanmo226:
@tomholub Could you check semaphoreci build?
It fails with ruby gem error package error and not sure what is wrong.

FlowCrypt/Controllers/Search/SearchViewController.swift
```diff
@@ -17,26 +17,11 @@ import UIKit
  */
 final class SearchViewController: TableNodeViewController {
```
tomholub, line 18: Could you see if it's feasible / reasonable to completely remove `SearchViewController` and use `InboxViewController` for displaying search results? Alternatively, maybe subclass from `InboxViewController` ?

I'm asking in case it actually helps to deduplicate / simplify code - let me know what you think.

ioanmo226: Nice suggestion.
Yeah, I also thought of that and there are so many duplicated codes in SearchViewController.
Let me refactor

tomholub:
This looks good. You will also have to adjust tests, since they rely on search functionality.

tomholub:
and resolve the conflicts before I can merge. Thanks!

FlowCrypt/Controllers/Inbox/InboxViewController.swift
```diff
@@ -174,14 +191,28 @@ extension InboxViewController {
...
+                if isSearch {
+                    state = .searching
+                    await self.tableNode.reloadData()
+                } else {
+                    state = .fetching
+                }
                 let context = try await service.fetchInboxItems(
                     using: FetchMessageContext(
-                        folderPath: viewModel.path,
+                        folderPath: isSearch ? nil : viewModel.path, // pass nil in search screen to search for all folders
```
ioanmo226, line 213: @tomholub In search screen, Do we have to search for all folders or just current folder?
I thought it should search for all folders so I added this logic.
Gmail and previous search screen also searched for all folders though.

ioanmo226:
We don't need to adjust tests as current version works same as previous search version(but with thread function)
By the way, @tomholub Please check semaphoreci build. 
> You will also have to adjust tests, since they rely on search functionality

FlowCrypt/Controllers/Inbox/InboxViewController.swift
tomholub: Aligning functionality with Gmail is usually the right thing.

ioanmo226: ok

ioanmo226:
@tomholub semaphoreci issue is fixed.
It was because of semaphoreci cache issue and i had to reset cache.
Please review

tomholub (approved):
:+1:

PR #1431, opened 2022-03-21
issue #1376 More consistent formatting

### LINKED ISSUE #1376: Consistent code formatting
Sure, it's well readable. I was just thinking for consistency. Let's file another issue to look into formatting across the TS code later.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1374#discussion_r805207497_

IvanPizhenko:
This PR adds some settings for formatting

close #1376

### To be filled by reviewers


tomholub:
That looks ok, now you want to make sure that the CI tests for it, like on browser extension:

```yaml
      jobs:

        - name: code quality
          commands:
            - npm run-script test_tslint
            - npm run-script test_eslint
            - npm run-script test_stylelint
 ```

**Comment by tomholub** [comment]
You'll also want to copy large part of the workspace file - recommeded extensions and workpace style settings:

https://github.com/FlowCrypt/flowcrypt-browser/blob/master/flowcrypt-browser.code-workspace

**Comment by IvanPizhenko** [comment]
There are lots of warnings from running eslint and tslint, so I'm going to fix that.

**Comment by tomholub** [comment]
To self-correct, `            - npm run-script test_stylelint` is not relevant for this repo

File: Core/source/platform/util.ts [hunk 1]
```diff
@@ -55,31 +55,34 @@ export const str_to_hex = (str: string): string => {
...
-}
+};
 
+/* tslint:disable:no-null-keyword */
 const maxDate = (dates: (Date | null)[]): Date | null => {
-  let res: Date | null = null;
+  let res: Date | undefined;
   for (const date of dates) {
-    if (res == null || (date != null && date > res)) {
+    if (date != null && (res === undefined || date > res)) {
```
tomholub, line 64: `!=` and `==` should be disabled with tslint or eslint. (use `!==` and `===` instead). I'm surprised this passes. Can you please enable the rule for it?

Also, this method was refactored in a way that it's not obvious (from code) if it still does the same thing.

Core/tooling/tslint-rules/noDirectAjaxRule.ts
```diff
@@ -0,0 +1,28 @@
+import * as ts from "typescript";
+import * as tslint from "tslint";
+
+const DO_NOT_USE_AJAX = `Direct ajax calls don't belong here.
+
```
tomholub, line 5: Can remove this rule - there is no network support in iOS JavaScriptCore anyway.

tomholub (commented):
120 is way too short. But I'll live with it. See comments

Core/source/platform/util.ts
IvanPizhenko: It should be already enabled, because I've seen it has worked in other places. Not sure why it didn't work here.
I will better restore previous code, since check for null keyword is already disabled.

IvanPizhenko:
@tomholub Swift tests also fail here

tomholub:
They should stabilize once https://github.com/FlowCrypt/flowcrypt-ios/pull/1480 merges today

Core/source/test.ts
```diff
@@ -45,13 +45,15 @@ ava.default('generateKey', async t => {
       variant: 'curve25519', passphrase: 'riruekfhydekdmdbsyd',
       userIds: [{ email: 'a@b.com', name: 'Him' }]
     }));
+  // tslint:disable:no-unsafe-any
```
tomholub, line 48: `no-unsafe-any` should not apply to tests. Can you exclude all tests from this rule?

tomholub (approved):

Core/source/test.ts
tomholub: Alternatively, can put no-unsafe-any exception on top of every test file

IvanPizhenko: Not sure how to disable specific rule for the specific file. Yes, I think it's easier to disable rule right in the file.

PR #1434, opened 2022-03-22
#1119 Remove private actors

### LINKED ISSUE #1119: remove private actors?
there are a few examples of `private actor` still used when they could probably be refactored away

ioanmo226:
This PR removed private actors.

close #1119 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Compose/ComposeViewController.swift
+++ b/FlowCrypt/Controllers/Compose/ComposeViewController.swift
@@ -523,7 +523,7 @@ extension ComposeViewController {
             signingPrv: signingKey
         )
         UIApplication.shared.isIdleTimerDisabled = true
-        try await encryptAndSend(
+        try await composeMessageService.encryptAndSend(
             message: sendableMsg,
             threadId: input.threadId
         )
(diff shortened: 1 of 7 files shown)


FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -1074,11 +1069,34 @@ extension ComposeViewController {
...
+
+    @MainActor
+    func findLocalContact(with email: String) async throws -> RecipientWithSortedPubKeys? {
+        return try await localContactsProvider.searchRecipient(with: email)
+    }
+
+    @MainActor
+    func fetchContact(with email: String) async throws -> RecipientWithSortedPubKeys {
+        return try await pubLookup.fetchRemoteUpdateLocal(with: email)
+    }
```
tomholub, line 1093: Sorry - I misdirected you. These should be part of an actor, yes. But not MainActor. These are all non-main thread tasks (making an API call, encrypting something etc - all are heavy tasks that should not be on main).

At a first glance, that would suggest that `localContactsProvider` and `cloudContactProvider` and `pubLookup` and `composeMessageService` should be actors. But I don't see any benefit in that? Actor will ensure that only one method runs concurrently. But there is no limitation on making concurrent API calls - no harm in not being implemented as an actor.

Therefore, please try another change - remove these 4 new `@MainActor` methods, and use their contents directly within this class. Let's see what happens.

tomholub (changes requested):

FlowCrypt/Controllers/Compose/ComposeViewController.swift
ioanmo226: @tomholub Yeah, I agree with you. Those tasks should not run on main thread (Sorry, i misunderstood you.)

ioanmo226:
@tomholub Please check

FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -1074,11 +1069,30 @@ extension ComposeViewController {
...
+        return try await cloudContactProvider.searchContacts(query: query)
+    }
+
+    func findLocalContact(with email: String) async throws -> RecipientWithSortedPubKeys? {
+        return try await localContactsProvider.searchRecipient(with: email)
+    }
+
+    func fetchContact(with email: String) async throws -> RecipientWithSortedPubKeys {
+        return try await pubLookup.fetchRemoteUpdateLocal(with: email)
+    }
```
tomholub, line 1089: These methods are no longer needed at all - the consumers in this class can be rewritten to use the direct methods like `await pubLookup.fetchRemoteUpdateLocal(with: email)` instead of `fetchContact(with email: String)`

tomholub (changes requested):

FlowCrypt/Controllers/Compose/ComposeViewController.swift
tomholub: Same probably applies to most of the rest as well

tomholub (approved):
This looks good to me 👍

ioanmo226:
@tomholub Can we merge this PR?

tomholub:
Yes - please mark it as ready for review, and it should auto-merge.

ioanmo226:
@tomholub Please merge PR

PR #1436, opened 2022-03-23
#1318 Check invalid email recipient in compose view

### LINKED ISSUE #1318: Add `X` next to each bad recipient
There should also be an `X` next to each bad recipient so that the user can tap them to remove. Would that be a separate issue?

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1315#issuecomment-1015765078_

ioanmo226:
This PR checked invalid email recipient in compose view and prevent adding invalid emails.

close #1318  // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Compose/ComposeViewController.swift
+++ b/FlowCrypt/Controllers/Compose/ComposeViewController.swift
@@ -866,14 +866,14 @@ extension ComposeViewController {
             } : nil
         )
         .onShouldReturn { [weak self] textField -> (Bool) in
-            if let isValid = self?.isTextFieldValidEmail(textField: textField), isValid {
+            if let isValid = self?.showAlertIfTextFieldNotValidEmail(textField: textField), isValid {
                 textField.resignFirstResponder()
                 return true
             }
             return false
         }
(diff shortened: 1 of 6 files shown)


FlowCrypt/Controllers/Compose/ComposeViewController.swift
```diff
@@ -879,6 +888,14 @@ extension ComposeViewController {
         }
     }
 
+    private func isTextFieldValidEmail(textField: UITextField) -> Bool {
```
tomholub, line 891: Let's rename this to `showAlertIfTextFieldNotValidEmail`. It can still return `isValid` which will be made obvious with the `if let isValid`.

It's better to track the behavior of methods in their names, even if that makes for some weird names: fewer surprises.

tomholub (commented):
Thank you for the test 👍 please do a small adjustment as below

FlowCrypt/Controllers/Compose/ComposeViewController.swift
ioanmo226: 👍

tomholub (approved):

PR #1440, opened 2022-03-24
#840 Use user email as login hint if current email is set

### LINKED ISSUE #840: use login hint for second sign-in if we already know the user email

ioanmo226:
This PR used user email as login hint if current email is set. (ex: when user is in compose screen and wants to grant contacts permission) 

close #840   // if it doesn't close the issue yet

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Services/GoogleUserService.swift
+++ b/FlowCrypt/Functionality/Services/GoogleUserService.swift
@@ -13,7 +13,7 @@ import GTMAppAuth
 import RealmSwift
 
 protocol UserServiceType {
-    func signIn(in viewController: UIViewController, scopes: [GoogleScope]) async throws -> SessionType
+    func signIn(in viewController: UIViewController, scopes: [GoogleScope], userEmail: String?) async throws -> SessionType
     func renewSession() async throws
 }
 
@@ -78,6 +78,7 @@ protocol AppDelegateGoogleSesssionContainer {
 // todo - should be refactored to not require currentUserEmail
 final class GoogleUserService: NSObject, GoogleUserServiceType {
 
+    @available(*, deprecated, message: "This variable will be removed in the near future.")
     let currentUserEmail: String?
     var appDelegateGoogleSessionContainer: AppDelegateGoogleSesssionContainer?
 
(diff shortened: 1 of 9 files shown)


FlowCrypt/Functionality/Services/GoogleUserService.swift
```diff
@@ -193,13 +193,17 @@ extension GoogleUserService: UserServiceType {
 extension GoogleUserService {
 
     private func makeAuthorizationRequest(scopes: [GoogleScope]) -> OIDAuthorizationRequest {
-        OIDAuthorizationRequest(
+        var additionalParameters = ["include_granted_scopes": "true"]
+        if let currentUserEmail = currentUserEmail {
+            additionalParameters["login_hint"] = currentUserEmail
+        }
```
tomholub, line 199: Please pass this as a parameter. to the method I'm trying to avoid this pattern of "methods getting their own stuff internally", instead values should be always dependency-injected through method parameters going forward.

tomholub (changes requested):

FlowCrypt/Functionality/Services/GoogleUserService.swift
ioanmo226: Then what's the usecase of `currentUserEmail` variable in GoogleUserService?

tomholub: I really, really want to get rid of it, completely if I can. Or mostly, if I cannot remove it completely.

But because it's still used across 30-40% of the app, it's a slow process. But it will have to involve not adding more usages :-)

Sorry this wasn't documented. Please add a comment on that property that we are trying to deprecate it. Actually, if there is a swift feature that you could mark it with, to produce warnings on all usages, that would be even better.

tomholub: The right way to get current user, under the newer design intention, is to get it from the VC which has it as an input from the global router.

From VC, it goes through method parameters all the way down to wherever it's needed.

ioanmo226: Ok, got it

ioanmo226: image


What do you think?

ioanmo226: @tomholub

tomholub: looks good

tomholub (approved):
👍

ioanmo226:
@tomholub I can see that you created separate issue (#1443) for login_hint issue in login screen.
In this case, I think we can close #840 once this PR is merged?

tomholub:
Unhappy tests, though. Maybe they expected the extra step of choosing the email on contacts prompt.

ioanmo226:
What do you mean unhappy tests?
> Unhappy tests

tomholub:
I mean this 

image

ioanmo226:
> I mean this
> 
> image

I see

PR #1441, opened 2022-03-24
#995 Side menu folder items capitalized + slow load issue

### LINKED ISSUE #995: menu folder items unnecessarily capitalized + slow load each time
(small) regression:
 - menu items (folders) now take visibly longer to load
 - it says `INBOX` instead of `Inbox` etc (had to adjust appium test for it)
Will file another issue.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/985#issuecomment-966301292_

ioanmo226:
This PR fixed side menu folder items capitalized + slow load issue.

close #995 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Services/Folders Services/RemoteFoldersProviderType/GmailService+folders.swift
+++ b/FlowCrypt/Functionality/Services/Folders Services/RemoteFoldersProviderType/GmailService+folders.swift
@@ -62,7 +62,7 @@ private extension Folder {
 
         self.init(
             path: path,
-            name: name,
+            name: name.capitalized,
             image: nil
         )
     }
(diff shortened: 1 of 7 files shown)


FlowCrypt/Functionality/Services/Folders Services/RemoteFoldersProviderType/GmailService+folders.swift
```diff
@@ -62,7 +62,7 @@ private extension Folder {
 
         self.init(
             path: path,
-            name: name,
+            name: name.capitalized,
```
tomholub, line 65: I worry, this will have to be more complicated. Because if I have a custom label called `wHaTeVer` than I want it displayed exactly that way. But `INBOX` should be `Inbox`, in English.

The logic will have to be as follows:
 - if the label is a standard Gmail label (which you want to hard-code a list of them), then translate the label into local language. For now we only have English. That would be `Inbox`, `Sent`, `All Mail` and so on. The translations should be in the strings file.
 - for other labels, show them exactly as they are

List of standard labels:

image

This may affect tests, which have to be updated to look for `Inbox` instead of `INBOX`, possibly.

ioanmo226: Nice idea. let me follow your suggestion

tomholub (approved):
👍

PR #1444, opened 2022-03-24
#1172 Completely remove foreach

### LINKED ISSUE #1172: replace forEach with for in
I'd like to ban `.forEach` in this codebase. Let's instead use`for thing in things {`, in this case `for newKeyToAdd in newKeysToAdd {`. I'll make another issue, to do this across the codebase.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1165#discussion_r762539544_

ioanmo226:
This PR removed foreach in all codebase.

close #1172 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
+++ b/FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
@@ -244,21 +244,17 @@ extension SetupManuallyEnterPassPhraseViewController {
         )
 
         if storageMethod == .memory {
-            try keysToUpdate
-                .map {
-                    PassPhrase(value: passPhrase, fingerprintsOfAssociatedKey: $0.fingerprints)
-                }
-                .forEach {
-                    try appContext.passPhraseService.updatePassPhrase(with: $0, storageMethod: storageMethod)
-                }
+            for newPassPhrase in (keysToUpdate.map {
+                PassPhrase(value: passPhrase, fingerprintsOfAssociatedKey: $0.fingerprints)
+            }) {
+                try appContext.passPhraseService.updatePassPhrase(with: newPassPhrase, storageMethod: storageMethod)
+            }
 
-            try newKeysToAdd
-                .map {
-                    PassPhrase(value: passPhrase, fingerprintsOfAssociatedKey: $0.fingerprints)
-                }
-                .forEach {
-                    try appContext.passPhraseService.savePassPhrase(with: $0, storageMethod: storageMethod)
-                }
+            for newPassPhrase in (newKeysToAdd.map {
+                PassPhrase(value: passPhrase, fingerprintsOfAssociatedKey: $0.fingerprints)
+            }){
+                try appContext.passPhraseService.savePassPhrase(with: newPassPhrase, storageMethod: storageMethod)
+            }
         }
 
         hideSpinner()
--- a/FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
+++ b/FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
@@ -244,15 +244,17 @@ extension SetupManuallyEnterPassPhraseViewController {
         )
 
         if storageMethod == .memory {
-            for newPassPhrase in (keysToUpdate.map {
+            let updatedPassPhrases = keysToUpdate.map {
                 PassPhrase(value: passPhrase, fingerprintsOfAssociatedKey: $0.fingerprints)
-            }) {
-                try appContext.passPhraseService.updatePassPhrase(with: newPassPhrase, storageMethod: storageMethod)
+            }
+            for updatedPassPhrase in updatedPassPhrases {
+                try appContext.passPhraseService.updatePassPhrase(with: updatedPassPhrase, storageMethod: storageMethod)
             }
 
-            for newPassPhrase in (newKeysToAdd.map {
+            let newPassPhrases = newKeysToAdd.map {
                 PassPhrase(value: passPhrase, fingerprintsOfAssociatedKey: $0.fingerprints)
-            }){
+            }
+            for newPassPhrase in newPassPhrases {
                 try appContext.passPhraseService.savePassPhrase(with: newPassPhrase, storageMethod: storageMethod)
             }
         }
--- a/FlowCrypt/Functionality/DataManager/Local Storage/LocalStorage.swift
+++ b/FlowCrypt/Functionality/DataManager/Local Storage/LocalStorage.swift
@@ -50,10 +50,8 @@ extension LocalStorage: LogOutHandler {
     }
 
     func cleanup() {
-        Constants.allCases
-            .map(\.rawValue)
-            .forEach {
-                storage.removeObject(forKey: $0)
-            }
+        for key in Constants.allCases.map(\.rawValue) {
+            storage.removeObject(forKey: key)
+        }
     }
 }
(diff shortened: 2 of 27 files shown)


FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
```diff
@@ -244,21 +244,17 @@ extension SetupManuallyEnterPassPhraseViewController {
...
-            try keysToUpdate
-                .map {
-                    PassPhrase(value: passPhrase, fingerprintsOfAssociatedKey: $0.fingerprints)
-                }
-                .forEach {
-                    try appContext.passPhraseService.updatePassPhrase(with: $0, storageMethod: storageMethod)
-                }
+            for newPassPhrase in (keysToUpdate.map {
+                PassPhrase(value: passPhrase, fingerprintsOfAssociatedKey: $0.fingerprints)
+            }) {
```
tomholub, line 249: If it reasonably fits on one line, then let's define it as:

```swift
for newPassPhrase in keysToUpdate.map { PassPhrase(value: passPhrase, fingerprintsOfAssociatedKey: $0.fingerprints) } {
  // ...
}
```

If it doesn't, then let's avoid multiline `for-in` header. In that case, let's define a variable.

```swift
let newPassPhrases = keysToUpdate.map { PassPhrase(value: passPhrase, fingerprintsOfAssociatedKey: $0.fingerprints) }
for newPassPhrase in  {
  // ...
}
```

In this particular case though, we could I guess simplify a bit as follows: add an overload `PassPhrase` init as follows: `init(value: passPhrase, associatedKey: Key)` (or similar)

Then you could comfortably on one line:

```swift
for newPassPhrase in keysToUpdate.map { PassPhrase(value: passPhrase, associatedKey: $0) } {
  // ...
}
```

tomholub: The PR looks good overall, just need to figure out the exact style - a multi-line for-in definition doesn't read well:

```swift
for x in (y.map {
  // ...
}) {

}
```

These two options read better:

```swift
val x = abc.map { $0.d }
for y in x {
  zz
}
```

or 

```swift
for y in abc.map { $0.d } {
  zz
}
```

tomholub (commented):
Thank you - see comments

FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
ioanmo226: I agree.
I will choose this style because if we put in one line, line length will be super long for some cases (which is not good for reading)
```
let x = abc.map { $0.d }
for y in x {
  zz
}

```

tomholub: consistency is good. Looks good to me

FlowCrypt/Functionality/DataManager/Local Storage/LocalStorage.swift
```diff
@@ -50,10 +50,8 @@ extension LocalStorage: LogOutHandler {
...
 
     func cleanup() {
-        Constants.allCases
-            .map(\.rawValue)
-            .forEach {
-                storage.removeObject(forKey: $0)
-            }
+        for key in Constants.allCases.map(\.rawValue) {
+            storage.removeObject(forKey: key)
+        }
```
tomholub, line 55: I find this much clearer :+1: even if it's sometimes a bit more code. It could be due to what I'm used to. But jumping between codebases to do review, consistency really helps me understand the code faster.

FlowCrypt/Functionality/Services/Local Contacts Service/LocalContactsProvider.swift
```diff
@@ -86,14 +86,13 @@ extension LocalContactsProvider: LocalContactsProviderType {
...
 
-        try recipient.pubKeys
-            .forEach { pubKey in
-                if let index = recipientObject.pubKeys.firstIndex(where: { $0.primaryFingerprint == pubKey.fingerprint }) {
-                    try update(pubKey: pubKey, for: recipientObject, at: index)
-                } else {
-                    try add(pubKey: pubKey, to: recipientObject)
-                }
+        for pubKey in recipient.pubKeys {
+            if let index = recipientObject.pubKeys.firstIndex(where: { $0.primaryFingerprint == pubKey.fingerprint }) {
```
tomholub, line 90: This is also nicer

FlowCrypt/Functionality/Services/Local Contacts Service/LocalContactsProvider.swift
```diff
@@ -121,14 +120,17 @@ extension LocalContactsProvider: LocalContactsProviderType {
...
             .pubKeys
             .filter { $0.primaryFingerprint == fingerprint }
-            .forEach { key in
-                try storage.write {
-                    storage.delete(key)
-                }
+        guard let keys = keys else {
+            return
+        }
+        for key in keys {
```
tomholub, line 129: A little bit closer to original code, and also shorter, could be `for key in keys ?? [] {` (assuming `[]` is empty array in swift). That's what I do in TypeScript. But either is fine.

FlowCryptAppTests/Functionality/Services/ComposeMessageServiceTests.swift
```diff
@@ -373,7 +373,7 @@ class ComposeMessageServiceTests: XCTestCase {
 
     func testSuccessfulMessageValidation() async throws {
         encryptedStorage.getKeypairsResult = [keypair]
-        recipients.enumerated().forEach { element, index in
+        for _ in recipients {
```
tomholub, line 376: This looks odd. Not your fault - it highlights that it was originally odd. I think we should review this test in another issue.

FlowCryptUI/Cell Nodes/ThreadMessageInfoCellNode.swift
```diff
@@ -247,8 +247,8 @@ public final class ThreadMessageInfoCellNode: CellNode {
         senderNode.accessibilityIdentifier = "aid-sender-\(input.index)"
         dateNode.accessibilityIdentifier = "aid-date-\(input.index)"
 
-        [senderNode, recipientButtonNode, senderNode, dateNode]
-            .forEach { $0.isAccessibilityElement = true }
+        let nodes = [senderNode, recipientButtonNode, senderNode, dateNode]
+        for node in nodes { node.isAccessibilityElement = true }
```
tomholub, line 251: Please use it's own block for the body

tomholub (commented):
:+1: with some more updates to do

FlowCryptAppTests/Functionality/Services/ComposeMessageServiceTests.swift
ioanmo226: ok

FlowCrypt/Functionality/Services/Local Contacts Service/LocalContactsProvider.swift
ioanmo226: We can't do like above for this kind because keys is not normal array.
I think better to add `guard` for this case.

image

tomholub: ah, got it

tomholub (dismissed):

ioanmo226:
@tomholub I had to refactor some codes which was badly written. Please review.

tomholub (approved):

tomholub (approved):

PR #1458, opened 2022-03-28
#1453 Focus textfield for reply/forward

### LINKED ISSUE #1453: after reply, message should be in focus
when I open a message and tap reply, currently I will end up in focus (caret), which then gets collapsed (correctly). But in this situation, caret should be on the beginning of the reply message I'm composing.

This is different from forward - for forward, it's correct that the recipient is in focus. (and recipients not in collapsed view)

Please add a test for it - for any of the tests that tap reply, verify where the caret is. Similarly for forward.

ioanmo226:
This PR focuses textfield for reply/forward in ComposeView

close #1453 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Compose/ComposeViewController.swift
+++ b/FlowCrypt/Controllers/Compose/ComposeViewController.swift
@@ -802,7 +802,7 @@ extension ComposeViewController {
                 $0.textView.attributedText = mutableString
                 let textView = $0
                 if input.isReply {
-                    DispatchQueue.main.asyncAfter(deadline: .now() + 0.05) {
+                    DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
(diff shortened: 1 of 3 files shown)


tomholub (approved):
Looks good. Please later investigate the failing test.

I wonder if we are going over the configured semaphore run limit for each CI run.

ioanmo226:
@tomholub I think so because tests run successfully in my local.

tomholub:
Let's see. Sometime in the future, we'll split the ui tests up to run on two machines, and take less time that way.

ioanmo226:
OK

ioanmo226:
@tomholub I'm stuck with testing because there seems no way to check if element is focused or not in Appium.
I'm finding hacky way now.

tomholub (approved):

PR #1461, opened 2022-03-29
#1348 EnterpriseServerApi refactor

### LINKED ISSUE #1348: refactor EnterpriseServerApi for single-fetching of fesUrl
I think that after finishing and merging https://github.com/FlowCrypt/flowcrypt-ios/pull/1331 we should also update `EnterpriseServerApi` to have `email` property (which will be passed from `AppContextWithUser`), so we don't need to pass it manually for every method.
Then we'll be able to fetch `fesUrl` only once and store result locally, while currently it's fetched for each request.

_Originally posted by @sosnovsky in https://github.com/FlowCrypt/flowcrypt-ios/issues/1342#issuecomment-1024041109_

sosnovsky:
This PR adds `email` property to `EnterpriseServerApi` to be able to fetch `fesUrl` only once

close #1348

- Tests added or updated - updated `EnterpriseServerApiTests`

### To be filled by reviewers


FlowCrypt/Functionality/Services/GlobalRouter.swift
```diff
@@ -171,8 +171,10 @@ extension GlobalRouter: GlobalRouterType {
             return
         }
 
-        let appContextWithUser = appContext.withSession(session: session, authType: authType, user: user)
-        AppStartup(appContext: appContextWithUser).initializeApp(window: keyWindow)
+        Task {
+            let appContextWithUser = try await appContext.with(session: session, authType: authType, user: user)
+            AppStartup(appContext: appContextWithUser).initializeApp(window: keyWindow)
+        }
```
tomholub, line 177: This could conceivably throw. There should be a do/catch and a way to render the problem and retry, eg with an alert.

FlowCryptAppTests/Functionality/Services/ComposeMessageServiceTests.swift
```diff
@@ -207,7 +209,7 @@ class ComposeMessageServiceTests: XCTestCase {
 
     func testValidateMessageInputWithAllEmptyRecipientPubKeys() async {
         encryptedStorage.getKeypairsResult = [keypair]
-        for recipient in recipients {
+        for _ in recipients {
             localContactsProvider.retrievePubKeysResult = { _ in
```
tomholub, line 213: I suppose the `for _ in recipients` can be removed here and a few more times below. Don't see reason to call `ocalContactsProvider.retrievePubKeysResult =` more than once.

tomholub (commented):
early review

FlowCryptAppTests/Functionality/Services/ComposeMessageServiceTests.swift
sosnovsky: Yeah, good point - no need to call it for each recipient 👍

FlowCrypt/App/AppContext.swift
```diff
@@ -129,30 +119,42 @@ class AppContext {
...
     let authType: AuthType
     let user: User
-    let userId: UserId
+
+    let enterpriseServer: EnterpriseServerApiType
+    let clientConfigurationService: ClientConfigurationServiceType
+
+    var userId: UserId {
+        UserId(email: user.email, name: user.name)
+    }
```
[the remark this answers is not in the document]
sosnovsky: Yes, it's a computed property, so it's re-evaluated on each call. 
Doesn't needed here, I'll replace it with `let` evaluated on `AppContext` init

FlowCrypt/Functionality/Services/GlobalRouter.swift
sosnovsky: Throwing is handled here, as `func proceed(with appContext: AppContext, session: SessionType)` is marked with `throws`. 
So calling it will need `do/catch` structure, currently it's implemented in `func signIn(appContext: AppContext, route: GlobalRoutingType)` - https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-1348-enterprise-server-refactor/FlowCrypt/Functionality/Services/GlobalRouter.swift#L83

tomholub: Is it possible to actually propagate a thrown exception from inside `Task { }` to the upper method? I was of the understanding that the error will just be ignored if there isn't a `do/catch` inside that Task.

sosnovsky: Yes, my bad, exception isn't propagated to parent method - it was called from another place when I tried to reproduce this issue. I'll add exception handling here

FlowCrypt/Functionality/Services/Account Server Services/EnterpriseServerApi.swift
```diff
@@ -48,52 +44,21 @@ class EnterpriseServerApi: NSObject, EnterpriseServerApiType {
...
+        self.email = email
 
-            guard isExpectedFesServiceResponse(responseData: response.data) else {
-                if Bundle.isEnterprise() { // on enterprise build, FES is expected to be running
-                    throw AppErr.general("Unpexpected response from FlowCrypt Enterprise Server")
-                }
-                return nil // on consumer installations, we only use FES if it returns as expected
-            }
+        let helper = EnterpriseServerApiHelper()
+        self.fesUrl = try await helper.getActiveFesUrl(for: email)
```
tomholub, line 56: I wonder - could there be situations when `EnterpriseServerApi` is initiated but a call to check if FES is active is not actually needed in that moment? We could maybe instead have `fesUrl` as a lazy computed property?

tomholub (approved):
I'm approving but would like to discuss the below point in another issue.

PR #1463, opened 2022-03-30
#1442 Added retry logic for get threads list and thread info

### LINKED ISSUE #1442: request timed out when loading inbox
image

We do a lot of requests at once when loading inbox, and it may sometimes time out.

When it times out, we seem to fail the whole loading and then we need to make all of the requests again.

I think it would be better if only the email messages that timed out had the failure showing on them, with a retry button.

Then retrying one of them would retry all of the failed ones. That's assuming that the "list of messages" API call succeeded, and it was some individual messages that failed. The initial list of messages (returned successfully) could result in "empty list" of messages already pre-rendered, and then individual message details could be rendered into that list as they become available. That way, the loading will become more interactive.

This way, at least, most of the inbox will be loaded when timeout happens on one or two messages.

After that, we can investigate in more detail how to make this happen less often. Maybe some automatic retry approach.

ioanmo226:
This PR added retry logic for Task level and added this logic in get threads list and get thread info calls.
Default retry count is `3` and can be changed by passing `maxRetryCount` parameter.

close #1442 // if this PR closes an issue

- Difficult to test (not sure what & how to test)

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Error Handling/Task+Retry.swift
+++ b/FlowCrypt/Functionality/Error Handling/Task+Retry.swift
@@ -15,7 +15,7 @@ extension Task where Failure == Error {
         operation: @Sendable @escaping () async throws -> Success
     ) -> Task {
         Task(priority: priority) {
-            for _ in 0..<maxRetryCount {
+            for _ in 0 ..< maxRetryCount {
                 do {
                     return try await operation()
                 } catch {
--- a/FlowCrypt/Functionality/Error Handling/Task+Retry.swift
+++ b/FlowCrypt/Functionality/Error Handling/Task+Retry.swift
@@ -22,8 +22,6 @@ extension Task where Failure == Error {
                     let oneMillisecond = UInt64(1_000_000)
                     let delay = oneMillisecond * retryDelayMs
                     try await Task<Never, Never>.sleep(nanoseconds: delay)
-
-                    continue
                 }
             }
 
--- a/FlowCrypt/Functionality/Error Handling/Task+Retry.swift
+++ b/FlowCrypt/Functionality/Error Handling/Task+Retry.swift
@@ -10,7 +10,7 @@ extension Task where Failure == Error {
     @discardableResult
     static func retrying(
         priority: TaskPriority? = nil,
-        maxRetryCount: Int = 3,
+        maxRetryCount: Int = 2,
         retryDelay: TimeInterval = 1,
         operation: @Sendable @escaping () async throws -> Success
     ) -> Task {
--- a/FlowCrypt/Functionality/Error Handling/Task+Retry.swift
+++ b/FlowCrypt/Functionality/Error Handling/Task+Retry.swift
@@ -11,7 +11,7 @@ extension Task where Failure == Error {
     static func retrying(
         priority: TaskPriority? = nil,
         maxRetryCount: Int = 2,
-        retryDelay: TimeInterval = 1,
+        retryDelayMs: UInt64 = 1000,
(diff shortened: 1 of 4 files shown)


mobilestar1:
For above case, when 19 threads are loaded and last one fails, app retries to fetch only that last message(not whole 20 threads).
App only tries to re-fetch failed threads (up to 3 retries).

image

ioanmo226:
For above case, when 19 threads are loaded and last one fails, app retries to fetch only that last message(not whole 20 threads).
App only tries to re-fetch failed threads (up to 3 retries).

image

tomholub:
That's better. I'll take another look.

FlowCrypt/Functionality/Error Handling/Task+Retry.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+//  Created by Ioan Moldovan on 3/30/22
+//  Copyright © 2017-present FlowCrypt a. s. All rights reserved.
+//
+
+extension Task where Failure == Error {
+    @discardableResult
+    static func retrying(
+        priority: TaskPriority? = nil,
+        maxRetryCount: Int = 3,
```
tomholub, line 13: Let's do a default of 2. So three API calls total maximum. Current setting with 3 may cause up to 4 API calls which is a bit too much.

FlowCrypt/Functionality/Error Handling/Task+Retry.swift
```diff
@@ -0,0 +1,34 @@
...
+        Task(priority: priority) {
+            for _ in 0 ..< maxRetryCount {
+                do {
+                    return try await operation()
+                } catch {
+                    let oneSecond = TimeInterval(1_000_000_000)
+                    let delay = UInt64(oneSecond * retryDelay)
+                    try await Task<Never, Never>.sleep(nanoseconds: delay)
+
+                    continue
```
tomholub, line 26: If you are at the end of a loop block, I don't think `continue` would have any effect.

tomholub (changes requested):
This looks good - please see comments below

FlowCrypt/Functionality/Error Handling/Task+Retry.swift
```diff
@@ -0,0 +1,34 @@
...
+//  Created by Ioan Moldovan on 3/30/22
+//  Copyright © 2017-present FlowCrypt a. s. All rights reserved.
+//
+
+extension Task where Failure == Error {
+    @discardableResult
+    static func retrying(
+        priority: TaskPriority? = nil,
+        maxRetryCount: Int = 3,
+        retryDelay: TimeInterval = 1,
```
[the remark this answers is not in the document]
ioanmo226: I agree

FlowCrypt/Functionality/Error Handling/Task+Retry.swift
ioanmo226: OK

FlowCrypt/Functionality/Error Handling/Task+Retry.swift
ioanmo226: Right. I added `continue` to make intent more clear.
Do you want me to remove that?

tomholub: yes - please remove

ioanmo226: OK

tomholub (approved):

PR #1464, opened 2022-03-31
#1455 Display recipient info popup when user tap recipients

### LINKED ISSUE #1455: add a way to see recipient email address on compose
Ever since we have used names for recipients, it's no longer possible to figure out which email address I'm emailing to on compose screen.

Gmail app has a solution for this - when I tap the recipient bubble, it shows me the name and email address.

I'd like something similar.

ioanmo226:
This PR changed way how recipients are displayed when user taps recipient

close #1455 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/appium/tests/screenobjects/new-message.screen.ts
+++ b/appium/tests/screenobjects/new-message.screen.ts
@@ -247,6 +247,16 @@ class NewMessageScreen extends BaseScreen {
     const addedRecipientEl = await $(`~aid-${type}-${order}-label`);
     await ElementHelper.waitAndClick(addedRecipientEl);
     await ElementHelper.waitAndClick(await this.recipientPopupRemoveButton);
+    await ElementHelper.waitElementInvisible(addedRecipientEl);
+  }
+
+  checkCopyForAddedRecipient = async (email: string, order: number, type = 'to') => {
+    await this.showRecipientInputIfNeeded();
+    const addedRecipientEl = await $(`~aid-${type}-${order}-label`);
+    await ElementHelper.waitAndClick(addedRecipientEl);
+    await ElementHelper.waitAndClick(await this.recipientPopupCopyButton);
+    const base64Encoded = new Buffer(email).toString('base64');
+    expect(await driver.getClipboard('plaintext')).toEqual(base64Encoded);
   }
 
   checkAddedAttachment = async (name: string) => {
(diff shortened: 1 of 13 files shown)


ioanmo226:
@tomholub MVP version is ready (There are still many bugs and no functions/tests are added yet).
I just want you to confirm if I'm in the correct direction.

tomholub:
Yes - that was exactly what I meant. Looks good.

tomholub:
When you do copy-all, I suppose it could be comma-separated list of recipients.

ioanmo226:
Yeah, correct. Okay got it.

> When you do copy-all, I suppose it could be comma-separated list of recipients. But you don't necessarily have to implement copy-all. Just copy would do.


By the way, I think there should also be button which will refresh recipient key info. (something like `Refresh Key`).
What do you think?

ioanmo226:
@tomholub Please confirm above and let me know what other features you want me to add in that popup.

tomholub:
> @tomholub Please confirm above and let me know what other features you want me to add in that popup.

Less is better. We can always add more later if we deem it needed.

ioanmo226:
Okay, got it

tomholub:
I just thought of (needed) another item on that menu - a way to edit the recipient. For example, I made a type in the recipient, but I don't want to remove it and re-type the whole thing. I'll file another issue.

appium/tests/screenobjects/new-message.screen.ts
```diff
@@ -236,7 +246,17 @@ class NewMessageScreen extends BaseScreen {
     await this.showRecipientInputIfNeeded();
     const addedRecipientEl = await $(`~aid-${type}-${order}-label`);
     await ElementHelper.waitAndClick(addedRecipientEl);
-    await driver.sendKeys(['\b']); // backspace
+    await ElementHelper.waitAndClick(await this.recipientPopupRemoveButton);
+    await ElementHelper.waitElementInvisible(addedRecipientEl);
```
tomholub, line 250: Both backspace and tap+remove should work, so we should ideally test both. Let's also add back the test for `await driver.sendKeys(['\b']); // backspace` here, in another issue.

appium/tests/specs/live/composeEmail/CheckRecipientPopup.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    await NewMessageScreen.setAddRecipient(recipient1);
+    await NewMessageScreen.setAddRecipient(recipient2);
+    await NewMessageScreen.setAddRecipient(recipient3);
+
+    await NewMessageScreen.checkCopyForAddedRecipient(recipient1, 0);
+
+    await NewMessageScreen.deleteAddedRecipient(2);
+    await NewMessageScreen.deleteAddedRecipient(1);
+    await NewMessageScreen.deleteAddedRecipient(0);
```
tomholub, line 27: In all of the popups, we should also check it displays expected email and name. Can be in another issue.

PR #1472, opened 2022-04-05
#1470 Added recipient edit feature

### LINKED ISSUE #1470: a way to edit recipient email by tapping on it and choose edit
I just thought of (needed) another item on that menu - a way to edit the recipient. For example, I made a type in the recipient, but I don't want to remove it and re-type the whole thing. I'll file another issue.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1464#issuecomment-1087893368_

ioanmo226:
This PR added recipient edit feature.

close #1470 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


FlowCrypt/Assets.xcassets/edit.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
+{
```
tomholub, line 1: Normally, similar icons would already be present from standard iOS icons. Could you check there first? So we wouldn't have to add our own.

FlowCrypt/Resources/en.lproj/Localizable.strings
```diff
@@ -129,6 +129,7 @@
 "compose_invalid_recipient_title" = "Invalid recipient";
 "compose_invalid_recipient_message" = "Please enter a valid email address.";
 "compose_recipient_copy" = "Copy";
+"compose_recipient_edit" = "Edit recipient";
```
tomholub, line 132: Since the other options are `Copy` and `Remove` then let's do simply `Edit`

tomholub (approved):

tomholub:
Looks good, once tests are passing

PR #1481, opened 2022-04-07
#1478 Add swift tests for how long it takes to do operations with rsa4096 key

### LINKED ISSUE #1478: add swift tests for how long it takes to do operations with rsa4096 key
Also need to add a Swift test, that uses the `Core`, that measures it takes less than 1 second each to:
 - sign message with rsa4096
 - encrypt message with rsa4096
 - decrypt message with rsa4096
 - encrypt rsa4096 key
 - decrypt rsa4096 key

These will prevent regression once we do #1362 .

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1469#issuecomment-1087904833_

ioanmo226:
This PR added swift tests for how long it takes to do operations with rsa4096 key

close #1478 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


FlowCryptCommon/FlowcryptTimer.swift
```diff
@@ -0,0 +1,30 @@
...
+        endTime = CFAbsoluteTimeGetCurrent()
+    }
+
+    var duration: CFAbsoluteTime {
+        if let startTime = startTime, let endTime = endTime {
+            return endTime - startTime
+        }
+        return 0
+    }
+}
```
tomholub, line 30: Let's add this into test utilities only, and call it `TestTimer`

FlowCryptCommon/FlowcryptTimer.swift
```diff
@@ -0,0 +1,30 @@
...
+
+    func start() {
+        startTime = CFAbsoluteTimeGetCurrent()
+    }
+
+    func stop() {
+        endTime = CFAbsoluteTimeGetCurrent()
+    }
+
+    var duration: CFAbsoluteTime {
```
tomholub, line 24: Let's add units for clarity. Physics values should always come with units.

Let's make this `durationMs`

tomholub (commented):
Please see comments - else looks good. Thanks!

FlowCryptCommon/FlowcryptTimer.swift
ioanmo226: ok

FlowCryptCommon/FlowcryptTimer.swift
ioanmo226: ok

ioanmo226:
By the way, @tomholub As I checked this code block takes about 1-2 minutes to execute for rsa4096.
Is this expected or we need to fix and add unit test for this too?

        let generateKeyRes = try await core.generateKey(
            passphrase: passphrase,
            variant: KeyVariant.rsa4096,
            userIds: [UserId(email: "test@example.com", name: "Test user")]
        )

tomholub:
> By the way, @tomholub As I checked this code block takes about 1-2 minutes to execute for rsa4096. Is this expected or we need to fix and add unit test for this too?
> 
> ```
>     let generateKeyRes = try await core.generateKey(
>         passphrase: passphrase,
>         variant: KeyVariant.rsa4096,
>         userIds: [UserId(email: "test@example.com", name: "Test user")]
>     )
> ```

I won't be optimizing that - instead we are generating ECC keys whenever keygen needs to happen on the phone.

To speed up the tests, use the zip sample I shared in the issue.

ioanmo226:
> > By the way, @tomholub As I checked this code block takes about 1-2 minutes to execute for rsa4096. Is this expected or we need to fix and add unit test for this too?
> > ```
> >     let generateKeyRes = try await core.generateKey(
> >         passphrase: passphrase,
> >         variant: KeyVariant.rsa4096,
> >         userIds: [UserId(email: "test@example.com", name: "Test user")]
> >     )
> > ```
> 
> I won't be optimizing that - instead we are generating ECC keys whenever keygen needs to happen on the phone.
> 
> To speed up the tests, use the zip sample I shared in the issue.

ok

tomholub:
@ioanmo226 I'll merge it while it's passing & you can improve it in a followup PR today. Thanks!

tomholub (approved):

PR #1482, opened 2022-04-07
Added new methods for checking Value and Text in tests using waitUntil

### LINKED ISSUE #1448: always use waitForContent when testing for presense of text
- test content always with `element.waitForContent("expected content")`. This will keep checking for presense of expected content repeatedly for up to 15 seconds and then throw. That way, you avoid "testing too early" which is the cause of 95% of flakiness of UI tests. This is what we do on browser extension.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1056#issuecomment-974681036_

fcvakintos:
This PR contains 2 new methods for checking text and value in elements, The methods will wait 15 seconds until the text or value will present in element, after will return the error - "expected value/text to be different after 15s". When the text is appears before 15s, the methods will complete.
close #1448 

**Tests** 
- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/appium/tests/helpers/ElementHelper.ts
+++ b/appium/tests/helpers/ElementHelper.ts
@@ -60,6 +60,30 @@ class ElementHelper {
     await this.waitAndClick(element);
     await element.setValue(text);
   }
+
+  //wait for text in element during 15 seconds (if the text doesn't appear during 15s, it will show the error)
+  static waitForText = async (element: WebdriverIO.Element, text: string, timeout: number = DEFAULT_TIMEOUT) => {
+    await this.waitElementVisible(element);
+    await element.waitUntil(async  function () {
+      return (await element.getText() === text)
+    }, {
+      timeout: timeout,
+      timeoutMsg: `expected text within ${timeout}ms to be "${text}" but got last value "${await element.getText()}"`
+    });
+  }
+
+  //wait for value in element during 15 seconds (if the value doesn't appear during 15s, it will show the error)
+  static waitForValue = async (element: WebdriverIO.Element, value: string, timeout: number = DEFAULT_TIMEOUT) => {
+    await this.waitElementVisible(element);
+    await element.waitUntil(async  function () {
+      return (await element.getValue() === value)
+    }, {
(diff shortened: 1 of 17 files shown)


appium/tests/helpers/ElementHelper.ts
```diff
@@ -60,6 +60,29 @@ class ElementHelper {
...
+  }
+
+  //wait for value in element during 15 seconds (if the value doesn't appear during 15s, it will show the error)
+  static waitForValue = async (element: WebdriverIO.Element, value: string, timeout: number = DEFAULT_TIMEOUT) => {
+    await this.waitElementVisible(element);
+    await element.waitUntil(async  function () {
+      return (await element.getValue() === value)
+    }, {
+      timeout: timeout,
+      timeoutMsg: 'expected value to be different after 15s'
```
tomholub, line 82: Do `expected value within ${timeout}s to be "${value}" but got last value "${lastValue}"`

Similarly above

appium/tests/screenobjects/new-message.screen.ts
```diff
@@ -216,8 +217,12 @@ class NewMessageScreen extends BaseScreen {
...
     const recipientCell = await $(`~aid-${type}-${order}-label`);
     await ElementHelper.waitElementVisible(recipientCell);
-    const name = await recipientCell.getValue();
-    expect(name).toEqual(`  ${recipient}  `);
+    await recipientCell.waitUntil(async  function () {
+      return (await recipientCell.getValue() === `  ${recipient}  `)
+    }, {
+      timeout: 15,
+      timeoutMsg: 'expected text to be different after 15s'
+    });
```
tomholub, line 225: Should use the new method here

fcvakintos: do you mean another method?

tomholub: You've just implemented a `waitForValue` method. Why not use it here?

tomholub (approved):

PR #1496, opened 2022-04-12
#1468 Trigger evaluation when user tap away from typed recipient address 

### LINKED ISSUE #1468: tapping away from typed recipient address should trigger evaluation
reproduce:
1) open secure compose
2) enter a full address that is not in the contacts
3) tap below into empty space (currently occupied by contact search results that are empty)

Current behavior: nothing happens. I have to hit enter to have the recipient accepted / evaluated

Expected behavior: input should be evaluated

A simple fix would be that search results would be more visually differentiated, and would not actually take up the whole screen? Especially when there are only a few results, I would expect search results to only take up appropriate amount of space relative to how many results it's showing.

Then, tapping below the search results will naturally evaluate the input.

Should come with ui test.

ioanmo226:
This PR triggers evaluation when user tap away from typed recipient address 

close #1468 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/appium/tests/helpers/TouchHelper.ts
+++ b/appium/tests/helpers/TouchHelper.ts
@@ -1,4 +1,4 @@
-
+type ScreenPosition = 'topLeft' | 'topCenter' | 'topRight' | 'centerLeft' | 'centerCenter' | 'centerRight' | 'bottomLeft' | 'bottomCenter' | 'bottomRight';
 
 class TouchHelper {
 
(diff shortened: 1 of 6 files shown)


appium/tests/helpers/TouchHelper.ts
```diff
@@ -17,12 +17,42 @@ class TouchHelper {
...
+   * Tap Screen
    */
-  static dismissPopover = async () => {
+  static tapScreen = async (position: ScreenPosition) => {
     const {width, height} = await driver.getWindowSize();
+    let pressOptions;
+    switch (position) {
+      case 'topLeft':
+        pressOptions = {x: 0, y: 0};
+        break;
```
tomholub, line 28: @sosnovsky could you please look at this PR / this method for a review - in case a simpler solution exists. Thanks!

tomholub (commented):
Thanks - looks ok to me. Will ask Roma for a closer look before merging.

appium/tests/helpers/TouchHelper.ts
sosnovsky: Current solution looks good for me 👍
The only possible improvement I see - using `enum` for `ScreenPosition`, so we'll have autocomplete for positions and it won't be possible to mistype them.

ioanmo226: @sosnovsky Thanks for your review.
By the way, currently it also supports autocompletion and it will throw error if user mistyped ScreenPosition.

tomholub (dismissed):

tomholub (dismissed):

tomholub (approved):

Issue #1379, opened 2022-02-13
make OpenPGP.js v5 code work with iOS JavaScriptCore


tomholub:
https://github.com/FlowCrypt/flowcrypt-ios/pull/1374/files#r805180770

<img width="831" alt="image" src="https://user-images.githubusercontent.com/6306961/153776588-cdef8ebb-0abc-4a57-bad1-9f8475a0f24f.png">

tomholub:
<img width="806" alt="image" src="https://user-images.githubusercontent.com/6306961/153776732-c99b0be6-b5d7-48b5-83e1-87aeb0857f70.png">

tomholub:
<img width="800" alt="image" src="https://user-images.githubusercontent.com/6306961/153776756-d27c1f58-e7bd-4907-8a18-3d7210687164.png">

tomholub:
<img width="800" alt="image" src="https://user-images.githubusercontent.com/6306961/153776885-38d63811-6612-44a6-b51e-a0e1a600c8b8.png">

ivan-ushakov:
@tomholub

I tried to use new `flowcrypt-ios-prod.js.txt` and when I run application, I get
```
"❗️[Js][Core] SyntaxError: Unexpected token \',\'. Cannot declare a lexical variable twice: \'ReadableStream\'., 44400:undefined"
```

I want to continue with integration of OpenPGP library update in iOS, if you don't mind.

ivan-ushakov:
After commenting that line out (because I found that such variables were already defines above), I got this error:
```
"⚙️ Can\'t find variable: exports\n -> js global code@"
"❗️[Js][Core] ReferenceError: Can\'t find variable: exports, 43914:30"
```
So it looks like merging (or generating) of JavaScript code produces wrong code for JavaScriptCore engine.

ivan-ushakov:
Also I get error when I run `npm test`:
```
> flowcrypt-mobile-core@0.0.1 test /Users/ivan/Documents/Projects/FlowCrypt/flowcrypt-ios/Core
> ava build/ts/test.js --verbose

internal/modules/cjs/loader.js:981
  throw new ERR_REQUIRE_ESM(filename);
  ^

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/ivan/Documents/Projects/FlowCrypt/flowcrypt-ios/Core/node_modules/ava/entrypoints/cli.mjs
[...]

PR #1497, opened 2022-04-13
issue 1469 rsa sign perf

### LINKED ISSUE #1469: RSA 4096 sign too slow
Takes even a minute or two minutes, depending on device, for a simple message. Reported by customer.

I just tested sending to a recipient with rsa4096 and it was fast (my own key was ECC).

This suggests that the problem may be be when signing using rsa4096 - need to test.

tomholub:
This PR fixes slow RSA sign performance on iOS.

**Because our TypeScript files now use OpenPGP.js v5 which doesn't work when built for iOS yet, we have kept using the original OpenPGP.js v4 JavaScript file in the app. Therefore in this PR, I'm patching the JS file directly without any change to the TypeScript -> JavaScript build process or TypeScript source. Once we're working on integrating v5 #1379 , the added test will ensure that we'll have to tackle this at that time as well, so this change should not regress even though I'm patching a file that will get overridden in the future.**

close #1469 

### To be filled by reviewers


tomholub:
In my tests, the speed of RSA 4096 sign went from 30 seconds on my device to about 500ms.

sosnovsky:
> In my tests, the speed of RSA 4096 signing went from 30 seconds on my device to about 500ms.

Wow, it's a great improvement!

Issue #1386, opened 2022-02-17
Video reporter for tests not working


sosnovsky:
Video reporter fails to start

```
[0-2] 2022-02-17T14:17:41.362Z INFO @wdio/runner: Wait for 1 reporter to synchronise
[0-2] 2022-02-17T14:17:41.468Z INFO @wdio/runner: Wait for 1 reporter to synchronise
[0-2] 2022-02-17T14:17:41.568Z INFO @wdio/runner: Wait for 1 reporter to synchronise
[0-2] 2022-02-17T14:17:41.670Z INFO @wdio/runner: Wait for 1 reporter to synchronise
[0-2] 2022-02-17T14:17:41.775Z INFO @wdio/runner: Wait for 1 reporter to synchronise
[0-2] 2022-02-17T14:17:41.882Z ERROR @wdio/runner: Error: Some reporters are still unsynced: Video
[0-2]     at Timeout.<anonymous> (/Users/semaphore/git/flowcrypt-ios/appium/node_modules/@wdio/runner/build/reporter.js:87:35)
[0-2]     at listOnTimeout (node:internal/timers:559:17)
[0-2]     at processTimers (node:internal/timers:502:7)
```

sosnovsky:
I temporary disabled video reporter as it caused some tests to fail (https://flowcrypt.semaphoreci.com/jobs/76d5f042-6100-4dca-9669-2a1adcebd3dc), probably because of reporter timeout.

PR #1510, opened 2022-04-20
#841 Periodically refresh keys from EKM

### LINKED ISSUE #841: periodically refresh keys from EKM
part of #275 

Each time the app starts, if `clientConfiguration.usesKeyManager` is true, the app should refetch keys from EKM and update the local ones. The fetch should be in the background, not holding up the app during startup. Thrown network errors can be ignored but not other types of errors (which should show error modal on whichever screen will load next)

The update mechanism should be similar to what we do for updating received public keys:
 - for each received key, check if we have that key in storage, by primary fingerprint
 - if we don't have the key in storage, add it
 - if we do, check if the received key is more recent than the local key, by dateLastModified
 - if received key is newer, replace stored key with the received one
 - else if it's same or older, leave the stored key as is

HOWEVER, these being private keys, we have to ensure to encrypt them with pass phrase before storing. If the pass phrase (for any of the existing keys already in storage) is already available, then we use the same pass phrase to encrypt each key that we are updating, and then store encrypted.

If the pass phrase is not available at that moment, then we display a pass phrase prompt (for iOS this would be just a modal) with `Enter pass phrase to keep your account keys up to date`. There would be an ok and cancel button, and cancel would do nothing, meaning next time the app restarts, they should see the same.

Ui tests can be written for this as follows:

test 1: `app auto updates keys from EKM during startup with a pass phrase prompt`
 - // stage 1 - setup
 - set up mock FES and EKM
      - FES should have `FORBID_STORING_PASS_PHRASE`
      - EKM should have any one key returned
 - set up the app
 - go to app settings -> keys and observe which key it has
 - // stage 2 - prompt appears / wrong pass phrase rejected / cancel
 - change EKM mock to start returning one more key
 - restart the app
 - observe that it shows EKM pass phrase prompt. Enter wrong pass phrase in the prompt. Observe it gets rejected / prompt remains.
 - cancel the prompt. check settings -> keys and observe it stayed the same
 - // stage 3 - new key gets added
 - restart the app. 
 - enter the correct pass phrase. Observe that a toast message shows up with `Account keys updated`
 - go to settings -> keys and observe that it updated the keys
 - // stage 4 - modified key gets updated, removed key does not get removed
 - change EKM mock to start returning an updated version of one of the existing keys. Stop returning the other key
 - restart app, enter pass phrase into prompt, observe `Account keys updated`
 - go to settings -> keys. Observe that no key was removed, and then updated key was updated
 - // stage 5 - older version of key does not get updated
 - change EKM mock to start returning only the older version of that modified key again
 - restart app
 - observe no prompt shows
 - go to settings -> keys. Observe keys remain the same as in stage 4.
 - success

test 2: `app auto updates keys from EKM during startup without pass phrase prompt`
 - // stage 1 - setup
 - set up mock FES and EKM
      - FES should NOT have `FORBID_STORING_PASS_PHRASE`
      - EKM should have any one key returned
 - set up the app
 - go to app settings -> keys and observe which key it has
 - // stage 2 - keys get autoupdated
 - change EKM mock to start returning one more key
 - restart the app
 - observe that a toast message shows up with `Account keys updated`
 - go to settings -> keys and observe that a key was added
 - // stage 3 - nothing to update
 - restart the app
 - observe that no toast shows during startup
 - go to settings -> keys and see they are unchanged

test 3: `EKM key update errors handled gracefully`
 - // stage 1 - setup
 - set up mock FES and EKM, with one key returned by EKM
 - set up the app
 - go to app settings -> keys and observe which key it has
 - // stage 2 - EKM down
 - change EKM mock to start returning error 500
 - restart the app
 - observe that no modal shows during startup. Go to app settings -> keys, notice keys stayed the same
 - // stage 3 - error shown to user
 - change EKM mock to start returning a broken key (for example, only return the first half of the armored key text)
 - restart the app
 - observe that the app shows a modal during startup `Could not update keys from EKM due to error:` and followed by whatever error happened.
 - tap `close` on the modal and observe that inbox loads normally
 - go to app settings -> keys and observe they stayed the same

ioanmo226:
This PR periodically refresh keys from EKM.

close #841  // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


ioanmo226:
OHhhh.. I see.
Indeed I was implementing EKM mocks similar to that one. It would be waste of time if you didn't notify it. 
Thank you for your notice.

ioanmo226:
@sosnovsky, @tomholub, @fcvakintos  I currently struggle to restart app in mock UI tests. No documentation in appium or webdriver for this case.
When I restart app in mock testing using below code, process arguments ` 'args': ['--mock-fes-api', '--mock-attester-api']` are lost and as a result app acts like it works in live environment (not mock environment).

 ```
 await driver.terminateApp(bundleId);
 await driver.activateApp(bundleId);
```
Do you have idea how to resolve this issue?

**Comment by sosnovsky** [comment]
@ioanmo226 were you able to find any solution?

**Comment by ioanmo226** [comment]
no

**Comment by tomholub** [comment]
That looks like a likely bug in appium itself. 

https://github.com/appium/appium-ios-simulator/blob/0b2314eef24f40acb0786808002c76b6d470b5d5/lib/utils.js#L17 I don't see any sys args here 🤷 

You should report this, with clear examples, to appium as github issue.

**Comment by tomholub** [comment]
No sysargs showing here https://github.com/appium/appium-mac2-driver/blob/c474f7e3cfc198480a78f6fcc282b0ffc77cbf85/lib/commands/app-management.js#L35

**Comment by tomholub** [comment]
Maybe here also https://github.com/appium/WebDriverAgent/blob/a1824895274c60798b506f31054e4ce1e0400c71/WebDriverAgentLib/Commands/FBSessionCommands.m#L145

**Comment by tomholub** [comment]
As a workaround, in another PR, please add an item `Settings` -> `Experimental` -> `Reload app`. Tapping it will:
 - forget all in-memory pass phrases
 - remove all VC from stack
 - go through global router init as if the app has just started

The goal is that you should ideally not be rewriting any code that happens in normal app init - ideally you would just point the app at the bootstrapping code so that all of the app starting actions will naturaly happen as if the app was starting.

Then update your tests to use this mechanism instead of actually restarting the app.

**Comment by ioanmo226** [comment]
> That looks like a likely bug in appium itself.
> 
> You should report this, with clear examples, to appium as github issue.

Sure

**Comment by ioanmo226** [comment]
Submitted issue to Appium.
Here's the link.
https://github.com/appium/appium/issues/16852

> That looks like a likely bug in appium itself.
> 
> You should report this, with clear examples, to appium as github issue.

**Comment by ioanmo226** [comment]
@tomholub Appium team responded and they told me to use driver.execute function with arguments option.
Here's the solution.

https://github.com/appium/appium/issues/16852#issuecomment-1108161975

```
const restartApp = async () => {
  await driver.terminateApp(bundleId);
  const processArgs = ['--mock-fes-api', '--mock-attester-api']; // your process arguments
  const args = {
    bundleId, // your bundle id
    arguments: processArgs
  }
  await driver.execute('mobile: launchApp', args);
}
```

**Comment by ioanmo226** [comment]
Do you wanna me to use above solution or our new experimental reload feature?

**Comment by tomholub** [comment]
> Do you wanna me to use above solution or our new experimental reload feature?

Definitely use what appium suggests, looks good.

**Comment by ioanmo226** [comment]
Then do we still need our experimental reload feature?

**Comment by tomholub** [comment]
> Then do we still need our experimental reload feature?

If it works, we can consider keeping it for now. It may be useful for manual debugging. If it turns out not useful, then we can drop it later.

**Comment by ioanmo226** [comment]
I can see that you already closed reload app PR.

> If it works, we can consider keeping it for now. It may be useful for manual debugging. If it turns out not useful, then we can drop it later.

**Comment by tomholub** [comment]
Yes, since it was still not merged, I decided to close it.

File: FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift [hunk 1]
```diff
@@ -96,6 +103,167 @@ final class SideMenuNavigationController: ENSideMenuNavigationController {
...
+                            isNewKey: true
+                        )
+                        isKeyUpdated = true
+                        continue
+                    }
+                    guard let lastModified = keyDetail.lastModified else {
+                        continue
+                    }
+                    // Key exists in local. Check if saved key is outdated by checking lastModified and update if needed
+                    if savedLocalKey.lastModified ?? 0 < lastModified {
```
tomholub, line 137: Let's treat both of them consistently. Therefore I guess `savedLocalKey.lastModified ?? 0 < keyDetail.lastModified ?? 0`

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
```diff
@@ -96,6 +103,167 @@ final class SideMenuNavigationController: ENSideMenuNavigationController {
...
+    private func refreshKeysFromEKMIfNeeded(context: AppContextWithUser) {
+        Task {
+            do {
+                let configuration = try await context.clientConfigurationService.configuration
+                guard configuration.checkUsesEKM() == .usesEKM else {
+                    return
+                }
+                let emailKeyManagerApi = EmailKeyManagerApi(clientConfiguration: configuration)
+                let idToken = try await IdTokenUtils.getIdToken(userEmail: context.user.email)
+                let keydDetails = try await emailKeyManagerApi.getPrivateKeys(idToken: idToken)
```
tomholub, line 117: typo

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
```diff
@@ -96,6 +103,167 @@ final class SideMenuNavigationController: ENSideMenuNavigationController {
...
+                        // No keys found in local. Add it
+                        savedPassPhrase = try await saveKeyToLocal(
+                            context: context,
+                            keyDetail: keyDetail,
+                            passPhrase: savedPassPhrase,
+                            isNewKey: true
+                        )
+                        isKeyUpdated = true
+                        continue
+                    }
```
tomholub, line 132: More obvious logic, I think, would be with two loops. 

Pseudo code

```swift
for keyDetail in keyDetals {
  if ... {
    keysToUpdate.append(keyDetail)
  }
}
if keysToUpdate.empty() {
  return
}
if passPhrase == nil {
  passPhrase = // show pass phrase modal
}
for keyToSave in keysToSave {
  saveKeyToLocal(context, keyToSave, passPhrase)
}
```

I don't think you'll need `isNewKey` in the pass phrase prompt. In either case, pass phrase should be requested for the operation as a whole (updating all account keys), not for each individual key.

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
```diff
@@ -96,6 +103,167 @@ final class SideMenuNavigationController: ENSideMenuNavigationController {
     }
 }
 
+// MARK: Refresh keys from EKM
+extension SideMenuNavigationController {
+    private func refreshKeysFromEKMIfNeeded(context: AppContextWithUser) {
+        Task {
+            do {
```
tomholub, line 110: I'm unsure if this is the best place to put the logic - maybe @sosnovsky will have some thoughts

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
```diff
@@ -57,6 +63,7 @@ final class SideMenuNavigationController: ENSideMenuNavigationController {
             $0.delegate = self
             $0.animationDuration = Constants.animationDuration
         }
+        refreshKeysFromEKMIfNeeded(context: appContext)
```
tomholub, line 66: Will this run only once per app run? I imagine this could run each time the view is rendered which could be more than once per app startup?

FlowCrypt/Models/Common/Keypair.swift
```diff
@@ -16,6 +16,7 @@ struct Keypair {
     var source: String
     var allFingerprints: [String]
     var allLongids: [String]
+    var lastModified: Int?
```
tomholub, line 19: You've added a value but I didn't see any migration to ensure these values are filled for already existing keys

appium/api-mocks/apis/ekm/ekm-endpoints.ts
```diff
@@ -62,298 +62,422 @@ const throwErrorIfConfigSaysSo = (config: EkmConfig) => {
...
+// pub, primaryFingerprint, name, date are used to check Keys in KeyScreen and can be optional
+export interface KeyDetailInfo {
+  prv: string;
+  pub?: string;
+  primaryFingerprint?: string;
+  name?: string;
+  date?: string;
+}
+
+type KeyTypes = 'e2e' | 'flowcryptCompability' | 'key0' | 'key0Updated' | 'key1' | 'e2eRevokedKey' | 'e2eValidKey';
```
tomholub, line 74: Let's call this type `KeySampleName`

appium/api-mocks/apis/ekm/ekm-endpoints.ts
```diff
@@ -62,298 +62,422 @@ const throwErrorIfConfigSaysSo = (config: EkmConfig) => {
   }
 }
 
+// pub, primaryFingerprint, name, date are used to check Keys in KeyScreen and can be optional
+export interface KeyDetailInfo {
+  prv: string;
+  pub?: string;
+  primaryFingerprint?: string;
+  name?: string;
+  date?: string;
```
tomholub, line 71: What does the date represent here?

appium/api-mocks/apis/ekm/ekm-endpoints.ts
```diff
@@ -62,298 +62,422 @@ const throwErrorIfConfigSaysSo = (config: EkmConfig) => {
   }
 }
 
+// pub, primaryFingerprint, name, date are used to check Keys in KeyScreen and can be optional
+export interface KeyDetailInfo {
+  prv: string;
+  pub?: string;
+  primaryFingerprint?: string;
+  name?: string;
```
tomholub, line 70: this would be `primaryUid`

appium/tests/data/index.ts
```diff
@@ -213,7 +213,12 @@ export const CommonData = {
     email: 'revoked@domain.test',
     name: 'revoekd@flowcrypt.com'
   },
-  keyManagerURL : {
+  refreshKeys: {
+    wrongPassPhrase: 'Error\nIncorrect pass phrase. Please try again.',
+    updatedSuccessfully: 'Account keys updated',
+    errorMessage: 'Error\nCould not update keys from EKM due to error: '
```
tomholub, line 219: After some thought, let's change to `Error updating account keys: `

appium/tests/helpers/AppiumHelper.ts
```diff
@@ -0,0 +1,19 @@
+import { CommonData } from "../data";
+
+class AppiumHelper {
+    /**
+     * Restart app with processArguments
+     */
+    static async restartApp() {
+        const bundleId = CommonData.bundleId.id;
+        await driver.terminateApp(bundleId);
+        const processArgs = ['--mock-fes-api', '--mock-attester-api'];
```
tomholub, line 10: This restarts app as a mock which may be surprising to someone who may wish to restart the normal app.

Let's pass processArgs as an argument of `restartApp` and then also let's save somewhere `const mockAppProcessArgs = ['--mock-fes-api', '--mock-attester-api']` so that we don't have to track two places where these are defined.

So you would call `AppiumHelper.restartApp(mockAppProcessArgs)` or something like that.

appium/tests/screenobjects/keys.screen.ts
```diff
@@ -1,12 +1,13 @@
...
   ADD_BUTTON: '~Add',
-  NAME_AND_EMAIL: '-ios class chain:**/XCUIElementTypeOther/XCUIElementTypeStaticText[1]',
-  DATE_CREATED: '-ios class chain:**/XCUIElementTypeOther/XCUIElementTypeStaticText[2]',
-  FINGERPRINT: '-ios class chain:**/XCUIElementTypeOther/XCUIElementTypeStaticText[3]',
+  NAME_AND_EMAIL: '~keyTitle',
+  DATE_CREATED: '~keyDate',
+  FINGERPRINT: '~keySubtitle',
   SHOW_PUBLIC_KEY_BUTTON: '~Show public key',
   SHOW_KEY_DETAILS_BUTTON: '~Show key details',
   COPY_TO_CLIPBOARD_BUTTON: '~Copy to clipboard',
```
tomholub, line 13: These should all ideally use custom `aid-`. I'll file another issue for that for later.

tomholub (changes requested):
Thank you - my review below. I haven't scrutinized the tests that well yet, mostly focused on implementation.

appium/tests/helpers/AppiumHelper.ts
ioanmo226: ok, I agree

appium/tests/screenobjects/keys.screen.ts
tomholub: #1545

appium/api-mocks/apis/ekm/ekm-endpoints.ts
ioanmo226: This is to check if key screen displays keys correctly.
we can't get name, date from appium ui tests so I added those infos in key details.
let me know your thought.

image

appium/api-mocks/apis/ekm/ekm-endpoints.ts
ioanmo226: Please check above comment

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
ioanmo226: Yeah, I thought about this too and couldn't find better place

appium/api-mocks/apis/ekm/ekm-endpoints.ts
tomholub: Let's call them `renderedPrimaryUid`, and then check in the app what does the date represent. Is it date created or date last modified? Based on that, call it `renderedDateCreated` or `renderedDateLastModified`

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
ioanmo226: I added `guard` above because there might some cases which lastModified is null but want to add/update keys to local

appium/api-mocks/apis/ekm/ekm-endpoints.ts
ioanmo226: `renderedPrimaryUid` doesn't seem to correct naming as it contains name + email

ioanmo226: like `Test1 <test1@example.net>`

appium/tests/data/index.ts
ioanmo226: ok

appium/api-mocks/apis/ekm/ekm-endpoints.ts
tomholub: That is exactly how user id (uid) is encoded in the public/private key. It's correct :-)

ioanmo226: ok :-)

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
ioanmo226: `isNewKey` is passed because after user enter passphrase, we need to check if entered passphrase matches previous key detail if it's already existing key.
 
```
// For new key just check if there are any matching keys
if isNewKey {
    return matchingKeys.isNotEmpty
}
// now figure out if the pass phrase also matched the signing prv itself
let matched = matchingKeys.first(where: { $0.fingerprints.first == keyDetail.primaryFingerprint })
return matched != nil// true if the pass phrase matched signing key
```

FlowCrypt/Models/Common/Keypair.swift
ioanmo226: As it's nullable, we don't need new migration.
If this value is nil, update logic happens .

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
ioanmo226: Not once per app run.
Each time app runs + when user logs into new account.

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
```diff
@@ -114,38 +114,39 @@ extension SideMenuNavigationController {
...
-                            keyDetail: keyDetail,
-                            passPhrase: savedPassPhrase
-                        )
-                        isKeyUpdated = true
+
+                var passPhrase = localKeys.first(where: { $0.passphrase != nil })?.passphrase
+                // If this is called when starting the app, then it doesn't make much difference
+                // but conceptually it would be better to look pass phrase both in memory and storage
+                if passPhrase == nil {
+                    passPhrase = try context.passPhraseService.getPassPhrases().first(where: { $0.value.isNotEmpty })?.value
```
tomholub, line 124: My understanding was that `context.passPhraseService.getPassPhrases()` already looked in both storage and memory. Could you check? Then you'd only need that one.

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
```diff
@@ -114,38 +114,39 @@ extension SideMenuNavigationController {
...
+                }
+
+                // second parameter indicates whether key detail is new key or not
+                var keysToUpdate: [(KeyDetails, Bool)] = []
+                for keyDetail in keyDetails {
+                    if let savedLocalKey = localKeys.first(where: { $0.primaryFingerprint == keyDetail.primaryFingerprint }) {
+                        if savedLocalKey.lastModified ?? 0 < keyDetail.lastModified ?? 0 {
+                            keysToUpdate.append((keyDetail, false))
+                        }
```
tomholub, line 133: I saw your comment. If you would want to prevent a nil `lastModified` from being saved to storage (I think that's a good idea), then let's take this a step further: 
 - make `lastModified` in storage not optional. Do a migration and in the migration, remove keys that are missing `lastModified`
 - add a `guard` here to ensure last modified is present. If not present, instead of skipping the key, throw as if the key was invalid.

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
tomholub: I guess also when user switches accounts? (in left menu?)

FlowCrypt/Models/Common/Keypair.swift
tomholub: I'd rather that you do a migration where you parse all existing keys. See https://github.com/FlowCrypt/flowcrypt-ios/pull/1510#discussion_r860691290 then you don't need an optional, which is better.

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
tomholub: At a minimum, it could be a separate file? `SideMenuNavigationController+RefreshAccountKeys.swift` ?

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
tomholub: That's the thing - I don't think that's needed. I'll take a closer look.

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
ioanmo226: Sure. Let's see if @sosnovsky has better suggestion

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
tomholub: That's the thing, I don't think we need that here.

Instead of asking for pass phrase "for a particular key", ask for a pass phrase "for the whole update opration". Then you don't have to fiddle with the `isNewKey` and the whole logic cleans up.

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
ioanmo226: Yeah, correct

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
ioanmo226: Yeah, indeed it looks in both storage.
But  `encryptedStorage.getPassPhrases()` function doesn't search pass phrases by user email.
What I worry is it might return incorrect pass phrase (when user have 2 acccounts (`A`, `B`), it might return `A` private key for `B` account)

What do you think?
```
func getPassPhrases() throws -> [PassPhrase] {
    try encryptedStorage.getPassPhrases() + inMemoryStorage.getPassPhrases()
}
```

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
ioanmo226: I set Realm `lastModified` to be nullable because api response type is also nullable.
image

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
tomholub: Wow, seriously? There is no "current account" in the init or somewhere either? Or init of encrypted storage or inMemoryStorage?

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
tomholub: Sure. But keys that don't have information in them when they were last modified are useless keys (they lack pgp self-signatures).

Therefore, I propose (inspired by your original code) that the storage should be not optional, and that we ensure that only keys that have this are saved to storage.

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
ioanmo226: This is `encryptedStorage.getPassPhrases` function and I can't find user email check.

```
func getPassPhrases() throws -> [PassPhrase] {
    return try storage.objects(KeypairRealmObject.self)
        .compactMap(PassPhrase.init)
}
```

FlowCrypt/Models/Common/Keypair.swift
ioanmo226: ok

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
tomholub: Ok. That needs to be fixed for sure. Well noticed.

ioanmo226: then how about we fix in this PR?

ioanmo226: as i have to use getPassPhrases function in this PR as well?
Or do you wanna to address in new issue?

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
ioanmo226: ok

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
sosnovsky: For me it'll be better to have separate `EKMService` which will handle all refresh logic, so view controller will just call refresh method like `ekmService.refreshKeysIfNeeded(in: self)`

Also splitting big `refreshKeysFromEKMIfNeeded` method to smaller ones will make it more readable, like this:
```swift
let configuration = try await context.clientConfigurationService.configuration

guard configuration.checkUsesEKM() == .usesEKM else { return }

let emailKeyManagerApi = EmailKeyManagerApi(clientConfiguration: configuration)
let idToken = try await IdTokenUtils.getIdToken(userEmail: context.user.email)
let keyDetails = try await emailKeyManagerApi.getPrivateKeys(idToken: idToken)
let localKeys = try context.encryptedStorage.getKeypairs(by: context.user.email)

let passPhrase = try fetchCurrentPassphrase()
let keysToUpdate = findExpiredKeys(from: keyDetails)

try await update(keys: keysToUpdate)
showToast("refresh_key_success".localized)
```

ioanmo226: Nice suggestion

ioanmo226: will follow your way

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
ioanmo226: @tomholub

tomholub: sure, let's fix it here

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
ioanmo226: @tomholub one more question regarding `lastModified`.
For all beta users, key `lastModified` value is missing and we have to delete all keys for all users which means they will lose all infos related to key.
Is this fine?

> make lastModified in storage not optional. Do a migration and in the migration, remove keys that are missing lastModified

tomholub: No, you have to parse the existing keys during the db migration, and figure out what their `lastModified` is.

Only when you parse it and it returns a nil `lastModified` then it's useless key anyway and can be deleted.

ioanmo226: ok

ioanmo226: @sosnovsky Do you have any idea how to pass asynchronous function in `enumerateObjects`?
I haven't found solution yet.

image

ioanmo226: @sosnovsky Ohh, yes So simple solution. I thought it in too difficult way.

sosnovsky: Just need to test if `Task` works well for Realm migrations

ioanmo226: Yeah

ioanmo226: as we expected, it doesn't work well.
After it gets parsed result I can't set newObject detail because migration block is already finished and displays this error.
Finding solution now though.
`Object has been deleted or invalidated `

ioanmo226: As I checked deeply we can't accomplish this with default Realm migration schema.
I think only way to implement this is like this.

1. Set lastModified default value to 0.
2. When app starts check if isLastModifiedUpdated (setting value which might be saved in UserDefaults or something like that)is true and perform update `lastModified` logic before any other operations.
3. Set `isLastModifiedUpdated` to true.

@tomholub, @sosnovsky Please confirm

tomholub: I don't have any better idea. Sounds ok to me. Add a comment to remind us to remove it a few versions later.

ioanmo226: ok

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
```diff
@@ -96,6 +103,153 @@ final class SideMenuNavigationController: ENSideMenuNavigationController {
...
+
+                var passPhrase = localKeys.first(where: { $0.passphrase != nil })?.passphrase
+                // If this is called when starting the app, then it doesn't make much difference
+                // but conceptually it would be better to look pass phrase both in memory and storage
+                if passPhrase == nil {
+                    passPhrase = try context.passPhraseService.getPassPhrases().first(where: { $0.value.isNotEmpty })?.value
+                }
+
+                var keysToUpdate: [KeyDetails] = []
+                for keyDetail in keyDetails {
```
tomholub, line 128: please add 

```swift
guard keyDetail.isFullyDecrypted else { throw ekmKeyUnexpectedlyEncrypted }
// "Failed to update account keys: key received from EKM was unexpectedly encrypted"
```

ioanmo226: ok

FlowCrypt/Controllers/SideMenu/Main/SideMenuNavigationController.swift
```diff
@@ -57,6 +57,8 @@ final class SideMenuNavigationController: ENSideMenuNavigationController {
             $0.delegate = self
             $0.animationDuration = Constants.animationDuration
         }
+        let ekmService = EKMService(appContext: appContext)
```
tomholub, line 60: This looks good. I propose to call it `EkmVcHelper`. 

We are calling so many things a `service` in this codebase that the word has lost its meaning (to me, service is something that runs, not a collection of methods).

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -218,9 +221,12 @@ extension EncryptedStorage: PassPhraseStorageType {
         try updateKeys(with: passPhrase.primaryFingerprintOfAssociatedKey, passphrase: nil)
     }
 
-    func getPassPhrases() throws -> [PassPhrase] {
-        return try storage.objects(KeypairRealmObject.self)
-            .compactMap(PassPhrase.init)
+    func getPassPhrases(for email: String?) throws -> [PassPhrase] {
```
tomholub, line 224: Let's make `email` required. I don't see any reason to be pulling across accounts. Should always be for a particular account.

FlowCrypt/Functionality/Services/EKMService.swift
```diff
@@ -0,0 +1,188 @@
...
+        Task {
+            do {
+                self.askedUserPassPhrase = false
+                let configuration = try await appContext.clientConfigurationService.configuration
+                guard configuration.checkUsesEKM() == .usesEKM else {
+                    return
+                }
+                let emailKeyManagerApi = EmailKeyManagerApi(clientConfiguration: configuration)
+                let idToken = try await IdTokenUtils.getIdToken(userEmail: appContext.user.email)
+                let keyDetails = try await emailKeyManagerApi.getPrivateKeys(idToken: idToken)
```
tomholub, line 41: Let's call these `fetchedKeys`

FlowCrypt/Functionality/Services/EKMService.swift
```diff
@@ -0,0 +1,188 @@
...
+                    return
+                }
+
+                for keyDetail in keysToUpdate {
+                    passPhrase = try await saveKeyToLocal(
+                        in: viewController,
+                        context: appContext,
+                        keyDetail: keyDetail,
+                        passPhrase: passPhrase
+                    )
```
tomholub, line 57: You no longer need to do this `passphrase = ` reassignment in each loop. Instead, ask for the pass phrase above the loop if it's missing. That way the logic is more obvious without inspecting the methods.

```
guard passPhrase else {
  passPhrase = try await requestPassPhraseWithModal(in: viewController)
} 
```

Then you don't need `askedUserPassPhrase` either. Less state is generally better, and it's best if code reads like prose from top to bottom, as events are happening. It makes for more readable code.

FlowCrypt/Functionality/Services/EKMService.swift
```diff
@@ -0,0 +1,188 @@
...
+                }
+                await viewController.showToast("refresh_key_success".localized)
+            } catch {
+                if error is ApiError {
+                    return
+                }
+                // Do not display error alert when no keys are returned
+                if let ekmError = error as? EmailKeyManagerApiError, ekmError == .noKeys {
+                    return
+                }
```
tomholub, line 67: We should change this in the future in another PR. `emailKeyManagerApi.getPrivateKeys` should not throw when the list is empty, handling that should be the job of the code that called it. That would mean we could also remove this check.

FlowCrypt/Functionality/Services/EKMService.swift
```diff
@@ -0,0 +1,188 @@
...
+                // Do not display error alert when user cancels pass phrase prompt
+                if let refreshError = error as? RefreshKeyError, refreshError == .cancelPassPhrase {
+                    return
+                }
```
tomholub, line 71: This one can be removed as follows by improving the suggestion above to be (pseudo code)

```
guard passPhrase else {
  guard userEnteredPassPhrase = await requestPassPhraseWithModal(in: viewController) // returns nil on cancel
  if !userEnteredPassPhrase {
    return
  }
} 
```

User pressing cancel is not really exceptional, so if it doesn't make code too complicated, it's better to handle it directly.

Exception would be fine if handling it directly would be too complicated code wise.

FlowCrypt/Functionality/Services/EKMService.swift
```diff
@@ -0,0 +1,188 @@
...
+                        context: appContext,
+                        keyDetail: keyDetail,
+                        passPhrase: passPhrase
+                    )
+                }
+                await viewController.showToast("refresh_key_success".localized)
+            } catch {
+                if error is ApiError {
+                    return
+                }
```
tomholub, line 63: Should add a comment "since this is an update function that happens on every startup, it's ok if it's skipped sometimes - keys will be updated next time"

FlowCrypt/Functionality/Services/EKMService.swift
```diff
@@ -0,0 +1,188 @@
...
+        var newPassPhrase = passPhrase
+        if newPassPhrase == nil {
+            if askedUserPassPhrase {
+                return nil
+            }
+            newPassPhrase = try await requestPassPhraseWithModal(in: viewController)
+        }
+        guard let newPassPhrase = newPassPhrase else {
+            return nil
+        }
```
tomholub, line 116: This can be removed when we ask for pass phrase in `refreshKeysFromEKMIfNeeded` as mentioned above.

FlowCrypt/Functionality/Services/Local Private Key Services/InMemoryPassPhraseStorage.swift
```diff
@@ -38,7 +38,7 @@ final class InMemoryPassPhraseStorage: PassPhraseStorageType {
         passPhraseProvider.remove(passPhrases: [passPhrase])
     }
 
-    func getPassPhrases() -> [PassPhrase] {
+    func getPassPhrases(for _: String?) -> [PassPhrase] {
```
tomholub, line 41: Why is it not used?

FlowCrypt/Functionality/Services/EKMService.swift
tomholub: You can, but the scope is getting quite large in this PR

ioanmo226: ok, already changed though

FlowCrypt/Functionality/Services/Local Private Key Services/InMemoryPassPhraseStorage.swift
```diff
@@ -38,7 +38,7 @@ final class InMemoryPassPhraseStorage: PassPhraseStorageType {
         passPhraseProvider.remove(passPhrases: [passPhrase])
     }
 
-    func getPassPhrases() -> [PassPhrase] {
+    func getPassPhrases(for email: String) -> [PassPhrase] {
         passPhraseProvider.passPhrases
```
ioanmo226, line 42: For `InMemoryPassPhraseStorage`, email field is not respected.
This also applies to `logOutUser` too.
To check/delete passphrase from user email in `InMemoryPassPhraseStorage` we need to pass KeyService or AppContext to get key pairs from user email first(And then get matching pass phrases from there).
But this seems big changes and need to be addressed in another issue I think.
@tomholub Please confirm.
 
image

FlowCrypt/Functionality/Services/EKMService.swift
```diff
@@ -74,10 +70,14 @@ final class EKMService: EKMServiceType {
...
+    private func getPassphrase(in viewController: UIViewController) async throws -> String? {
         // If this is called when starting the app, then it doesn't make much difference
         // but conceptually it would be better to look pass phrase both in memory and storage
-        return try appContext.passPhraseService.getPassPhrases(for: appContext.user.email).first(where: { $0.value.isNotEmpty })?.value
+        var passphrase = try appContext.passPhraseService.getPassPhrases(for: appContext.user.email).first(where: { $0.value.isNotEmpty })?.value
+        if passphrase == nil {
+            passphrase = try await requestPassPhraseWithModal(in: viewController)
+        }
+        return passphrase
     }
```
tomholub, line 81: could be a bit simpler with 

```swift
if let passPhrase = try appContext.passPhraseService.getPassPhrases(for: appContext.user.email).first(where: { $0.value.isNotEmpty })?.value {
  return passPhrase
}
return try await requestPassPhraseWithModal(in: viewController)
```

`var` should generally be avoided in favor of `let`, everything else being equal.

FlowCrypt/Functionality/Services/Local Private Key Services/InMemoryPassPhraseStorage.swift
tomholub: That's a real mess. I'll file another issue for afterwards. Too much in this PR.

FlowCrypt/Functionality/Services/EKMService.swift
tomholub: unresolving - wasn't actually resolved in code

FlowCrypt/Functionality/Services/EKMService.swift
```diff
@@ -58,13 +56,11 @@ final class EKMService: EKMServiceType {
...
                 if error is ApiError {
                     return
                 }
-                // Do not display error alert when no keys are returned
-                if let ekmError = error as? EmailKeyManagerApiError, ekmError == .noKeys {
-                    return
-                }
                 // Do not display error alert when user cancels pass phrase prompt
                 if let refreshError = error as? RefreshKeyError, refreshError == .cancelPassPhrase {
                     return
```
tomholub, line 66: I've unresolved a previous comment, since this one is still here but can be removed

FlowCrypt/Functionality/Services/EKMService.swift
```diff
@@ -9,18 +9,17 @@
...
 import UIKit
 
-protocol EKMServiceType {
+protocol EKMVcHelperType {
     func refreshKeysFromEKMIfNeeded(in viewController: UIViewController)
 }
 
 enum RefreshKeyError: Error {
     case cancelPassPhrase
 }
```
tomholub, line 18: Won't be needed if we can just `return` on pass phrase cancel

FlowCrypt/Functionality/Services/EKMService.swift
```diff
@@ -103,24 +103,17 @@ final class EKMService: EKMServiceType {
...
-                return nil
-            }
-            newPassPhrase = try await requestPassPhraseWithModal(in: viewController)
-        }
-        guard let newPassPhrase = newPassPhrase else {
-            return nil
+    ) async throws {
+        guard let passPhrase = passPhrase else {
+            return
         }
```
tomholub, line 109: Can be removed if `passPhrase: String?` is changed to `passPhrase: String`

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -218,9 +221,12 @@ extension EncryptedStorage: PassPhraseStorageType {
...
 
-    func getPassPhrases() throws -> [PassPhrase] {
-        return try storage.objects(KeypairRealmObject.self)
-            .compactMap(PassPhrase.init)
+    func getPassPhrases(for email: String) throws -> [PassPhrase] {
+        var keypairObjects = try storage.objects(KeypairRealmObject.self)
+        if email.isNotEmpty {
+            keypairObjects = keypairObjects.where({ $0.user.email == email })
+        }
+        return keypairObjects.compactMap(PassPhrase.init)
```
tomholub, line 229: No point ever returning pass phrases of all accounts.

Always filter by provided email.

FlowCrypt/Functionality/Services/Local Private Key Services/InMemoryPassPhraseStorage.swift
tomholub: filed https://github.com/FlowCrypt/flowcrypt-ios/issues/1551

FlowCrypt/Functionality/Services/EKMService.swift
ioanmo226: Yeah, I'm currently taking care of it.

ioanmo226: Yeah, I agree.
Addressed all review comments though.

ioanmo226: And ready to review

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -221,9 +221,9 @@ extension EncryptedStorage: PassPhraseStorageType {
...
     }
 
-    func getPassPhrases(for email: String?) throws -> [PassPhrase] {
+    func getPassPhrases(for email: String) throws -> [PassPhrase] {
         var keypairObjects = try storage.objects(KeypairRealmObject.self)
-        if let email = email, email.isNotEmpty {
+        if email.isNotEmpty {
             keypairObjects = keypairObjects.where({ $0.user.email == email })
         }
         return keypairObjects.compactMap(PassPhrase.init)
```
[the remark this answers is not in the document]
tomholub: The unit test can get pass phrases specifically of accounts that it's interested in. Is that feasible?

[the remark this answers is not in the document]
ioanmo226: Yeah, it already does. Let me remove this check.

tomholub:
Let me know once ready for review

FlowCrypt/Functionality/Services/EKMService.swift
tomholub: this wasn't addressed, unresolving

appium/tests/specs/mock/setup/CheckRefreshKeyFromEKM.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      await KeysScreen.checkKeysScreen([ekmPrivateKeySamples.key0Updated, ekmPrivateKeySamples.key1]);
+
+      // stage 5 - older version of key does not get updated
+      mockApi.ekmConfig = {
+        returnKeys: [ekmPrivateKeySamples.key0.prv]
+      }
+      await AppiumHelper.restartApp(processArgs);
+      await goToKeysScreen();
+      await KeysScreen.checkKeysScreen([ekmPrivateKeySamples.key0Updated, ekmPrivateKeySamples.key1]);
+      await logoutUserFromKeyScreen();
```
tomholub, line 99: `logoutUserFromKeyScreen` is probably needed because there are three tests in one spec. So the state persists, I suppose?

It would probably be appropriate to split these three into three separate files (as other tests are) and then you don't have to log out in between. More importantly, if one test fails, it won't fail all three of these.

tomholub (commented):
A few more comments. The tests are looking good, and the implementation is much better now.

FlowCrypt/Functionality/Services/EKMVcHelper.swift
```diff
@@ -0,0 +1,169 @@
...
+        _ passPhrase: String
+    ) async throws -> Bool {
+        // since pass phrase was entered (an inconvenient thing for user to do),
+        //  let's find all keys that match and save the pass phrase for all
+        let allKeys = try await appContext.keyService.getPrvKeyInfo(email: appContext.user.email)
+        guard allKeys.isNotEmpty else {
+            // tom - todo - nonsensical error type choice https://github.com/FlowCrypt/flowcrypt-ios/issues/859
+            //   I copied it from another usage, but has to be changed
+            throw KeyServiceError.retrieve
+        }
```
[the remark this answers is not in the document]
ioanmo226: ok

FlowCrypt/Functionality/Services/Local Private Key Services/PassPhraseService.swift
```diff
@@ -89,7 +89,7 @@ final class PassPhraseService: PassPhraseServiceType {
         case .persistent:
             try encryptedStorage.save(passPhrase: passPhrase)
         case .memory:
-            let storedPassPhrases = try encryptedStorage.getPassPhrases()
+            let storedPassPhrases = try encryptedStorage.getPassPhrases(for: "")
```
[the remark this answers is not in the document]
ioanmo226: This should be addressed in #1551 I think. This requires `savePassPhrase` refactor.

appium/tests/specs/mock/setup/CheckRefreshKeyFromEKM.spec.ts
ioanmo226: Ok, got it

FlowCrypt/Functionality/Services/Local Private Key Services/PassPhraseService.swift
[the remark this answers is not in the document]
tomholub: I'll add the comment there. It really is messy.

appium/config/wdio.mock.conf.js
```diff
@@ -7,6 +7,9 @@ config.suites = {
   all: [
     './tests/specs/mock/**/*.spec.ts'
   ],
+  test: [
+    './tests/specs/mock/setup/CheckRefreshKeyFromEKM.spec.ts'
+  ],
```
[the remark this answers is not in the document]
ioanmo226: Yeah, this code needs to be removed.

FlowCrypt/Functionality/Services/EKMVcHelper.swift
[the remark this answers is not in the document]
tomholub: https://github.com/FlowCrypt/flowcrypt-ios/issues/1556

appium/tests/screenobjects/keys.screen.ts
```diff
@@ -64,20 +65,25 @@ class KeysScreen extends BaseScreen {
     return $(SELECTORS.BACK_BUTTON)
   }
 
-  checkKeysScreen = async () => {
-    //will add value verification for key later, need to create Api request for get key value
+  checkKeysScreen = async (keys?: KeyDetailInfo[]) => {
```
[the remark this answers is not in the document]
tomholub: I'd rather they are always passed explicitly, even if that means these usages need to be updated. It will make it more obvious exactly what is being tested.

appium/tests/specs/mock/setup/CheckRefreshKeyFromEKMFirst.spec.ts
```diff
@@ -12,35 +12,26 @@ import RefreshKeyScreen from "../../../screenobjects/refresh-key.screen";
...
-const logoutUserFromKeyScreen = async () => {
-  await KeysScreen.clickBackButton();
-  await MenuBarScreen.clickMenuIcon();
-  await MenuBarScreen.clickLogout();
-  await SplashScreen.checkLoginPage();
-}
-
 describe('SETUP: ', () => {
 
   it('app auto updates keys from EKM during startup with a pass phrase prompt', async () => {
```
tomholub, line 24: Let's rename the file too - `CheckRefreshKeyFromEkmWithPassPhrasePrompt.spec.ts`

appium/tests/specs/mock/setup/CheckRefreshKeyFromEKMSecond.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const goToKeysScreen = async () => {
+  await MenuBarScreen.clickMenuIcon();
+  await MenuBarScreen.clickSettingsButton();
+  await SettingsScreen.clickOnSettingItem('Keys');
+}
+
+describe('SETUP: ', () => {
+
+  it('app auto updates keys from EKM during startup without pass phrase prompt', async () => {
```
tomholub, line 22: Let's rename the file too - `CheckRefreshKeyFromEkmWithoutPassPhrasePrompt.spec.ts`

appium/tests/specs/mock/setup/CheckRefreshKeyFromEKMThird.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const goToKeysScreen = async () => {
+  await MenuBarScreen.clickMenuIcon();
+  await MenuBarScreen.clickSettingsButton();
+  await SettingsScreen.clickOnSettingItem('Keys');
+}
+
+describe('SETUP: ', () => {
+
+  it('EKM key update errors handled gracefully', async () => {
```
tomholub, line 23: Let's rename the file too - `CheckRefreshKeyFromEkmErrorHandling.spec.ts`

appium/tests/specs/mock/setup/CheckRefreshKeyFromEKMSecond.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+import { ekmPrivateKeySamples } from "../../../../api-mocks/apis/ekm/ekm-endpoints";
+import { CommonData } from "../../../data";
+import BaseScreen from "../../../screenobjects/base.screen";
+import AppiumHelper from "../../../helpers/AppiumHelper";
+
+const goToKeysScreen = async () => {
+  await MenuBarScreen.clickMenuIcon();
+  await MenuBarScreen.clickSettingsButton();
+  await SettingsScreen.clickOnSettingItem('Keys');
+}
```
tomholub, line 18: Since this is written three times, you could add it as a method to `KeysScreen`. Or alternatively to `InboxScreen` if there is one.

tomholub (commented):
Getting closer :-)

appium/tests/screenobjects/keys.screen.ts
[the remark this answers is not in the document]
ioanmo226: ok

FlowCrypt/Functionality/Services/EKMVcHelper.swift
```diff
@@ -96,7 +100,8 @@ final class EKMVcHelper: EKMVcHelperType {
         in viewController: UIViewController,
         context: AppContextWithUser,
         keyDetail: KeyDetails,
-        passPhrase: String
+        passPhrase: String,
+        storageMethod: StorageMethod
```
tomholub, line 104: let's call it `passPhraseStorageMethod` else it's confusing - it may seem like it's about the key, but it's about the pass phrase

FlowCrypt/Functionality/Services/EKMVcHelper.swift
```diff
@@ -30,6 +30,7 @@ final class EKMVcHelper: EKMVcHelperType {
                 guard configuration.checkUsesEKM() == .usesEKM else {
                     return
                 }
+                let storageMethod: StorageMethod = configuration.forbidStoringPassPhrase ? .memory : .persistent
```
tomholub, line 33: Also here `passPhraseStorageMethod`

tomholub (commented):
Looks good. Following these two comments, I'm satisfied code-wise.

FlowCrypt/Functionality/Services/EKMVcHelper.swift
ioanmo226: OK

tomholub (approved):

ioanmo226:
I haven't fixed semaphoreci tests fail issue(which succeed in local) yet though.

tomholub:
Sure - I'll reapprove once that is fixed.

ioanmo226:
This mock UI test failure seems real strange issue.
Mock UI tests should not fail. (And it succeed well in local even with fresh clone).
Digging more now.

ioanmo226:
@tomholub, @sosnovsky I still couldn't reproduce sempaphoreci test failure issue in my local.
Could you try in your local to see if `CheckRefreshKeyFromEkmErrorHandling` and `CheckRefreshKeyFromEkmWithPassPhrasePrompt` test fails in your local?

tomholub:
You can try to re-enable the video reporter. Then see video of the fail

tomholub:
https://github.com/FlowCrypt/flowcrypt-ios/issues/1386

ioanmo226:
ok

ioanmo226:
Where are videos supposed to be saved?
`outputDir: './video',`
I guess it should be saved in root `video` folder.
I enabled video reporter and looked up video folder in [Artifacts](https://flowcrypt.semaphoreci.com/jobs/975edb3f-0d94-40e2-9f4e-2b73b97996fa/artifacts) but couldn't find it.

sosnovsky:
There are some issues with video reporter, as it fails to start, so no videos are recorded.
That's why I disabled it before https://github.com/FlowCrypt/flowcrypt-ios/issues/1386 is fixed

Screenshot 2022-05-02 at 23 04 52

ioanmo226:
@sosnovsky Could you have a look at this `CheckRefreshKeyFromEkmWithPassPhrasePrompt` test failure issue in semaphoreci?

From the semaphoreci screenshot, I thought it failed because when empty folder list is returned in `InboxViewContainerController`.
So I displayed alert when no folders are returned. (https://github.com/FlowCrypt/flowcrypt-ios/pull/1510/commits/e766c02a28d529300e1fcbaeec22ff6a97778d85)
But sempaphoreci still displays below image when running above test. (FYI this test always succeed in my local.)
It seems like there's a problem with in `InboxViewContainerController` when EKM returns error in sempahoreci, but couldn't find why.

image

sosnovsky:
> @sosnovsky Could you have a look at this `CheckRefreshKeyFromEkmWithPassPhrasePrompt` test failure issue in semaphoreci?

@ioanmo226 were you able to reproduce it locally?

ioanmo226:
No, Btw as you can see, I added this logic and i think this might help(https://github.com/FlowCrypt/flowcrypt-ios/pull/1510/commits/5b1150e308137d8bc435a2212941bf41156af62c). waiting for semaphoreci test to finish.

sosnovsky:
> No, Btw as you can see, I added this logic and i think this might help([5b1150e](https://github.com/FlowCrypt/flowcrypt-ios/commit/5b1150e308137d8bc435a2212941bf41156af62c)). waiting for semaphoreci test to finish.

So test fails because messages list is still loading when app tries to refresh EKM keys? 
But what if it'll happen in prod too - user's messages list will be loading while app tries to refresh EKM keys, will it fail?

ioanmo226:
> So test fails because messages list is still loading when app tries to refresh EKM keys?
I also am not 100% certain why it happens as it doesn't happen on my local at all. But my best guess is that InboxViewContainerController doesn't navigate to InboxViewController while semaphoreci EKM error alert  is displayed in semaphoreci.
> But what if it'll happen in prod too - user's messages list will be loading while app tries to refresh EKM keys, will it fail?
I don't think so.

sosnovsky:
Got it, let's wait for semaphore results then to see if this update fixed it.

ioanmo226:
I think this fixed issue and mock tests succeed.

FlowCrypt/Functionality/Services/EKMVcHelper.swift
```diff
@@ -0,0 +1,184 @@
...
+
+    init(appContext: AppContextWithUser) {
+        self.appContext = appContext
+        self.keyMethods = KeyMethods()
+    }
+
+    func refreshKeysFromEKMIfNeeded(in viewController: UIViewController) {
+        Task {
+            do {
+                // Sleep for 3 seconds when mock testing
```
ioanmo226, line 29: @sosnovsky, @tomholub So Tests should pass now.
@tomholub FYI only changes I added after your approval is this logic.
Let me know what you think.

```
if Bundle.isDebugBundleWithArgument("--mock-fes-api") {
  try await Task.sleep(nanoseconds: 3 * 1000 * 1_000_000)
}
```

FlowCrypt/Functionality/Services/EKMVcHelper.swift
```diff
@@ -0,0 +1,184 @@
...
+                // (This is to prevent refresh key UI test failure in semaphoreCI)
+                if Bundle.isDebugBundleWithArgument("--mock-fes-api") {
+                    try await Task.sleep(nanoseconds: 3 * 1000 * 1_000_000)
+                }
```
ioanmo226, line 33: @sosnovsky, @tomholub So Tests should pass now.
@tomholub FYI only changes I added after your approval is this logic (https://github.com/FlowCrypt/flowcrypt-ios/commit/5b1150e308137d8bc435a2212941bf41156af62c).
Let me know what you think.

```
if Bundle.isDebugBundleWithArgument("--mock-fes-api") {
  try await Task.sleep(nanoseconds: 3 * 1000 * 1_000_000)
}
```

tomholub: Quite horrible 👀 but I'm ok to merge it so we can progress.

There must be a more reasonable way to fix the issue. It could be a change to test code or production code (maybe the flow would have to be changed to be easier to test, idk).

tomholub (approved):

PR #1511, opened 2022-04-20
#1500 Add GMP library

### LINKED ISSUE #1500: add GMP library to the project - priority
@sosnovsky please pause your other work to look at this one

Roughly this:
1) please add a script that builds GMP into `Script/build-gmp.sh` that downloads https://gmplib.org/download/gmp/gmp-6.2.1.tar.lz and builds it for iPhone arm64 . Even better if this was somehow done in `make dependencies`
2) do a reverse of https://github.com/FlowCrypt/flowcrypt-ios/pull/237/files which means adding  back gmplib to the project and removing Swift `BigInt`. Instead of Lib/Bignum please call it `Lib/GMP`
3) you'll want to copy `bignum.c` and `bignum.h` that was removed in that PR but call it `gmp-interface.c` and `gmp-interface.h` or similar.

Then update this method back to use the method from `gmp-interface` as you see it below:

image

Performance is the reason. It takes 30-90 seconds to do RSA 4096 sign in JavaScript. Swift BigInt is only marginally faster, because it's not optimized. This is causing issues to our customer who uses RSA 4096. GMP should have it near instant.

GMP was removed at the time because it was causing issues for our build, which may no longer be the case, or is likely something you'll know how to resolve.

Thanks!

sosnovsky:
This PR replaces `BigNum` with `GMP` library for better performance

close #1500

### To be filled by reviewers


sosnovsky:
@tomholub I added script for building `GMP` from source, but it takes 10-12 minutes:

Group 1

What about creating Swift Package from `GMP.xcframework` (like we did with `mailcore` https://github.com/FlowCrypt/mailcore2)? It won't need pre-building and will be fetched along with other packages.

Then I'll just need new empty repository https://github.com/FlowCrypt/GMP

tomholub:
sounds good, here: https://github.com/FlowCrypt/GMP/invitations

tomholub (approved):
Looks good so far. Will wait for the Swift Package

Makefile
```diff
@@ -5,7 +5,6 @@ dependencies:
 	gem install bundler:2.3.11
 	bundle config set path 'vendor/bundle'
 	bundle install
-	sh Scripts/build-gmp.sh
```
tomholub: Please add this script to FlowCrypt/GMP instead - can maintain it over there.

tomholub (approved):

FlowCrypt/Lib/GMP/gmp-interface.m
```diff
@@ -0,0 +1,27 @@
+#include "gmp.h"
+#include <stdio.h>
+
+const char* c_gmp_mod_pow(const char* base, const char* exponent, const char* modulo) {
+#ifdef __arm64__
```
tomholub, line 5: @sosnovsky you mentioned making this work for other platforms (simulator, presumably amd64). Should this be updated to also allow this method for whatever other platforms that you built for?

sosnovsky: Yes, I'll remove this check, as now GMP library should work on all platforms - devices and simulators.

tomholub (approved):

PR #1512, opened 2022-04-21
Remember pass phrase when DEFAULT_REMEMBER_PASS_PHRASE is passed from FES

ioanmo226:
This PR remembers pass phrase when DEFAULT_REMEMBER_PASS_PHRASE is passed from FES

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
+++ b/FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
@@ -19,13 +19,15 @@ import FlowCryptUI
 final class SetupGenerateKeyViewController: SetupCreatePassphraseAbstractViewController {
 
     private let attester: AttesterApiType
+    let clientConfiguration: ClientConfiguration
 
     init(
         appContext: AppContextWithUser,
         decorator: SetupViewDecorator = SetupViewDecorator()
     ) async throws {
+        self.clientConfiguration = try await appContext.clientConfigurationService.configuration
         self.attester = AttesterApi(
-            clientConfiguration: try await appContext.clientConfigurationService.configuration
+            clientConfiguration: clientConfiguration
         )
         super.init(
             appContext: appContext,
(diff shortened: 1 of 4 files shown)


ioanmo226:
@tomholub @sosnovsky I have one question.
What's the difference between saving passphrase with `passPhraseService.savePassPhrase` and `encryptedStorage.putKeypairs` (by passing passphrase parameter)?
Do we have to save passphrase into passPhraseService too when DEFAULT_REMEMBER_PASS_PHRASE is passed?

Attached code below.
```
// 1st case
let passPhrase = PassPhrase(
    value: passPhrase,
    fingerprintsOfAssociatedKey: encryptedPrv.key.fingerprints
)
try appContext.passPhraseService.savePassPhrase(with: passPhrase, storageMethod: .memory)
```

```
// 2nd case
try appContext.encryptedStorage.putKeypairs(
    keyDetails: [encryptedPrv.key],
    passPhrase: storageMethod == .persistent || clientConfiguration.shouldRememberPassphraseByDefault ? passPhrase: nil,
    source: .generated,
    for: appContext.user.email
)
```

tomholub:
In 1st case, you are using `storageMethod: .memory`. That just stores it in RAM.

In general, adding pass phrase to `putKeypairs`, or omitting it with subsequent call to `savePassPhrase` with `.persistent` is the same, except 1 db call (insert) vs 2 db calls (insert + update).

ioanmo226:
Got it. I think this PR doesn't need new changes then?

FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
```diff
@@ -89,7 +91,7 @@ final class SetupGenerateKeyViewController: SetupCreatePassphraseAbstractViewCon
     private func putKeypairsInEncryptedStorage(encryptedPrv: CoreRes.GenerateKey, storageMethod: StorageMethod, passPhrase: String) throws {
         try appContext.encryptedStorage.putKeypairs(
             keyDetails: [encryptedPrv.key],
-            passPhrase: storageMethod == .persistent ? passPhrase: nil,
+            passPhrase: storageMethod == .persistent || clientConfiguration.shouldRememberPassphraseByDefault ? passPhrase: nil,
```
tomholub, line 94: Same here. Plus, in situations when user actually has an option to choose (like when creating a key, there is radio select about method of storing pass phrase) you should not give user options that will then be ignored.

Also, I have misled you somewhat.

See https://flowcrypt.com/docs/business/org-rules.html

`DEFAULT_REMEMBER_PASS_PHRASE` means `Changes a setup checkbox default, in a way that browser will remember user pass phrase` which I'll rephrase more precisely as `Sets persistently storing pass phrase as the default choice in setup UI.`

Given there is no user UI involved regarding this in the EKM setup flow, `DEFAULT_REMEMBER_PASS_PHRASE` is not the right choice.

Simply removing flag `FORBID_STORING_PASS_PHRASE` should achieve the desired result (for EKM flow, at least). And if it doesn't, that's the fix that should be changed.

I'll close this PR and file a more appropriate issue.

tomholub (changes requested):

PR #1514, opened 2022-04-21
#1513 enable EKM setup without FORBID_STORING_PASS_PHRASE

### LINKED ISSUE #1513: check / update functionality of FORBID_STORING_PASS_PHRASE during EKM setup flow
Simply removing flag `FORBID_STORING_PASS_PHRASE` should achieve the desired result (for EKM flow, at least). And if it doesn't, that's the fix that should be changed.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1512#discussion_r854904299_

ioanmo226:
This PR saves pass phrase in memory when FORBID_STORING_PASS_PHRASE is set in EKM Setup

close #1513 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

- [ ] is accompanied with tests, or tests are not needed (will be tested in https://github.com/FlowCrypt/flowcrypt-ios/pull/1510 )

### DIFF
--- a/FlowCrypt/Controllers/Setup/SetupEKMKeyViewController.swift
+++ b/FlowCrypt/Controllers/Setup/SetupEKMKeyViewController.swift
@@ -72,7 +72,7 @@ extension SetupEKMKeyViewController {
     private func setupAccountWithKeysFetchedFromEkm(with passPhrase: String) async throws {
         self.showSpinner()
         let clientConfiguration = try await appContext.clientConfigurationService.configuration
-        storageMethod = clientConfiguration.forbidStoringPassPhrase ? .memory : .persistent
+        self.storageMethod = clientConfiguration.forbidStoringPassPhrase ? .memory : .persistent
         try await self.validateAndConfirmNewPassPhraseOrReject(passPhrase: passPhrase)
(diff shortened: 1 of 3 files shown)


FlowCrypt/Controllers/Setup/SetupEKMKeyViewController.swift
```diff
@@ -72,6 +71,7 @@ extension SetupEKMKeyViewController {
 
     private func setupAccountWithKeysFetchedFromEkm(with passPhrase: String) async throws {
         self.showSpinner()
+        let clientConfiguration = try await appContext.clientConfigurationService.configuration
```
tomholub, line 74: I would rather create `storageMethod` variable here, using the client configuration. 

That way the code below will be almost unchanged, and also clearer.

tomholub (changes requested):

FlowCrypt/Controllers/Setup/SetupEKMKeyViewController.swift
ioanmo226: Sounds good.
Just changed it

tomholub (approved):

PR #1523, opened 2022-04-22
#1522 Fix “Reset storage” button

sosnovsky:
This PR fixes storage reset on `InvalidStorageViewController`

close #1522

----------------------------------

**Tests** _(delete all except exactly one)_:
- Difficult to test (explain why) - difficult to reproduce "invalid storage" error

--------------------------------

### To be filled by reviewers

I have reviewed that this PR... _(tick whichever items you personally focused on during this review)_:
- [x] addresses the issue it closes (if any)
- [x] code is readable and understandable
- [x] is accompanied with tests, or tests are not needed
- [x] is free of vulnerabilities
- [x] is documented clearly and usefully, or doesn't need documentation

PR #1525, opened 2022-04-26
#1524 Handle parse key from EKM gracefully

### LINKED ISSUE #1524: ensure to properly parse each received key from EKM
> For `test 3: EKM key update errors handled gracefully`, when we want to parse broken key with `core.parseKeys` it doesn't return error but return 0 parsed keys.
> 
> > change EKM mock to start returning a broken key (for example, only return the first half of the armored key text)
> > restart the app
> > _observe that the app shows a modal during startup Could not update keys from EKM due to error: and followed by whatever error happened._

That's odd. In this situation, the parsing should fail. It should not swallow keys silently.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/841#issuecomment-1109067112_

ioanmo226:
This PR ensured to properly parse each received key from EKM.

close #1524 // if this PR closes an issue

- Tests will be added later (issue #841)

### To be filled by reviewers


ioanmo226:
@sosnovsky Please review this. This is blocking #841.
By the way, I can see that UI tests success in local but fail in semaphoreci. Do you have idea why?

FlowCrypt/Controllers/Setup/SetupInitialViewController.swift
```diff
@@ -158,30 +158,28 @@ extension SetupInitialViewController {
...
+                            }
+                        )
+                    } else if ekmError == .keysAreNotDecrypted {
+                        showAlert(
+                            message: ekmError.errorMessage,
+                            onOk: { [weak self] in
+                                self?.signOut()
+                            }
+                        )
+                    }
```
sosnovsky, line 182: It'll be better to use `switch` here, as currently we don't handle 2 `EmailKeyManagerApiError` cases.

sosnovsky:
> By the way, I can see that UI tests success in local but fail in semaphoreci. Do you have idea why?

Some tests are failing locally for me too, trying to fix them in my other PR - https://github.com/FlowCrypt/flowcrypt-ios/pull/1523

Reviewing this PR now 👀

ioanmo226:
> > By the way, I can see that UI tests success in local but fail in semaphoreci. Do you have idea why?
> 
> Some tests are failing locally for me too, trying to fix them in my other PR - #1523
> 
> Reviewing this PR now 👀

ok, got it

FlowCrypt/Functionality/Services/Remote Private Key Services/EmailKeyManagerApi.swift
```diff
@@ -68,22 +68,25 @@ actor EmailKeyManagerApi: EmailKeyManagerApiType {
...
+        var keys: [KeyDetails] = []
+        for privateKey in decryptedPrivateKeysResponse.privateKeys {
+            let parsedPrivateKey = try await core.parseKeys(armoredOrBinary: privateKey.decryptedPrivateKey.data())
+            // todo - check that parsedPrivateKeys don't contain public keys
+            let isKeyDecrypted = parsedPrivateKey.keyDetails
+                .compactMap { $0.isFullyDecrypted }
+            if isKeyDecrypted.contains(false) {
+                throw EmailKeyManagerApiError.keysAreNotDecrypted
+            }
+            if parsedPrivateKey.keyDetails.count < 1 {
```
sosnovsky, line 83: Instead of `count < 1` it'll be better to use `parsedPrivateKey.keyDetails.isEmpty`

sosnovsky (changes requested):

sosnovsky (dismissed):

FlowCrypt/Functionality/Services/Remote Private Key Services/EmailKeyManagerApi.swift
```diff
@@ -68,22 +68,25 @@ actor EmailKeyManagerApi: EmailKeyManagerApiType {
...
+            let parsedPrivateKey = try await core.parseKeys(armoredOrBinary: privateKey.decryptedPrivateKey.data())
+            // todo - check that parsedPrivateKeys don't contain public keys
+            let isKeyDecrypted = parsedPrivateKey.keyDetails
+                .compactMap { $0.isFullyDecrypted }
+            if isKeyDecrypted.contains(false) {
+                throw EmailKeyManagerApiError.keysAreNotDecrypted
+            }
+            if parsedPrivateKey.keyDetails.isEmpty {
+                throw EmailKeyManagerApiError.keysAreInvalid
+            }
```
tomholub, line 85: Here, we always expect exactly one key in each `decryptedPrivateKey`. Therefore `count !== 1` would be more appropriate.

tomholub (approved):

ioanmo226:
FYI no new changes. Just merged Roma's changes to save UI test run time.

FlowCrypt/Controllers/Setup/SetupInitialViewController.swift
```diff
@@ -158,30 +158,32 @@ extension SetupInitialViewController {
...
+                    case .keysAreNotDecrypted, .keysAreInvalid:
+                        showAlert(
+                            message: errorMessage,
+                            onOk: { [weak self] in
+                                self?.signOut()
+                            }
+                        )
+                    case .noPrivateKeysUrlString:
+                        break
+                    }
```
tomholub, line 186: All of the situations in this catch could be handled exactly the same way, with an `ok` and `retry` button as follows:

```swift
                        showRetryAlert(
                            message: errorMessage,
                            onRetry: { [weak self] in
                                self?.state = .fetchingKeysFromEKM
                            },
                            onOk: { [weak self] in
                                self?.signOut()
                            }
                        )
```

That means we probably wouldn't need `if let ekmError = error as? EmailKeyManagerApiError {` and we also wouldn't need the switch. It can all be handled the same.

tomholub: I'm unsure why this handler evolved to have three ways to handle the error, I think one way is enough, as long as it prints the appropriate error message.

tomholub: Yeah, that's how it was written, but there wasn't any reason for it that I know.

ioanmo226: Got it

tomholub (approved):

tomholub:
The most obvious examples are passing tests on PR that fail once merged to master. Those situations should be documented as issues so that someone will look into improving them, else eventually we'll be stuck re-running tests forever.

ioanmo226:
@tomholub Right.
We should improve CheckInboxPaginationAfterLoadingAllEmails test first. It fails 100% in my local as it takes 1-2 seconds to load list.

PR #1544, opened 2022-04-28
issue #1532 refactor + choose usable keys for outgoing messages

### LINKED ISSUE #1532: choose usable signing key for outgoing messages
I'll create a separate issue for the actual fix: the iOS app should carefully choose the signing key.

If private keys contain at least one key that is usable for signing, then use that. Else show error `None of your account keys are usable for signing`

Something like on the browser:

```ts
  private static chooseSigningKeyAndDecryptIt = async (
    view: ComposeView,
    senderKis: KeyInfoWithIdentity[]
  ): Promise<ParsedKeyInfo | undefined> => {
    const parsedSenderPrvs = await KeyStoreUtil.parse(senderKis);
    // to consider - currently we choose first valid key for signing. Should we sign with all?
    //   alternatively we could use most recenlty modified valid key
    const parsedSenderPrv = parsedSenderPrvs.find(k => k.key.usableForSigning);
    if (!parsedSenderPrv) {
      throw Error("None of your account keys are usable for signing")
    }
    return await view.storageModule.decryptSenderKey(parsedSenderPrv);
  };
}
```

Then this test has to be updated:

```ts
      await NewMessageScreen.clickSendButton();
      //TODO need to fix, the error should not be displayed for this case
      await BaseScreen.checkModalMessage(noPrivateKeyError);
```

and also this test will change:

```ts
    const noPrivateKeyError = 'Error\n' +
      'Could not compose message\n' +
      '\n' +
      'Error: Error encrypting message: Could not find valid key packet for signing in key bf79556b2cad5c1e';
```

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1526#discussion_r859905756_

ioanmo226:
This PR choosed usable signing key for outgoing messages

close #1532 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


appium/tests/specs/mock/setup/SetupOnlyRevokedKeyFromEKM.spec.ts
```diff
@@ -21,10 +21,10 @@ describe('SETUP: ', () => {
     const emailSubject = CommonData.simpleEmail.subject;
     const emailText = CommonData.simpleEmail.message;
 
+    // When private key is revoked key, there are no public keys to pick. So missing sender public key error occurs.
     const noPrivateKeyError = 'Error\n' +
-      'Could not compose message\n' +
-      '\n' +
-      'Error: Error encrypting message: Could not find valid key packet for signing in key bf79556b2cad5c1e';
+      'Could not compose message\n\n' +
+      'Missing sender public key. Is FlowCrypt iOS app well set up?';
```
tomholub, line 27: This is not a very good error.

There are actually two separate concerns when sending a signed+encrypted message, as it relates to my own keys:
 - do I have any **private** key that is usable for **signing**? -> error `Your account keys are not usable for signing`
 - do I have any **public** key that is usable for **encryption**? -> error `Your account keys are not usable for encryption`

FlowCrypt/Functionality/Services/Local Private Key Services/KeyService.swift
```diff
@@ -71,12 +76,14 @@ final class KeyService: KeyServiceType {
             }
```
tomholub, line 76: The `// todo - should be refactored with https://github.com/FlowCrypt/flowcrypt-ios/issues/812` above is outdated - please help me remove

FlowCrypt/Core/Models/KeyDetails.swift
```diff
@@ -40,6 +40,15 @@ extension KeyDetails {
...
+    
+    var isKeyUsuable: Bool {
+        var expired = false
+        let now = Date()
+        if let expiration = expiration, TimeInterval(expiration) < now.timeIntervalSince1970 {
+            expired = true
+        }
+        return !revoked && !expired
+    }
 }
```
tomholub, line 52: This is not ideal, but I see our current js response is missing this. We have to improve this after #1362 

Since this is only a rough approximation

tomholub (commented):
early review - I have to take another look

FlowCrypt/Core/Models/KeyDetails.swift
tomholub: https://github.com/FlowCrypt/flowcrypt-ios/issues/1546

appium/tests/specs/mock/setup/SetupOnlyRevokedKeyFromEKM.spec.ts
ioanmo226: https://github.com/FlowCrypt/flowcrypt-ios/issues/1546#issuecomment-1111980975

tomholub: At least for now, `Missing sender public key. Is FlowCrypt iOS app well set up?` should become `None of your account public keys are usable for encryption.` which will be slightly more correct than acting as if there are no keys (eg if one of them is revoked or expired).

ioanmo226: yeah, implemented that way.

tomholub (commented):
delta looks ok, will do another review on the whole PR

FlowCrypt/Core/Models/KeyDetails.swift
```diff
@@ -40,6 +40,15 @@ extension KeyDetails {
     var pgpUserEmails: [String] {
         users.map { MCOAddress(nonEncodedRFC822String: $0).mailbox }
     }
+    
+    var isKeyUsuable: Bool {
+        var expired = false
+        let now = Date()
+        if let expiration = expiration, TimeInterval(expiration) < now.timeIntervalSince1970 {
```
tomholub, line 47: Maybe `expiration.toDate() < Date()` or similar? I noticed there's an extension for this

Then you don't need a `let now` . You also don't need `expired = false`

Finally, in another issue I just thought to look at `lastModified` as well as uids

Roughly this:

```swift
// revoked keys are not usable
guard !revoked else { return false }
// keys without lastModified don't have valid signatures on them - not usable
guard lastModified else { return false }
// keys without uids on them are not usable
guard users.isNotEmpty else { return false } 
// expired keys are not usable
if let expiration = expiration, expiration.toDate() < Date() { return false } 
// non-revoked keys, with lastModified and at least one user, that are not expired are usable
// gross simplification until https://github.com/FlowCrypt/flowcrypt-ios/issues/1546
return true
```

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -166,7 +166,7 @@ final class ComposeMessageService {
 
         let subject = contextToSend.subject ?? "(no subject)"
 
-        guard let myPubKey = try storage.getKeypairs(by: sender).map(\.public).first else {
+        guard let keyPair = try await keyService.getKeyPair(from: sender) else {
             throw MessageValidationError.missingPublicKey
         }
```
tomholub, line 171: We have some mess in the keyService and storage methods that's hard to exactly understand or explain. Once the other PR is merged, I'll look into this to clean it up (in this PR).

ioanmo226: what do you mean `other PR`?

FlowCrypt/Core/Models/KeyDetails.swift
```diff
@@ -40,6 +40,20 @@ extension KeyDetails {
     var pgpUserEmails: [String] {
         users.map { MCOAddress(nonEncodedRFC822String: $0).mailbox }
     }
+    
+    var isKeyUsuable: Bool {
```
sosnovsky, line 44: typo, should be `isKeyUsable`

FlowCrypt/Core/Models/KeyDetails.swift
```diff
@@ -40,6 +40,20 @@ extension KeyDetails {
...
+        // revoked keys are not usable
+        guard !revoked else { return false }
+        // keys without lastModified don't have valid signatures on them - not usable
+        guard lastModified != nil else { return false }
+        // keys without uids on them are not usable
+        guard users.isNotEmpty else { return false }
+        // expired keys are not usable
+        if let expiration = expiration, expiration.toDate() < Date() { return false }
```
sosnovsky, line 52: You can use `timeIntervalSinceNow` here, so no need to initialise another `Date()`:
```swift
if let expiration = expiration, expiration.timeIntervalSinceNow < 0 { return false }
```

FlowCrypt/Core/Models/KeyDetails.swift
ioanmo226: Thanks for notice

sosnovsky:
@ioanmo226 looks like `CheckRefreshKeyFromEkmErrorHandling` fails after `master` merge. Test screenshot in Semaphore shows iOS home screen, so probably app crashes for some reason.

ioanmo226:
Yeah, I guess so. This CheckRefreshKeyFromEkmErrorHandling fails for this PR. Needs to be updated.

ioanmo226:
@tomholub Have you already started working on this PR?
We need to fix CheckRefreshKeyFromEkmErrorHandling test failure(crash) first.

tomholub:
should
 - [ ] remove all methods that pull and choose a single key from KeyService. Instead should pull all
 - [ ] add KeyMethods methods to then filter these down to a particular key
 - [ ] stop using PrvKeyInfo where I can. Use Keypair instead. Only leave PrvKeyInfo when calling core (make a method Keypair -> PrvKeyInfo)
 - [ ] remove KeyService. Remaining method `getAllPrvKeyInfos` should be moved to PassPhrase and renamed `addPassPhraseToKeys`, updated accordingly

FlowCrypt/Functionality/Pgp/KeyMethods.swift
```diff
@@ -63,8 +75,46 @@ final class KeyMethods: KeyMethodsType {
...
+
+    func chooseSenderSigningKey(keys: [Keypair], senderEmail: String) async throws -> Keypair? {
+        // TODO - needs to be implemented
+        return keys.first
+    }
+
+    func chooseSenderEncryptionKeys(keys: [Keypair], senderEmail: String) async throws -> [Keypair] {
+        // TODO - needs to be implemented
+        return keys
+    }
```
tomholub, line 91: Refactoring is more or less done. Now I need to actually implement these two methods.

This should make it a lot clearer what is going on.

tomholub (changes requested):

FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+Drafts.swift
```diff
@@ -42,7 +42,7 @@ extension ComposeViewController {
                     input: input,
                     contextToSend: contextToSend,
                     includeAttachments: false,
-                    viewController: self
+                    vcForPassPhraseModal: self
                 )
```
tomholub, line 46: I filed issue for this https://github.com/FlowCrypt/flowcrypt-ios/issues/1587

FlowCrypt/Functionality/Pgp/KeyMethods.swift
```diff
@@ -10,9 +10,38 @@ import FlowCryptCommon
...
+    func chooseSenderEncryptionKeys(keys: [Keypair], senderEmail: String) async throws -> [Keypair]
+}
+
+enum KeyMethodsError: Error {
+//    case unexpected
+    case parsingError
+    case retrieve
+//    case missingCurrentUserEmail
+    case expectedPrivateGotPublic
+}
```
tomholub, line 25: I'll look into potentially trimming these

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -73,20 +73,25 @@ final class ComposeMessageService {
         self.onStateChanged = completion
     }
 
-    func prepareSigningKey(viewController: UIViewController?) async throws -> PrvKeyInfo? {
-        guard let viewController = viewController, let signingKey = try await keyService.getSigningKey(email: sender) else {
+    func prepareSigningKey(viewController: UIViewController?) async throws -> Keypair? {
```
tomholub, line 76: here also https://github.com/FlowCrypt/flowcrypt-ios/issues/1587

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -73,20 +73,25 @@ final class ComposeMessageService {
...
+        // todo - sender email will differ from account email once
+        //   https://github.com/FlowCrypt/flowcrypt-ios/issues/1298 is done
+        let keys = try await keyAndPassPhraseStorage.getKeypairsWithPassPhrases(email: sender)
+        guard var signingKey = try await keyMethods.chooseSenderSigningKey(keys: keys, senderEmail: sender) else {
+            // todo - throw user error for missing signing key
             return nil
         }
```
tomholub, line 83: potentially todo in this PR. If not, then I'll file an issue or link existing issue

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -114,20 +119,20 @@ final class ComposeMessageService {
...
         // since pass phrase was entered (an inconvenient thing for user to do),
         //  let's find all keys that match and save the pass phrase for all
-        let allKeys = try await keyService.getPrvKeyInfo(email: sender)
+        let allKeys = try await keyAndPassPhraseStorage.getKeypairsWithPassPhrases(email: sender)
         guard allKeys.isNotEmpty else {
             // tom - todo - nonsensical error type choice https://github.com/FlowCrypt/flowcrypt-ios/issues/859
             //   I copied it from another usage, but has to be changed
-            throw KeyServiceError.retrieve
+            throw KeyMethodsError.retrieve
         }
```
tomholub, line 130: Could change this error now

FlowCrypt/Functionality/Services/EKMVcHelper.swift
```diff
@@ -170,11 +170,11 @@ final class EKMVcHelper: EKMVcHelperType {
...
         // since pass phrase was entered (an inconvenient thing for user to do),
         //  let's find all keys that match and save the pass phrase for all
-        let allKeys = try await appContext.keyService.getPrvKeyInfo(email: appContext.user.email)
+        let allKeys = try await appContext.keyAndPassPhraseStorage.getKeypairsWithPassPhrases(email: appContext.user.email)
         guard allKeys.isNotEmpty else {
             // tom - todo - nonsensical error type choice https://github.com/FlowCrypt/flowcrypt-ios/issues/859
             //   I copied it from another usage, but has to be changed
-            throw KeyServiceError.retrieve
+            throw KeyMethodsError.retrieve
         }
```
tomholub, line 178: could fix this here

FlowCrypt/Models/Common/Keypair.swift
```diff
@@ -34,4 +34,39 @@ extension Keypair {
...
+        self.private = privateKey
+        self.public = k.public
+        self.passphrase = passPhrase
+        self.source = source
+        self.allFingerprints = k.fingerprints
+        self.allLongids = k.ids.map { $0.longid }
+        guard let lastModified = k.lastModified else {
+            // todo - make a new error like `keyMissingSelfSignature`
+            throw KeyMethodsError.parsingError
+        }
```
tomholub, line 60: This new init method could probably be used wherever we have generated a new key during setup and are storing it. There is probably similar code written in the VC, or somewhere in setup procedure in "generate key" flow.

FlowCrypt/Functionality/Pgp/KeyMethods.swift
tomholub: Also there is `KeypairError` that overlaps. Could join them together.

FlowCrypt/Models/Common/Keypair.swift
tomholub: #1589

tomholub (approved):
code looks ok but tests probably need to be adjusted

tomholub (approved):

tomholub:
@ioanmo226 @sosnovsky next week please have a look and leave a review - it's quite a large change, so there are bound to be things to improve or change. I'm merging it now to avoid conflicts.

FlowCrypt/Core/Core.swift
```diff
@@ -87,7 +87,7 @@ actor Core: KeyDecrypter, KeyParser, CoreComposeMessageType {
         }
 
         let decrypted = try await call("decryptFile", jsonDict: [
-            "keys": try keys.map { try $0.toJsonEncodedDict() },
+            "keys": keys.map { $0.prvKeyInfoJsonDictForCore },
```
sosnovsky, line 90: Such `map` can be simplified to
```swift
keys.map(\.prvKeyInfoJsonDictForCore)
```

There are some other places in this PR where it can be used too.

FlowCrypt/Functionality/Services/Compose Message Service/ComposeMessageService.swift
```diff
@@ -73,20 +73,25 @@ final class ComposeMessageService {
...
+        guard var signingKey = try await keyMethods.chooseSenderSigningKey(keys: keys, senderEmail: sender) else {
+            // todo - throw user error for missing signing key
             return nil
         }
-
-        guard let existingPassPhrase = signingKey.passphrase else {
-            return signingKey.copy(with: try await self.requestMissingPassPhraseWithModal(for: signingKey, viewController: viewController))
+        if signingKey.passphrase == nil {
+            guard let vc = viewController else {
+                throw AppErr.unexpected("missing UIViewController when promting for pass phrase for signing key")
```
sosnovsky, line 86: typo `prompting`

FlowCrypt/Functionality/Pgp/KeyMethods.swift
```diff
@@ -63,8 +41,45 @@ final class KeyMethods: KeyMethodsType {
...
+        if let byPrimaryUid = filter(keys, usable, ({ $0.pgpUserEmailsLowercased.first == senderEmail })) {
+            return byPrimaryUid // if any keys match by primary uid, use them
+        }
+        if let byAnyUid = filter(keys, usable, ({ $0.pgpUserEmailsLowercased.contains(senderEmail) })) {
+            return byAnyUid // if any keys match by any uid, use them
+        }
+        return filter(keys, usable, { _ in true }) ?? [] // use any usable keys
+    }
+
+    private func filter(_ toReturn: [Keypair], _ parsed: [KeyDetails], _ criteria: (KeyDetails) -> Bool) -> [Keypair]? {
```
sosnovsky, line 74: What about making `criteria` parameter optional here?
So we don't need to write `{ _ in true }` to use all keys without filtering.
Then it'll be just `return filter(keys, usable) ?? []`

FlowCrypt/Functionality/Pgp/KeyMethods.swift
```diff
@@ -10,45 +10,23 @@ import FlowCryptCommon
...
+        var matching: [T] = []
         for key in keys {
-            guard let privateKey = key.private else {
-                throw KeyServiceError.expectedPrivateGotPublic
+            guard let privateKey = key.getArmoredPrv() else {
+                throw KeypairError.expectedPrivateGotPublic
             }
             do {
-                _ = try await self.decrypter.decryptKey(armoredPrv: privateKey, passphrase: passPhrase)
+                _ = try await Core.shared.decryptKey(armoredPrv: privateKey, passphrase: passPhrase)
```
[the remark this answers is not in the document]
tomholub: It was used for unit tests that we no longer have. Therefore no longer needed. (prod code wouldn't need this)

That means maybe `KeyDecrypter` type can be removed too? I haven't checked for other usages. Maybe it was only used here.

tomholub:
Thank you for the review - I'll file an issue for this.

FlowCrypt/Functionality/Pgp/KeyMethods.swift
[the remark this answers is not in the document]
sosnovsky: Yes, `KeyDecrypter` was used only here, so can be removed.

FlowCrypt/Functionality/Pgp/KeyMethods.swift
```diff
@@ -63,8 +41,45 @@ final class KeyMethods: KeyMethodsType {
...
+    private func filter(_ toReturn: [Keypair], _ parsed: [KeyDetails], _ criteria: (KeyDetails) -> Bool) -> [Keypair]? {
+        let filteredPrimaryFingerprints = parsed
+            .filter { criteria($0) }
+            .map { $0.primaryFingerprint }
+        let filteredKeypairs = toReturn
+            .filter { filteredPrimaryFingerprints.contains($0.primaryFingerprint) }
+        guard filteredKeypairs.isNotEmpty else {
+            return nil
+        }
+        return filteredKeypairs
```
ioanmo226, line 83: Do we need to return nil when array is empty when return type is array?
I think we don't need above guard  and `return filteredKeypairs` is enough.
And we can change `if let byPrimaryUid = filter(keys, usable, ({ $0.pgpUserEmailsLowercased.first == senderEmail }))` to `if let byPrimaryUid = filter(keys, usable, ({ $0.pgpUserEmailsLowercased.first == senderEmail })), byPrimaryUid.isNotEmpty`

FlowCrypt/Models/Realm Models/KeypairRealmObject.swift
```diff
@@ -45,25 +38,23 @@ extension KeypairRealmObject {
     convenience init(_ keyDetails: KeyDetails, passphrase: String?, source: KeySource, user: UserRealmObject) throws {
         self.init()
 
-        guard let privateKey = keyDetails.private else {
-            throw KeyInfoError.missingPrivateKey("storing pubkey as private")
+        guard let privateKey = keyDetails.private, keyDetails.isFullyEncrypted != nil else {
```
ioanmo226, line 41: Is `keyDetails.isFullyEncrypted != nil` check necessary here? As we check `keyDetails.isFullyEncrypted == true` at below line?
And maybe we need to take care of localizations? Error message is returned as static string here.

tomholub: Agree with localizations

nil is different than false. nil means it's not a private key. false means it's a private key that is not encrypted fully. Notice nil and false produces a different error, as implemented.

ioanmo226: 👍

ioanmo226: In that case we may update code like this.

```
guard let privateKey = keyDetails.private, let isFullyEncrypted = keyDetails.isFullyEncrypted else {
    throw KeypairError.missingPrivateKey("storing pubkey as private")
}
guard isFullyEncrypted else {
    throw KeypairError.notEncrypted("Will not store Private Key that is not fully encrypted")
}
```

tomholub: either is fine with me

PR #1549, opened 2022-04-29
#1509 Fix "Invalid Storage" error + use static keychain index

### LINKED ISSUE #1509: storage invalid still possibly happening on iOS v0.9.0
From my own email https://mail.google.com/mail/u/0/#sent/KtbxLzfhbgwNPZJkzZQkQwlHVkWNpQWcxq

> Hello [customer],
> 
> There were no storage changes following 0.8.1, which was meant to fix the issue. As far as we understood the issue, it is related to how the app goes to sleep / background and foreground, with regards to its permission to access the apple OS keychain. Before 0.8.1 the app could be in a state when it was pre-warmed (started in limited environment by OS in background) without access to key chain, which would make it think that it doesn't have any storage initialized, so it would re-initialize wrongly, corrupting previous storage.
> 
> I encountered the storage error about once a week before version 0.8.1, and never since.
> 
> In the email from Apr 15 it seems you encountered it following an update to 0.9.0 and then in email from Apr 19 you encountered it again, still on 0.9.0.
> 
> Please keep me informed on this issue as it happens, on your or anyone else's devices, and app versions and app update history to the extent it is known.
> 
> Thank you,
> Tom

sosnovsky:
This PR fixes "Invalid Storage" error:
- keychain prefix is stored in plain file, without encryption
- new keychain index isn't generated if database file already exists at target path

I also added `Retry` button on `Invalid Storage` error screen, so users will be able to retry database decryption before resetting it.

close #1509

- Difficult to test (explain why) - difficult to reproduce

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
+++ b/FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
@@ -88,7 +88,7 @@ final class EncryptedStorage: EncryptedStorageType {
             return Realm.Configuration(inMemoryIdentifier: UUID().uuidString)
         }
 
-        let path = try EncryptedStorage.getDocumentDirectory() + "/" + Constants.encryptedDbFilename
+        let path = try EncryptedStorage.path
         let latestSchemaVersion = currentSchema.version.dbSchemaVersion
 
         return Realm.Configuration(
@@ -272,6 +272,18 @@ extension EncryptedStorage {
 }
 
 extension EncryptedStorage {
+    static var path: String {
+        get throws {
+            try getDocumentDirectory() + "/" + Constants.encryptedDbFilename
+        }
+    }
+
+    static var isStorageExists: Bool {
+        get throws {
+            FileManager.default.fileExists(atPath: try path)
+        }
+    }
+
     static func getDocumentDirectory() throws -> String {
         guard let documentDirectory = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true).first else {
             throw AppErr.general("No path direction for .documentDirectory")
@@ -280,7 +292,6 @@ extension EncryptedStorage {
     }
 
     static func reset() throws {
-        let path = try getDocumentDirectory() + "/" + Constants.encryptedDbFilename
-        try FileManager.default.removeItem(atPath: path)
+        try FileManager.default.removeItem(atPath: try path)
     }
 }
--- a/FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
+++ b/FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
@@ -65,7 +65,6 @@ final class EncryptedStorage: EncryptedStorageType {
     private let currentSchema: EncryptedStorageSchema = .version8
     private let supportedSchemas = EncryptedStorageSchema.allCases
 
-    private let keyChainService: KeyChainService
     private let storageEncryptionKey: Data
 
     var storage: Realm {
@@ -77,7 +76,7 @@ final class EncryptedStorage: EncryptedStorageType {
     }
 
     init() async throws {
-        self.keyChainService = KeyChainService()
+        let keyChainService = KeyChainService()
         self.storageEncryptionKey = try await keyChainService.storageEncryptionKey
     }
 
--- a/FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
+++ b/FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
@@ -24,7 +24,7 @@ protocol EncryptedStorageType {
     func validate() throws
     func cleanup() throws
 
-    static func reset() throws
+    static func removeStorageFile() throws
 }
 
 final class EncryptedStorage: EncryptedStorageType {
@@ -59,16 +59,13 @@ final class EncryptedStorage: EncryptedStorageType {
         }
     }
 
-    private enum Constants {
-        static let encryptedDbFilename = "encrypted.realm"
-    }
-
     private lazy var migrationLogger = Logger.nested(in: Self.self, with: .migration)
     private lazy var logger = Logger.nested(Self.self)
 
     private let currentSchema: EncryptedStorageSchema = .version8
     private let supportedSchemas = EncryptedStorageSchema.allCases
 
+    private let keyChainService: KeyChainService
     private let storageEncryptionKey: Data
 
     var storage: Realm {
@@ -79,8 +76,9 @@ final class EncryptedStorage: EncryptedStorageType {
         }
     }
 
-    init(storageEncryptionKey: Data) {
-        self.storageEncryptionKey = storageEncryptionKey
+    init() async throws {
+        self.keyChainService = KeyChainService()
+        self.storageEncryptionKey = try await keyChainService.storageEncryptionKey
     }
 
     private func getConfiguration() throws -> Realm.Configuration {
@@ -274,11 +272,11 @@ extension EncryptedStorage {
 extension EncryptedStorage {
     static var path: String {
         get throws {
-            try getDocumentDirectory() + "/" + Constants.encryptedDbFilename
+            try getDocumentDirectory() + "/encrypted.realm"
         }
     }
 
-    static var isStorageExists: Bool {
+    static var doesStorageFileExist: Bool {
         get throws {
             FileManager.default.fileExists(atPath: try path)
         }
@@ -291,7 +289,7 @@ extension EncryptedStorage {
         return documentDirectory
     }
 
-    static func reset() throws {
+    static func removeStorageFile() throws {
         try FileManager.default.removeItem(atPath: try path)
     }
 }
--- a/FlowCrypt/Functionality/DataManager/Encrypted Storage/KeyChainService.swift
+++ b/FlowCrypt/Functionality/DataManager/Encrypted Storage/KeyChainService.swift
@@ -11,23 +11,6 @@ import Foundation
 import Security
 import UIKit
 
-enum KeyChainServiceError: Error {
-    case storageEncryptionKeyFetchFailed(String)
-}
-
-extension KeyChainServiceError: CustomStringConvertible {
-    var description: String {
-        switch self {
-        case .storageEncryptionKeyFetchFailed(let message):
-            return message
-        }
-    }
-}
-
-struct KeyChainConfig: Codable {
-    let indexPrefix: String
-}
-
 /// keychain is used to generate and retrieve encryption key which is used to encrypt local DB
 /// it does not contain any actual data or keys other than the db encryption key
 /// index of the keychain entry is dynamic (set up once per app installation), set in user defaults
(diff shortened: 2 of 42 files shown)


FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -272,6 +272,18 @@ extension EncryptedStorage {
 }
 
 extension EncryptedStorage {
+    static var path: String {
+        get throws {
+            try getDocumentDirectory() + "/" + Constants.encryptedDbFilename
+        }
```
tomholub, line 278: Is `Constants.encryptedDbFilename` only used here? If yes, you can remove that variable and write `+ "/encrypted.realm"`

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -272,6 +272,18 @@ extension EncryptedStorage {
...
+    }
+
+    static var isStorageExists: Bool {
```
tomholub, line 281: let's call it `doesStorageFileExist`. This re-evaluates on each call, right?

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -280,7 +292,6 @@ extension EncryptedStorage {
     }
 
     static func reset() throws {
```
tomholub, line 294: Let's rename to `removeStorageFile` to be explicit what happens here.

FlowCrypt/Functionality/DataManager/Encrypted Storage/KeyChainService.swift
```diff
@@ -19,23 +36,57 @@ actor KeyChainService {
     private let logger = Logger.nested("KeyChain")
     private let keyByteLen = 64
 
+    private let filesManager = FilesManager()
+    private let userDefaults = UserDefaults.standard
+
+    private enum Constants {
+        static let prefixUserDefaultsKey = "indexSecureKeychainPrefix"
```
tomholub, line 43: Let's call this `legacyKeychainIndexPrefixInUserDefaults` ? (`legacy` here means something we are moving away from)

tomholub (commented):
some easier comments - I have to open this code in IDE to review

FlowCrypt/Functionality/DataManager/Encrypted Storage/KeyChainService.swift
```diff
@@ -19,23 +36,57 @@ actor KeyChainService {
...
+        static let configFileName = "keychain-config"
+        static let indexSuffix = "-indexStorageEncryptionKey"
+    }
+
     /// this dynamic keychainIndex ensures that we use a different keychain index
-    ///   after deleting the app, because keychain entries survive app uninstall
+    /// after deleting the app, because keychain entries survive app uninstall
```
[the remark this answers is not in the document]
sosnovsky: There is no way to remove keychain entries on uninstall, but it's possible to delete them on the first run of the app after reinstall.
`UserDefaults` can be used for this, as they're removed on app uninstall. Then we can check some key in `UserDefaults` (for example `didAppLaunched`) - if key is nil, then it's the first run and all keychain items should be removed.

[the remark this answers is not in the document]
tomholub: Yes, except when `UserDefaults` get mysteriously dropped in some contexts 👎 :-)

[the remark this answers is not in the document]
tomholub: But I got the point. It's interesting the OS persists these.

sosnovsky:
@tomholub yeah, it will make our logic much simpler. And `Invalid storage` error should be fixed too, as it was caused by `UserDefaults` encryption, while Keychain part works well.

It's also possible to regenerate encryption key on app reinstall - on the first app start `doesStorageFileExist` will be `false`, then we can generate new encryption key and rewrite `app-storage-encryption-key` value in Keychain. What do you think?

tomholub:
> @tomholub yeah, it will make our logic much simpler. And `Invalid storage` error should be fixed too, as it was caused by `UserDefaults` encryption, while Keychain part works well.

👍 let's do that

> It's also possible to regenerate encryption key on app reinstall - on the first app start `doesStorageFileExist` will be `false`, then we can generate new encryption key and rewrite `app-storage-encryption-key` value in Keychain. What do you think?

Theoretically yes. But..

We were running into bugs somewhere between UserDefaults and NSFileProtectionComplete. Sometimes the contents wouldn't be accessible. Database file also uses NSFileProtectionComplete and presumably, in some contexts it won't be there. I'm afraid the app may be fooled into rewriting the encryption key, again.

tomholub:
To be clear, we are not 100% sure that the current problem is caused by `UserDefaults`. It's just the best guess. There is some chance the the keychain itself is buggy, or our use of it is buggy, somehow.

By cutting everything down to it's simplest form, we'll find out.

sosnovsky:
> We were running into bugs somewhere between UserDefaults and NSFileProtectionComplete. Sometimes the contents wouldn't be accessible. Database file also uses NSFileProtectionComplete and presumably, in some contexts it won't be there. I'm afraid the app may be fooled into rewriting the encryption key, again.

I think it shouldn't be an issue here, as with `UserDefaults` problem was because of encrypted file - app wasn't able to get its content and parse `keychain prefix`.

While here we'll just check if this file exists, we don't need to decrypt it, so in my opinion it will be more stable.

But if persisting encryption key between reinstalls is okay - then we can implement `encryption key` logic as in your initial proposal, it'll be simpler.

tomholub:
I don't want to take any chances this time, so let's go with the simplest option (persist).

If that works well for a year, we can consider rewriting it as per your proposal. :-)

tomholub:
Marking back as draft

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
sosnovsky: Yes, it re-checks storage file on each call

FlowCrypt/Functionality/DataManager/Encrypted Storage/KeyChainService.swift
```diff
@@ -19,67 +19,85 @@ actor KeyChainService {
...
 
-    @MainActor private func constructKeychainIndex(dynamicPart: String) -> String {
-        return dynamicPart + "-indexStorageEncryptionKey"
+    @MainActor private func fetchLegacyEncryptionKey() throws -> Data? {
+        let userDefaults = UserDefaults.standard
+        let prefixKey = Constants.legacyKeychainIndexPrefixInUserDefaults
+
+        guard let storedDynamicPart = userDefaults.string(forKey: prefixKey) else { return nil }
+
+        userDefaults.removeObject(forKey: prefixKey)
```
tomholub, line 49: I think this is a bit dangerous here - there is some chance that the above code successfully gets the prefix, but the code below fails before the actual key it's migrated to new place in key chain. That could cause trouble or make it harder to debug problems later.

Therefore I propose we only delete this from user defaults after we've successfully stored the key at new location.

FlowCrypt/Functionality/DataManager/Encrypted Storage/KeyChainService.swift
```diff
@@ -19,67 +19,85 @@ actor KeyChainService {
...
+        let storageKey = storedDynamicPart + "-indexStorageEncryptionKey"
+
+        guard let encryptionKey = try fetchEncryptionKey(property: storageKey) else {
+            if try !EncryptedStorage.doesStorageFileExist {
+                throw AppErr.general("KeyChainService: failed to get encryption key prefix")
+            }
```
tomholub, line 55: This is odd - I think it should test the opposite:

```swift
        guard let encryptionKey = try fetchEncryptionKey(property: storageKey) else {
            // we got index in user defaults but could not retrieve encryption key from it
            if try EncryptedStorage.doesStorageFileExist {
                // the storage file was already initiated before - problem
                throw AppErr.general("KeyChainService: got legacy dynamic prefix from user defaults but could not find entry in key chain based on it")
            }
            // the storage was not initiated yet - no problem
            return nil
        }
```

FlowCrypt/Functionality/DataManager/Encrypted Storage/KeyChainService.swift
```diff
@@ -19,67 +19,85 @@ actor KeyChainService {
...
+            return nil
+        }
+
+        return encryptionKey
     }
 
-    @MainActor private func generateAndSaveStorageEncryptionKey() throws {
-        logger.logInfo("generateAndSaveStorageEncryptionKey")
+    @MainActor private func generateStorageEncryptionKey() throws -> Data {
```
tomholub, line 62: As currently implemented, this method should still be called `generateAndSaveStorageEncryptionKey` - since calling the method will do both actions as currently implemented.

If you'd like to have two distinct methods called `generateStorageEncryptionKey` and `saveStorageEncryptionKey`, then you should not have one of them call the other, instead call both independently.

A method name should describe all important things that will happen when calling it, when possible, to reduce surprises.

appium/tests/specs/live/update/CheckAppAfterUpdateFromOldVersion.spec.ts
```diff
@@ -34,6 +34,8 @@ describe('UPDATE: ', () => {
     await driver.terminateApp(bundleId);
     //remove current app version
     await driver.removeApp(bundleId);
+    //reset keychain
+    await driver.execute('mobile: clearKeychains')
```
tomholub, line 38: This should not be part of tests if it doesn't happen in prod

tomholub (changes requested):

appium/tests/specs/live/update/CheckAppAfterUpdateFromOldVersion.spec.ts
sosnovsky: The problem here is because other tests using new build then we have `encryption key` stored in Keychain.

When we try to migrate from `UserDefaults` in old build to `Keychain` in new build - it fails because `Keychain` already has  `encryption key` from previous tests.

This situation shouldn't happen in prod as users won't have `encryption key` stored in `Keychain` when migrating from `UserDefaults`.
I can add handle for this situation - in case if decrypting with `Keychain` key fails then app will check if `UserDefaults` has encryption key and try to decrypt with it. Do we need such additional check?

tomholub: I see - I'll have a second look

FlowCrypt/Functionality/DataManager/Encrypted Storage/KeyChainService.swift
sosnovsky: Yeah, you're right - I forgot to rename it back after adding migration code.
Currently `saveStorageEncryptionKey` also used when migrating from `UserDefaults`, I'll move it's code back to `generateAndSaveStorageEncryptionKey` after removing migration functionality.

appium/tests/specs/live/update/CheckAppAfterUpdateFromOldVersion.spec.ts
tomholub: You are right, it looks good as it is. 👍 I misunderstood.

FlowCrypt/Functionality/DataManager/Encrypted Storage/KeyChainService.swift
sosnovsky: Fixed ✅

FlowCrypt/Functionality/DataManager/Encrypted Storage/KeyChainService.swift
sosnovsky: Ah, missed it when changing from `guard try` to `if try`, fixed now ✅

tomholub (approved):
Thanks!

PR #1579, opened 2022-05-05
#593 Use ifNotNil

### LINKED ISSUE #593: see if we can avoid both .map and force unwrapping when dealing with non-array nullables
I know this is probably my problem not coming from a Swift background, but to me a `.map` implies that `keyDetail.expiration` is some sort of an iterable, and you're applying a method over it to return something iterable again.

But that's not the case here - `keyDetail.expiration` is an `Int?` and so the `.map` on values like this will never stop confusing me :-( seems to be a common practice in Swift.

Is there a pattern possible in Swift that's a bit more obvious for a non-Swifter? A ternary operator maybe? Like `keyDetail.expiration != nil ? Date(timeIntervalSince1970: Double(keyDetail.expiration)) : nil` ?

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/566#discussion_r717843388_

ioanmo226:
This PR uses `ifNotNil` instead of `map` for optionals to avoid confusions for non-swifters.

close #593 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Services/Contacts Service/Models/PubKey.swift
+++ b/FlowCrypt/Functionality/Services/Contacts Service/Models/PubKey.swift
@@ -59,7 +59,7 @@ extension PubKey {
             armored: keyDetails.public,
             lastSig: keyDetails.lastModified.map { Date(timeIntervalSince1970: TimeInterval($0)) },
             lastChecked: Date(),
-            expiresOn: keyDetails.expiration.map { Date(timeIntervalSince1970: TimeInterval($0)) },
+            expiresOn: keyDetails.expiration.ifNotNil { Date(timeIntervalSince1970: TimeInterval($0)) },
             longids: longids,
             fingerprints: fingerprints,
             created: Date(timeIntervalSince1970: Double(keyDetails.created)),
--- a/FlowCryptCommon/Extensions/OptionalExtensions.swift
+++ b/FlowCryptCommon/Extensions/OptionalExtensions.swift
@@ -1,0 +1,20 @@
+//
+//  OptionalExtensions.swift
+//  FlowCryptCommon
+//
+//  Created by Ioan Moldovan on 5/5/22
+//  Copyright © 2017-present FlowCrypt a. s. All rights reserved.
+//
+
+import Foundation
+
+extension Optional {
+    public func ifNotNil<U>(_ transform: (Wrapped) throws -> U) rethrows -> U? {
+        switch self {
+        case .some(let value):
+          return .some(try transform(value))
+        case .none:
+          return .none
+        }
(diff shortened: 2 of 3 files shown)


FlowCrypt/Functionality/Services/Contacts Service/Models/PubKey.swift
```diff
@@ -59,7 +59,7 @@ extension PubKey {
             armored: keyDetails.public,
             lastSig: keyDetails.lastModified.map { Date(timeIntervalSince1970: TimeInterval($0)) },
             lastChecked: Date(),
-            expiresOn: keyDetails.expiration.map { Date(timeIntervalSince1970: TimeInterval($0)) },
+            expiresOn: keyDetails.expiration.ifNotNil { Date(timeIntervalSince1970: TimeInterval($0)) },
```
ioanmo226, line 62: @tomholub Is this what you are looking for?

FlowCryptCommon/Extensions/OptionalExtensions.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Copyright © 2017-present FlowCrypt a. s. All rights reserved.
+//
+
+import Foundation
+
+extension Optional {
+    public func ifNotNil<U>(_ transform: (Wrapped) throws -> U) rethrows -> U? {
+        switch self {
+        case .some(let value):
+          return .some(try transform(value))
```
sosnovsky, line 15: @ioanmo226 is `try` needed here?
It looks like `return .some(transform(value))` should work well here

ioanmo226: image

transform might throw error

sosnovsky: It's because you marked it with `throws` in method description.
Xcode won't give any errors after removing `throws` there:
```swift
public func ifNotNil<U>(_ transform: (Wrapped) -> U) -> U? {
        switch self {
        case .some(let value):
          return .some(transform(value))
        case .none:
          return .none
        }
    }
```

ioanmo226: yeah, correct. because closure inside ifNotNil might throw error

ioanmo226: we may want to do something like `expiration.ifNotNil { try getSomething($0) }`

sosnovsky: Ah, got it, then all good 👍

sosnovsky (approved):

Issue #1535, opened 2022-04-27
reported vulnerabilities in npm packages


tomholub:
<img width="935" alt="image" src="https://user-images.githubusercontent.com/6306961/165581341-751d2c4f-3129-49c0-a08f-6d3675e983d3.png">

Edited by mart:
Additionally, I should have to break this task into smaller chunks as (1) I need to update webpack 5 on `/Core` path (2) I'm having difficulties on updating of most vulnerable npm packages on the `/appium` path but I'm still into it.

- [x] #1580
- [x] #1597 (may need to wait for appium 2.0 non-beta version)

sosnovsky:
Currently `appium` and `Core` code uses the latest versions of npm packages without reported vulnerabilities.

PR #1580, opened 2022-05-06
Update webpack to version 5 to fix the vulnerability on glob-parent depedency

martgil:
This PR will update webpack to version 5 to fix the vulnerability on glob-parent dependency

image

related issue: #1535

### To be filled by reviewers


sosnovsky:
@martgil I tried to build updated `Core` with `npm run-script deploy` command, but iOS app gives error `"❗️[Js][Core] SyntaxError: Unexpected token \',\'. Cannot declare a lexical variable twice: \'ReadableStream\'., 39515:undefined"` on startup.

Will you be able to fix it in this PR or it'll be better to create another issue for this error?

Screenshot 2022-05-07 at 21 10 07

tomholub:
As long as it builds, it should be ok, even if it doesn't work when you use `npm run-script deploy`. We won't be able to update (deploy) the javascript for a while because we have updated OpenPGP.js to v5 #1362 but we haven't made the updated version work with JavaScriptCore yet #1379 .

tomholub:
That means currently, the the TypeScript code and the `.js.txt` file are out of sync, until we can update the source and build scripts in a way to make it palatable for the iOS app again.

sosnovsky (approved):

PR #1601, opened 2022-05-12
#1551 Fix ambiguity in account email in passphrase service

### LINKED ISSUE #1551: ambiguity in account email vs pass phrase methods
For `InMemoryPassPhraseStorage`, email field is not respected.
This also applies to `logOutUser` too.
To check/delete passphrase from user email in `InMemoryPassPhraseStorage` we need to pass KeyService or AppContext to get key pairs from user email first(And then get matching pass phrases from there).
But this seems big changes and need to be addressed in another issue I think.
@tomholub Please confirm.
 
image

_Originally posted by @ioanmo226 in https://github.com/FlowCrypt/flowcrypt-ios/pull/1510#discussion_r861765160_

ioanmo226:
This PR fixed ambiguity in account email in passphrase service

close #1551 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


FlowCrypt/Functionality/Services/Local Private Key Services/InMemoryPassPhraseStorage.swift
```diff
@@ -15,13 +15,16 @@ final class InMemoryPassPhraseStorage: PassPhraseStorageType {
...
         passPhraseProvider: InMemoryPassPhraseProviderType = InMemoryPassPhraseProvider.shared,
+        encryptedStorage: EncryptedStorageType,
         timeoutInSeconds: Int = 4*60*60 // 4 hours
     ) {
         self.passPhraseProvider = passPhraseProvider
         self.timeoutInSeconds = timeoutInSeconds
+        self.encryptedStorage = encryptedStorage
     }
 
     func save(passPhrase: PassPhrase) {
```
tomholub, line 30: either `PassPhrase` should contain `email` which should be used, or this method needs an `email` input and separate them.

As I said - all of these methods need an email input and need to keep pass phrases separate.

The `update`, `remove` methods too.

FlowCrypt/Functionality/Services/Local Private Key Services/PassPhraseService.swift
```diff
@@ -89,7 +90,7 @@ final class PassPhraseService: PassPhraseServiceType {
         case .persistent:
             try encryptedStorage.save(passPhrase: passPhrase)
         case .memory:
-            let storedPassPhrases = try encryptedStorage.getPassPhrases(for: "")
+            let storedPassPhrases = try encryptedStorage.getPassPhrases(from: passPhrase.primaryFingerprintOfAssociatedKey)
```
tomholub, line 93: either search by `email` and then filter by fingerprint below, or search here by `email+fingerprint` then no need to filter below. If searching by email, expect array. If you decide to write the method to accept email + fingerprint then expect optional one item and return the first found.

Don't use ambiguous input like `for`, `from` because it's not clear if it's email or fingerprint or something else to input. Just `email`, `primaryFingerprint` and similar for clarity.

FlowCryptAppTests/Functionality/Services/PassPhraseStorageTests/InMemoryPassPhraseStorageTest.swift
```diff
@@ -12,14 +12,17 @@ import XCTest
 class InMemoryPassPhraseStorageTest: XCTestCase {
     var sut: InMemoryPassPhraseStorage!
     var passPhraseProvider: InMemoryPassPhraseProviderType!
+    var encryptedStorageMock: EncryptedStorageMock!
```
tomholub, line 15: will get removed

FlowCryptAppTests/Mocks/EncryptedStorageMock.swift
```diff
@@ -38,6 +38,22 @@ final class EncryptedStorageMock: EncryptedStorageType {
         getKeypairsResult
     }
 
+    func mockGetKeyPairs(with fingerPrints: [String]) throws {
```
tomholub, line 41: same

tomholub (changes requested):
as above - thanks

FlowCrypt/App/AppContext.swift
```diff
@@ -106,9 +108,11 @@ class AppContext {
 
     @MainActor
     func getFoldersService() -> FoldersService {
+        let passPhraseStorage = InMemoryPassPhraseStorage(encryptedStorage: self.encryptedStorage)
         return FoldersService(
             encryptedStorage: self.encryptedStorage,
-            remoteFoldersProvider: self.getRequiredMailProvider().remoteFoldersProvider
+            remoteFoldersProvider: self.getRequiredMailProvider().remoteFoldersProvider,
+            trashPathStorage: LocalStorage(passPhraseStorage: passPhraseStorage)
```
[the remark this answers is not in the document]
ioanmo226: Because FolderService requires trashPathStorage as LocalStorage type. and local storage requires passphrase storage.
Check below comment.

FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
```diff
@@ -68,12 +68,19 @@ final class ThreadDetailsViewController: TableNodeViewController {
         }
         self.threadOperationsProvider = threadOperationsProvider
         self.messageOperationsProvider = appContext.getRequiredMailProvider().messageOperationsProvider
+        let localStorage = LocalStorage(
+            passPhraseStorage: InMemoryPassPhraseStorage(
+                encryptedStorage: appContext.encryptedStorage
+            )
+        )
```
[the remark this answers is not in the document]
ioanmo226: Not sure either, LocalStorage was supposed to save trash folder path and log out user(not sure logout method is in LocalStorage method)
Need to be discussed.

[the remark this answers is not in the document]
ioanmo226: cc: @sosnovsky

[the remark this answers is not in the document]
sosnovsky: > Not sure either, I think LocalStorage was supposed to save trash folder path and log out user(not sure why logout method is in LocalStorage method)
> Need to be discussed.

Not sure why I placed `logOut` there, but I think it'll be better to move it to `SessionService` as more appropriate place.
Then `LocalStorage` won't have `passPhraseStorage` dependency.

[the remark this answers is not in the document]
tomholub: Sounds good to me

I think Anton (early developer) had a tendency to put `xxxHandler` everywhere so we ended up with 10 logOutHandler all over the app. I don't like that approach. I prefer that there is one as Roma suggested, and then it calls the appropriate methods everywhere which should be named based on what they do, not just `logOutHandler`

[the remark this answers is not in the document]
tomholub: I will be traveling today. @ioanmo226 you can finalize the PR based on my and @sosnovsky comments. Then Roman please add your approval once you are happy. Once I get to it, I'll also approve to override my own change request, maybe at night.

ioanmo226:
@sosnovsky Please review

FlowCrypt/Functionality/DataManager/SessionService.swift
```diff
@@ -160,9 +156,7 @@ extension SessionService: SessionServiceType {
             logger.logWarning("currentAuthType is not resolved")
         }
         do {
-            for storage in self.storages {
-                try storage.logOutUser(email: user.email)
-            }
+            try encryptedStorage.logOutUser(email: user.email)
```
sosnovsky, line 159: @ioanmo226 after this change `LocalStorage.cleanup` won't be called on log out.

What I meant - we should just remove `try passPhraseStorage.logOutUser(email: email)` from `logOutUser` method of `LocalStorage` and call it in `SessionService` instead (as it's better place to store `passPhraseStorage` dependency)

ioanmo226: Ohh, I missed it. Thanks for your notice

sosnovsky: And we still need to add `try passPhraseStorage.logOutUser(email: email)` here 🙂
(which was previously called in `LocalStorage.logOutUser`)

ioanmo226: Yeah. Right 👍

sosnovsky (approved):

FlowCrypt/Functionality/Services/Local Private Key Services/InMemoryPassPhraseStorage.swift
```diff
@@ -57,12 +57,14 @@ final class InMemoryPassPhraseStorage: PassPhraseStorageType {
 
                 return isPassPhraseValid ? passPhrase : nil
             }
+            .filter { $0.email == email }
     }
 }
 
 extension InMemoryPassPhraseStorage: LogOutHandler {
     func logOutUser(email: String) throws {
```
tomholub, line 65: should be called `removePassPhrases` and doesnt have to be LogOutHandler. in another issue

tomholub (approved):

tomholub (approved):

PR #1609, opened 2022-05-13
#1587 Refactor missing pass phrase entry

### LINKED ISSUE #1587: to throw on missing pass phrase for signing instead of needing VC as argument
```swift
    func validateAndProduceSendableMsg(
        input: ComposeMessageInput,
        contextToSend: ComposeMessageContext,
        includeAttachments: Bool = true,
        // todo - throw Error on missing pass phrase then catch it in VC
        //   instead of passing VC it here
        vcForPassPhraseModal: UIViewController? = nil
    ) async throws -> SendableMsg {
```

The thrown error should be called `PromptUserToEnterPassPhraseForSigningKey` and can include the intended `Keypair` that we need pass phrase for.

ioanmo226:
This PR refactored missing pass phrase entry

close #1587 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+MessageSend.swift
+++ b/FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+MessageSend.swift
@@ -30,8 +30,7 @@ extension ComposeViewController {
 
         let sendableMsg = try await self.composeMessageService.validateAndProduceSendableMsg(
             input: self.input,
-            contextToSend: self.contextToSend,
-            vcForPassPhraseModal: self
+            contextToSend: self.contextToSend
         )
         UIApplication.shared.isIdleTimerDisabled = true
         try await composeMessageService.encryptAndSend(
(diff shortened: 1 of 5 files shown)


FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+MessageSend.swift
```diff
@@ -41,11 +40,44 @@ extension ComposeViewController {
...
+                },
+                onCompletion: { [weak self] passPhrase in
+                    guard let self = self else {
+                        return
+                    }
+                    Task<Void, Never> {
+                        do {
+                            let matched = try await self.composeMessageService.handlePassPhraseEntry(passPhrase, for: signingKey)
+                            if matched {
+                                return continuation.resume(returning: passPhrase)
```
ioanmo226, line 57: @tomholub 
With current refactor, once user enters correct pass phrase, we have to pass entered key to `validateAndProduceSendableMsg` method or somewhere like that.
And then need to check if this is for draft message or send message. (and things like that)

IMHO this makes logic more difficult and it's better to leave as it is now.
Let me know your thought.

tomholub: You don't have to. Instead, you can store the pass phrase in memory (InMemoryPassPhraseService I think) and then retry the whole operation (sending message) without any special arguments. As if user tapped on send again.

The next time it runs, it will pick up the pass phrase from memory. Should be ~ 2 lines of code.

ioanmo226: That sounds good. Will follow your suggestion 👍

FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+MessageSend.swift
```diff
@@ -41,11 +40,48 @@ extension ComposeViewController {
         handleSuccessfullySentMessage()
     }
 
+    internal func requestMissingPassPhraseWithModal(for signingKey: Keypair, isDraft: Bool = false) {
+        Task {
```
sosnovsky, line 44: `Task` is not needed here, as no async operations are performed

sosnovsky (changes requested):

FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+MessageSend.swift
ioanmo226: You are right 👍

PR #1617, opened 2022-05-17
#1603 Added uuid to email subject when composing

### LINKED ISSUE #1603: Add UUID to email subject which is composed during UI tests.
When we compose email in UI tests with below code, we delete sent email after email compose.
`await NewMessageScreen.composeEmail(recipient, emailSubject, emailText);`

The problem is when for some reason, ui tests fail after compose email but before deleting sent email, there might be multiple `emailSubject` emails.
And after that happens, UI tests will treat like sent email is not deleted even after successful deletion and tests will fail after that (As there are multiple emails with same subject).

So my suggestion is to add UUID to email subject when composing email.
If we do like this, even if specific UI test fails and composed email remains, further tests will not fail because of previous test.

Like this.

```
const uniqueId = Date.now().toString(36) + Math.random().toString(36).substring(2);
const testEmailSubject = `${emailSubject}-${uniqueId}`;
await NewMessageScreen.composeEmail(recipient, testEmailSubject, emailText);

```

ioanmo226:
This PR added uuid to email subject when composing

close #1603 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


ioanmo226:
As I checked, uuid infra was there already. Just used that.

sosnovsky (approved):

PR #1625, opened 2022-05-18
#1479 Updated Compose Message updateState

### LINKED ISSUE #1479: improve naming of compose message updateState
I see. We can leave it as is for now. Later, maybe it needs some renaming to make it clearer what it does.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1477#discussion_r843931517_

ioanmo226:
This PR updated Compose Message updateState

close #1479 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+ActionHandling.swift
+++ b/FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+ActionHandling.swift
@@ -22,7 +22,7 @@ extension ComposeViewController {
                                     .unique()
                                     .sorted()
 
-                updateView(with: .searchEmails(recipients))
+                updateView(newState: .searchEmails(recipients))
             } catch {
                 showAlert(message: error.localizedDescription)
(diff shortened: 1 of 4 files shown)


FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+ActionHandling.swift
```diff
@@ -22,7 +22,7 @@ extension ComposeViewController {
                                     .unique()
                                     .sorted()
 
-                updateState(with: .searchEmails(recipients))
+                updateView(with: .searchEmails(recipients))
```
tomholub, line 25: Could it be `updateView(newState: .searchEmails(recipients))` ? that would make it clearer. These `with:`,  `for:` etc keywords often obscure what the thing is if the variable doesn't make it obvious.

tomholub (approved):
:+1: thanks

PR #1626, opened 2022-05-18
issue 1488 update semaphoreci

### LINKED ISSUE #1488: update semaphore file swift/typescript tests
it's messy atm

tomholub:
This PR updates semaphoreci file, mostly to run a bit faster.

 - due to the way semaphoreci runs these, when tasks are defined as many commands (lines), it runs slower than when it's defined in fewer commands (lines), even if the actual shell commands are the same. That means stringing shell commands with `&&` instead of new line actually makes it run faster, at times (sometimes there are delays going from line to line, and sometimes these delays can be long). On the other hand, putting everything on one line makes it messy and harder to debug when a command goes wrong, so I'm trying to strike a balance.
 - I've changed the arrangement so that we only need two machines instead of one. Because our account has only 4 apple machines available, running on fewer machines will counter-intuitively make things run faster (when more than one set of tests is running in parallel, eg two competing PRs). // edit - actually I only reduced concurrent device usage from 4 to 3 per run. To keep Swift and TypeScript tests separate, and separate Appium tests.

close #1488

**Tests** 
- Test infrastructure updated

### To be filled by reviewers

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -30,8 +30,6 @@ blocks:
           - checkout && cd ~/git/flowcrypt-ios/ && cache restore && make dependencies && cache store
           - mv ~/appium-env ~/git/flowcrypt-ios/appium/.env
           - sem-version node 16 && cache restore appium-npm && cd ./appium && npm i && cd .. && cache store appium-npm appium/node_modules
-          - brew install ideviceinstaller
-          - npm install ios-deploy -g --unsafe-perm=true --allow-root && npm install ios-sim -g --unsafe-perm=true --allow-root
       jobs:
         - name: Appium UI tests
           commands:
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -10,6 +10,7 @@ auto_cancel:
   running:
     when: branch != 'master'
 blocks:
+
   - name: Appium UI tests
     dependencies: []
     run:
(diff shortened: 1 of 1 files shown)


.semaphore/semaphore.yml
```diff
@@ -26,13 +27,11 @@ blocks:
...
         commands:
-          - checkout && cd ~/git/flowcrypt-ios/
-          - cache restore && make dependencies && cache store
+          - checkout && cd ~/git/flowcrypt-ios/ && cache restore && make dependencies && cache store
           - mv ~/appium-env ~/git/flowcrypt-ios/appium/.env
           - sem-version node 16 && cache restore appium-npm && cd ./appium && npm i && cd .. && cache store appium-npm appium/node_modules
           - brew install ideviceinstaller
-          - npm install ios-deploy -g --unsafe-perm=true --allow-root
-          - npm install ios-sim -g --unsafe-perm=true --allow-root
+          - npm install ios-deploy -g --unsafe-perm=true --allow-root && npm install ios-sim -g --unsafe-perm=true --allow-root
```
sosnovsky, line 34: By the way, do we use `ideviceinstaller`, `ios-deploy` and `ios-sim` libs?
As I haven't found any other mentions of them in the code

tomholub: I have no clue. Guess I could try removing them.

tomholub: I especially don't like they are not versioned. Should be installed using npm instead, if needed.

tomholub:
At least this one is not needed:

image

sosnovsky (approved):

PR #1629, opened 2022-05-19
#387 get rid of fatalError and use throw instead

### LINKED ISSUE #387: replace all fatalError with throw

ioanmo226:
This PR get rid of fatalError and use throw instead.

close #387 // if this PR closes an issue

### To be filled by reviewers


FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
```diff
@@ -59,23 +59,21 @@ final class ThreadDetailsViewController: TableNodeViewController {
...
+        let threadOperationsProvider = try appContext.getRequiredMailProvider().threadOperationsProvider
         self.threadOperationsProvider = threadOperationsProvider
-        self.messageOperationsProvider = appContext.getRequiredMailProvider().messageOperationsProvider
+        self.messageOperationsProvider = try appContext.getRequiredMailProvider().messageOperationsProvider
         self.trashFolderProvider = TrashFolderProvider(
             user: appContext.user,
             foldersService: FoldersService(
                 encryptedStorage: appContext.encryptedStorage,
-                remoteFoldersProvider: appContext.getRequiredMailProvider().remoteFoldersProvider
+                remoteFoldersProvider: try appContext.getRequiredMailProvider().remoteFoldersProvider
```
sosnovsky, line 76: I think it'll be better to fetch `requiredMailProvider` only once like this `let mailProvider = try appContext.getRequiredMailProvider()` and then use it's properties:
```swift
let threadOperationsProvider = mailProvider.threadOperationsProvider
self.messageOperationsProvider = mailProvider.messageOperationsProvider
...
```

sosnovsky (commented):
Looks good, but there are a couple places where `fatalError` is still presented:
- https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-387/FlowCrypt/Models/Realm%20Models/KeypairRealmObject.swift#L59
- https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-387/FlowCrypt/App/AppStartup.swift#L223

FlowCrypt/Controllers/Threads/ThreadDetailsViewController.swift
ioanmo226: I agree

ioanmo226:
@sosnovsky, @tomholub 

I think we should fatalError for below 2 cases.

1. When user doesn't have email address, I thought we should fatalError. 
2. We already displayed error alert and when user clicks `OK`, it seemed me to reasonable to fatalError.

Let me know what to do for above cases.

> https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-387/FlowCrypt/Models/Realm%20Models/KeypairRealmObject.swift#L59
```
>   var account: String {
>         guard let email = user?.email else { fatalError() }
>         return email
>     }
```

https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-387/FlowCrypt/App/AppStartup.swift#L223
```
guard
            let user = try? appContext.encryptedStorage.activeUser,
            let authType = user.authType
        else {
            let sessionName = appContext.session?.description ?? "nil"
            let message = "error_wrong_app_state".localizeWithArguments(sessionName)

            logger.logError(message)

            if window.rootViewController == nil {
                window.rootViewController = UIViewController()
            }

            window.rootViewController?.showAlert(
                title: "error".localized,
                message: message,
                onOk: { fatalError() }
            )

            return
        }
```

tomholub:
> We already displayed error alert and when user clicks OK, it seemed me to reasonable to fatalError.

Doesn't seem reasonable to me. It may be an error alert based on some action, and after tapping ok you may have a chance to try the action again or do something else. Don't see any reason to crash the app intentionally.

> When user doesn't have email address, I thought we should fatalError.

I wouldn't. I'd just throw, then show an appropriate error, like for other errors.

sosnovsky:
I agree with @tomholub - it'll be better to handle errors in these 2 places, as just throwing `fatalError` can lead to constant crashes on app startup, so users will need to reinstall app.
When something wrong with user data (like empty email or app has invalid state) we can add `Re-login` button to modal, which will lead to sign in process.

FlowCrypt/Functionality/DataManager/Encrypted Storage/EncryptedStorage.swift
```diff
@@ -176,14 +176,15 @@ extension EncryptedStorage {
     }
 
     func getKeypairs(by email: String) throws -> [Keypair] {
-        return try storage.objects(KeypairRealmObject.self).where({
-            $0.account == email
+        // Can't ues where because `where` doesn't support throws
```
sosnovsky, line 179: What about using `user.email` directly, like this:
```swift
return try storage.objects(KeypairRealmObject.self).where({
    $0.user.email == email
})
```

ioanmo226: Yeah. In that way,  that's fine. But `account` variable declaration becomes useless then.

sosnovsky: Then I think we can remove `account` variable, as it just adds more complexity to code, and use `user.email` directly.

ioanmo226: what do you suggest?

ioanmo226: Ok, I agree

PR #1635, opened 2022-05-20
#1547 Remove == operator for PassPhrase and KeyDetails

### LINKED ISSUE #1547: get rid of == override for KeyDetails
There are three ways to possibly compare key details:
 - is the fingerprint the same?
 - is the fingerprint and last modified the same?
 - is all of the information the same?

Whenever one uses `==` I'd rather that the comparison was explicit like `keyDetails1.id == keyDetails2.id` or similar.

Therefore I'd like to remove 

```swift
    static func == (lhs: KeyDetails, rhs: KeyDetails) -> Bool {
        lhs.ids == rhs.ids
    }
```

To prevent unnecessary surprises.

ioanmo226:
This PR removed == operator for PassPhrase and KeyDetails

close #1547 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
+++ b/FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
@@ -238,9 +238,11 @@ extension SetupManuallyEnterPassPhraseViewController {
     }
 
     private func importKeys(with existedKeys: [KeyDetails], and passPhrase: String) throws {
-        let keysToUpdate = existedKeys.filter { existedKey in
-            return fetchedKeys.contains(where: { $0.fingerprints == existedKey.fingerprints })
-        }
+        let keysToUpdate = existedKeys
+            .getUniqueByFingerprintByPreferingLatestLastModified()
+            .filter { existedKey in
+                return fetchedKeys.contains(where: { $0.fingerprints == existedKey.fingerprints })
+            }
         let newKeysToAdd = fetchedKeys.filter { fetchedKey in
             return !existedKeys.contains(where: { $0.fingerprints == fetchedKey.fingerprints })
         }
--- a/FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
+++ b/FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
@@ -237,14 +237,14 @@ extension SetupManuallyEnterPassPhraseViewController {
         try importKeys(with: keyDetails, and: passPhrase)
     }
 
-    private func importKeys(with existedKeys: [KeyDetails], and passPhrase: String) throws {
-        let keysToUpdate = existedKeys
+    private func importKeys(with existingKeys: [KeyDetails], and passPhrase: String) throws {
+        let keysToUpdate = existingKeys
             .getUniqueByFingerprintByPreferingLatestLastModified()
-            .filter { existedKey in
-                return fetchedKeys.contains(where: { $0.fingerprints == existedKey.fingerprints })
+            .filter { existingKey in
+                return fetchedKeys.contains(where: { $0.fingerprints == existingKey.fingerprints })
             }
         let newKeysToAdd = fetchedKeys.filter { fetchedKey in
-            return !existedKeys.contains(where: { $0.fingerprints == fetchedKey.fingerprints })
+            return !existingKeys.contains(where: { $0.fingerprints == fetchedKey.fingerprints })
         }
 
         try appContext.encryptedStorage.putKeypairs(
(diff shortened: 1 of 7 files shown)


FlowCrypt/Core/Models/KeyAlgo.swift
```diff
@@ -14,3 +14,9 @@ struct KeyAlgo: Decodable {
     let bits: Int?
     let curve: String?
 }
+
+extension KeyAlgo: Equatable {
+    static func == (lhs: KeyAlgo, rhs: KeyAlgo) -> Bool {
+        lhs.algorithmId == rhs.algorithmId
```
tomholub, line 20: Should equate all fields. The curve may still be different, amount of bits may still be different, etc.

tomholub:
`keyDetails1 == keyDetails2` should mean all fields are different, in either case.

You should check where this `Set(listOfKeyDetails)` is used, and make sure that they are unique in another way.

As implemented, it's too obviousness where such functionality is employed which is not good.

ioanmo226:
To make `Set` from struct, it requires struct to confirm `Hashable` protocol and `Hashable` protocol requires to confirm to `Equatable` Protocol.
Therefore, we had custom `==` operator which we compare fingerprints(ids) of KeyDetails.

I don't think we can make `Set` without confirming to `Hashable` protocol.

tomholub:
It's used here:

```swift
        let matchingKeyBackups = Set(try await keyMethods.filterByPassPhraseMatch(keys: backups, passPhrase: passPhrase))
```

`Set` by only fingerprint is silly here, anyway. If there are two keys with the same fingerprint, should choose the version that has newer `lastModified`. You can write a method for this: `keyDetailsArray.getUniqueByFingerprintByPreferingLatestLastModified()` and then you could unit-test that method instead.

tomholub:
> I don't think we can make Set without confirming to Hashable protocol.

We don't need to use a `Set` at all. It was apparently only used to make the entries unique by fingerprint, which anyway was not the right thing to do, as implemented.

ioanmo226:
Now, I got you. Then let me get rid of `Set` here and implement your proposed solution.

FlowCrypt/Controllers/Setup/SetupManuallyEnterPassPhraseViewController.swift
```diff
@@ -238,8 +238,12 @@ extension SetupManuallyEnterPassPhraseViewController {
     }
 
     private func importKeys(with existedKeys: [KeyDetails], and passPhrase: String) throws {
```
tomholub, line 240: `existedKeys` should be called `existingKeys` :-) thanks

ioanmo226: ok

tomholub:
Will let Roma review

FlowCrypt/Functionality/Services/Local Private Key Services/CombinedPassPhraseStorage.swift
```diff
@@ -35,12 +35,13 @@ struct PassPhrase: Codable, Hashable, Equatable {
...
-    // itself, and not by primary fingerprint of the associated key. I understand this is being used somewhere,
-    // but I suggest to refactor it to avoid defining this == overload.
+    // We still need == operator here because we use `withUpdatedDate` to set `date` field to up-to-date
+    // Therfore, 2 passphrases might be treated differently even though
+    // they are exactly same if we don't implement custom == operator
     static func == (lhs: PassPhrase, rhs: PassPhrase) -> Bool {
-        lhs.primaryFingerprintOfAssociatedKey == rhs.primaryFingerprintOfAssociatedKey
+        return lhs.primaryFingerprintOfAssociatedKey == rhs.primaryFingerprintOfAssociatedKey
+                || lhs.value == rhs.value
+                || lhs.email == rhs.email
```
sosnovsky, line 44: Is it okay that passphrases will be equal even when they have different `primaryFingerprintOfAssociatedKey` but the same `email`?

FlowCrypt/Core/Models/KeyDetails.swift
```diff
@@ -92,4 +84,18 @@ extension Array where Element == KeyDetails {
     var joinedPrivateKey: String {
         compactMap(\.private).joined(separator: "\n")
     }
+    
+    func getUniqueByFingerprintByPreferingLatestLastModified() -> [KeyDetails] {
+        var newArray: [KeyDetails] = []
```
sosnovsky, line 89: It'll be better to use more meaningful array name here, like `uniqueKeyDetails`

FlowCrypt/Functionality/Services/Local Private Key Services/CombinedPassPhraseStorage.swift
ioanmo226: OHhh, no. wrong implementation.
Should be `&&`, not `||`.
Thanks for your notice

PR #1636, opened 2022-05-20
#1171 Put db calls into a storage method

### LINKED ISSUE #1171: related db calls should be put into a storage method that has a tx
This should be made a transaction in storage (can be another PR later) else it produces inconsistent results.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1165#discussion_r762539199_

ioanmo226:
This PR put db calls into a storage method

close #1171 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Services/Local Contacts Service/LocalContactsProvider.swift
+++ b/FlowCrypt/Functionality/Services/Local Contacts Service/LocalContactsProvider.swift
@@ -127,8 +127,8 @@ extension LocalContactsProvider: LocalContactsProviderType {
         guard let keys = keys else {
             return
         }
-        for key in keys {
-            try storage.write {
+        try storage.write {
+            for key in keys {
                 storage.delete(key)
             }
         }


ioanmo226:
@tomholub I couldn't find anywhere for such case.
So I only updated `removePubKey` function.

> applies to any place that we are doing more than one db write call within some method. Whether that is done in a loop or updating two tables etc.

sosnovsky:
@ioanmo226 I found one more place with this case - https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/App/AppStartup.swift#L58

ioanmo226:
I also thought of that.

But I didn't change it because of following reasons.
1. It had `await` syntax in for loop and therefore we can't wrap for loop in storage.write.
2. That code should be removed after a few days/months.

sosnovsky:
Ah, I see, then everything is good 👍

sosnovsky (approved):

PR #1643, opened 2022-05-24
#1560 Public key revoke logic

### LINKED ISSUE #1560: public key update - better handling of revoked keys
Currently, if we receive a newer version of the key (during compose or during signature verification), we will update our own older key with the newer version of the key.

We should update this mechanism to ensure that if the key in our own storage is revoked, it will never be updated anymore, even if we received a newer version of that key.

ioanmo226:
This PR implemented public key revoke logic

close #1560 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


appium/tests/specs/mock/setup/CheckRevokedKeyUpdate.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+import {
+  SplashScreen
+} from '../../../screenobjects/all-screens';
+import MailFolderScreen from "../../../screenobjects/mail-folder.screen";
+import NewMessageScreen from "../../../screenobjects/new-message.screen";
+import SetupKeyScreen from "../../../screenobjects/setup-key.screen";
+
+describe('SETUP: ', () => {
+
+  it('check revoked key update logic', async () => {
```
tomholub, line 12: Let's rename it as `will not update a revoked public key with valid one`

appium/tests/specs/mock/setup/CheckRevokedKeyUpdate.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+          [contactEmail]: ekmKeySamples.key0Updated.pub!
+        }
+      };
+
+      await NewMessageScreen.deleteAddedRecipient(0);
+
+      // Should display name in red background(which means revoked key) even though we receive new updated key
+      // Because we already have revoked key on local
+      await NewMessageScreen.setAddRecipient(contactEmail);
+      await NewMessageScreen.checkAddedRecipientColor(contactName, 0, 'red');
```
tomholub, line 50: I think the implementation broke general key update functionality. You can check it by adding the following steps at the end:
 - delete added recipient
 - set Attester to some other unrelated public key that is valid
 - add recipient again, and wait for it to update (maybe add two seconds to give it a chance to update)

You should see that the recipient should turn green. But as implemented, I think it will wrongly stay red.

tomholub (changes requested):

appium/tests/specs/mock/setup/CheckRevokedKeyUpdate.spec.ts
ioanmo226: Ok

appium/tests/specs/mock/setup/CheckRevokedKeyUpdate.spec.ts
ioanmo226: I added test and it works fine.

appium/tests/specs/mock/setup/CheckRevokedKeyUpdate.spec.ts
```diff
@@ -48,6 +48,17 @@ describe('SETUP: ', () => {
...
+      // Now check if other recipients display fine 
+      const testEmail = 'test@example.com';
+      mockApi.attesterConfig = {
+        servedPubkeys: {
+          [testEmail]: ekmKeySamples.key1.pub!
+        }
+      };
+      await NewMessageScreen.deleteAddedRecipient(0);
+      await NewMessageScreen.setAddRecipient(testEmail);
+      await NewMessageScreen.checkAddedRecipientColor('Test2', 0, 'green');
```
tomholub, line 61: Not another recipient. Another functioning key (with different fingerprint than the revoked one) for the same recipient as before.

ioanmo226: You were indeed correct. 👍 
It broke functionality you mentioned and I fixed that issue.

FlowCrypt/Functionality/Api/Remote Pub Key Apis/PubLookup.swift
```diff
@@ -69,6 +69,12 @@ class PubLookup: PubLookupType {
 
     func fetchRemoteUpdateLocal(with recipient: Recipient) async throws -> RecipientWithSortedPubKeys {
         let recipient = try await self.lookup(recipient: recipient)
+        // Return local contact(with revoked key) if local stored key is revoked key and fingerprints are same
+        if let contact = try await localContactsProvider.searchRecipient(with: recipient.email),
+           contact.keyState == .revoked,
+           contact.pubKeys == recipient.pubKeys {
+            return contact
+        }
```
tomholub, line 77: I don't fully understand this code - it seems to have not much effect. If `contact.pubKeys == recipient.pubKeys` means that all already existing pubkeys are the same as the received pubkeys, then they won't get updated even if you removed these lines.

I think you don't need this - try removing it. The implementation in `LocalContactsProvider` looks good to me.

tomholub (changes requested):

FlowCrypt/Functionality/Services/Local Contacts Service/LocalContactsProvider.swift
```diff
@@ -163,7 +163,10 @@ extension LocalContactsProvider {
     }
 
     private func update(pubKey: PubKey, for recipient: RecipientRealmObject, at index: Int) throws {
```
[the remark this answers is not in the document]
ioanmo226: Sounds good

FlowCrypt/Functionality/Api/Remote Pub Key Apis/PubLookup.swift
ioanmo226: This logic is needed. We first try to get contact from local (to show recipient to user) and then try to get it from remote.
If we remove this logic, newer key(same fingerpirnt as revoked key) won't be saved into local but recipient will be displayed as correct key user.(with green background with new key username)

image

FlowCrypt/Functionality/Services/Local Contacts Service/LocalContactsProvider.swift
```diff
@@ -162,17 +162,20 @@ extension LocalContactsProvider {
         }
     }
 
-    private func update(pubKey: PubKey, for recipient: RecipientRealmObject, at index: Int) throws {
+    private func update(storedRecipient: RecipientRealmObject, updateAtIndex: Int, newPubKey: PubKey) throws {
```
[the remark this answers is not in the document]
ioanmo226: Yeah, sure

FlowCrypt/Functionality/Api/Remote Pub Key Apis/PubLookup.swift
```diff
@@ -69,6 +69,12 @@ class PubLookup: PubLookupType {
 
     func fetchRemoteUpdateLocal(with recipient: Recipient) async throws -> RecipientWithSortedPubKeys {
         let recipient = try await self.lookup(recipient: recipient)
+        // Return local contact(with revoked key) if local stored key is revoked key and fingerprints are same
+        if let contact = try await localContactsProvider.searchRecipient(with: recipient.email),
+           contact.keyState == .revoked,
+           contact.primaryFingerprint == recipient.primaryFingerprint {
```
ioanmo226, line 75: Changed logic as follows. Let me know your thought

ioanmo226: @tomholub

tomholub: See other comment. I really think this should be removed. If it does have any effect at all (I don't think it does), then it should be achieved differently / in another place.

ioanmo226: That's why we have revoked key state check here. 

```
contact.keyState == .revoked,
contact.pubKeys == recipient.pubKeys
```

ioanmo226: So here's logic.

1. We get recipient info from remote.
2. Get locally saved recipient
3. If locally saved recipient key state is revoked & locally saved recipient pubkeys & remote recipient pubkeys are same, then use local saved recipient (with revoke status & revoke key user name) to display recipient in recipient 
input.
4. If not, then use new remote recipient info to display recipient in recipient input

tomholub: > That's why we have revoked key state check here.
> 
> ```
> contact.keyState == .revoked,
> contact.pubKeys == recipient.pubKeys
> ```

Doesn't change what I said. 

swift
```
if a == b && b.revoked == true {
  return b
}
return a
```

can be written as

```swift
return a
```

Same effect, provided that `a == b` means `a deep equals b`

tomholub: > If locally saved recipient key state is revoked & locally saved recipient pubkeys & remote recipient pubkeys are same, then use local saved recipient (with revoke status & revoke key user name) to display recipient in recipient input.

if locally saved pubkeys are the same as received pubkeys, then it doesn't matter whether you render local or remote. They are the same. Regardless whether one of them is revoked or not.

ioanmo226: I think we are on the different line.
Current logic is not same as this one.
```
if a == b && b.revoked == true {
  return b
}
return a
```

I added comments at every lines to let everyone better understands.
 
```
let recipient = try await self.lookup(recipient: recipient) // This is recipient info from remote
if let contact = try await localContactsProvider.searchRecipient(with: recipient.email), // First check if we have local saved contact
  contact.keyState == .revoked, // Check if local contact key is revoked because if local contact key state is revoked, we should not use further keys with same fingerprints. Instead we should use local revoked key from now on
  contact.pubKeys == recipient.pubKeys { // Even local saved key is revoked key, we should use newer(remote) key if fingerprints(pubkeys) are not same
   return contact
}
try localContactsProvider.updateKeys(for: recipient)
return recipient

```

@tomholub please check

ioanmo226: I think we are on the different line.
Current logic is not equal to this one.

```
if a == b && b.revoked == true {
  return b
}
return a
```

I added comments to let everyone better understands.

```
let remoteRecipient = try await self.lookup(recipient: recipient) // This is recipient info from remote
if let localContact = try await localContactsProvider.searchRecipient(with: recipient.email), // First check if we have local saved contact
  localContact.keyState == .revoked, // Check if local contact key is revoked because if local contact key state is revoked, we should not use further keys with same fingerprints. Instead we should use local revoked key from now on
  localContact.pubKeys == remoteRecipient.pubKeys { // Even local saved key is revoked key, we should use newer(remote) key if fingerprints(pubkeys) are not same
   return localContact
}
try localContactsProvider.updateKeys(for: remoteRecipient)
return remoteRecipient
```

@tomholub Please check

FlowCrypt/Functionality/Api/Remote Pub Key Apis/PubLookup.swift
```diff
@@ -68,8 +68,13 @@ class PubLookup: PubLookupType {
...
-        return recipient
+        let remoteRecipient = try await self.lookup(recipient: recipient) // This is recipient info from remote
+        if let localContact = try await localContactsProvider.searchRecipient(with: recipient.email), // First check if we have local saved contact
+          localContact.keyState == .revoked, // Check if local contact key is revoked because if local contact key state is revoked, we should not use further keys with same fingerprints. Instead we should use local revoked key from now on
+          localContact.pubKeys == remoteRecipient.pubKeys { // Even local saved key is revoked key, we should use newer(remote) key if fingerprints(pubkeys) are not same
+           return localContact
+        }
+        try localContactsProvider.updateKeys(for: remoteRecipient)
+        return remoteRecipient
     }
```
tomholub, line 79: I like the rename - it's clearer. My comment still stands. The `if` branch will only execute when `localContact.pubKeys == remoteRecipient.pubKeys`. If they are equal, why not return `remoteContact`? Therefore it could be rewritten as:

```swift
        let remoteRecipient = try await self.lookup(recipient: recipient) // This is recipient info from remote
        if let localContact = try await localContactsProvider.searchRecipient(with: recipient.email), // First check if we have local saved contact
          localContact.keyState == .revoked, // Check if local contact key is revoked because if local contact key state is revoked, we should not use further keys with same fingerprints. Instead we should use local revoked key from now on
          localContact.pubKeys == remoteRecipient.pubKeys { // Even local saved key is revoked key, we should use newer(remote) key if fingerprints(pubkeys) are not same
           return remoteRecipient // you just checked it's the same
        }
        try localContactsProvider.updateKeys(for: remoteRecipient)
        return remoteRecipient
```

Further, when `localContact.pubKeys == remoteRecipient.pubKeys` then `updateKeys` is a no-op. Nothing to update. Therefore can be rewritten as:

```swift
        let remoteRecipient = try await self.lookup(recipient: recipient) // This is recipient info from remote
        try localContactsProvider.updateKeys(for: remoteRecipient)
        return remoteRecipient
```

ioanmo226: There's exception when local contact key state is revoked.
Let's say you have revoked key in local and you receive updated key (with same fingerprint)
With below code, app will try to display updated key info. (In this case, we should display revoked key information to user as same fingerprint key should never be replaced)
Let me know if it makes sense.

```
let remoteRecipient = try await self.lookup(recipient: recipient) // This is recipient info from remote
try localContactsProvider.updateKeys(for: remoteRecipient)
return remoteRecipient
```
> The if branch will only execute when localContact.pubKeys == remoteRecipient.pubKeys. If they are equal, why not return remoteContact?

tomholub: >  Let's say you have revoked key in local and you receive updated key (with same fingerprint)

If received public keys are different from local public keys (received is valid and local is revoked), then `localContact.pubKeys == remoteRecipient.pubKeys` **will be false** and this branch will not execute. You will and up executing the following:

```swift
        try localContactsProvider.updateKeys(for: remoteRecipient)
        return remoteRecipient
```

FlowCrypt/Functionality/Api/Remote Pub Key Apis/PubLookup.swift
```diff
@@ -68,8 +68,13 @@ class PubLookup: PubLookupType {
...
-        try localContactsProvider.updateKeys(for: recipient)
-        return recipient
+        let remoteRecipient = try await self.lookup(recipient: recipient) // This is recipient info from remote
+        if let localContact = try await localContactsProvider.searchRecipient(with: recipient.email), // First check if we have local saved contact
+          localContact.keyState == .revoked, // Check if local contact key is revoked because if local contact key state is revoked, we should not use further keys with same fingerprints. Instead we should use local revoked key from now on
+           localContact.pubKeys.map(\.primaryFingerprint) == remoteRecipient.pubKeys.map(\.primaryFingerprint) { // Even local saved key is revoked key, we should use newer(remote) key if fingerprints(pubkeys) are not same
+           return localContact
+        }
+        try localContactsProvider.updateKeys(for: remoteRecipient)
+        return remoteRecipient
```
tomholub, line 78: Ok, this makes a bit more sense but looks brittle.

You can have `try localContactsProvider.updateKeys(for: remoteRecipient)` return the updated set of keys, including local ones. So the signature would be `try localContactsProvider.updateKeys(for: remoteRecipient) -> RecipientWithSortedPubKeys` or similar.

Then the whole method here could be:

```swift
        let remoteRecipient = try await self.lookup(recipient: recipient)
        return try localContactsProvider.updateKeys(for: remoteRecipient)
```

That would make implementation less brittle because you don't have to guess in which situations you should be returning which set of keys. I'm pretty sure there are edge cases in which the current implementation would have incorrect results.

tomholub (commented):
It looks better but it's brittle - see below.

FlowCrypt/Functionality/Api/Remote Pub Key Apis/PubLookup.swift
ioanmo226: Sounds good.

FlowCrypt/Functionality/Mail Provider/Message Provider/MessageService.swift
```diff
@@ -230,7 +230,8 @@ extension MessageService {
         let pubKeys = try localContactsProvider.retrievePubKeys(for: sender.email, shouldUpdateLastUsed: false)
         if pubKeys.isNotEmpty || onlyLocal { return pubKeys }
 
-        guard let contact = try? await pubLookup.fetchRemoteUpdateLocal(with: sender)
+        try await pubLookup.fetchRemoteUpdateLocal(with: sender)
```
tomholub, line 233: should use `try?` - I think this was intentional to ignore errors, since local pubkeys can be used most of the time, if present. Would be good to add a comment to explain this reason

FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+ActionHandling.swift
```diff
@@ -48,9 +48,12 @@ extension ComposeViewController {
                 }
 
                 let contact = Recipient(recipient: recipient)
-                let contactWithFetchedKeys = try await pubLookup.fetchRemoteUpdateLocal(with: contact)
+                try await pubLookup.fetchRemoteUpdateLocal(with: contact)
+
+                if let updatedContact = try await localContactsProvider.searchRecipient(with: recipient.email) {
+                    handleEvaluation(for: updatedContact)
+                }
```
tomholub, line 55: Looks good, just worried that there will be some inconsistent ui state if it comes back empty. Empty result should still be handled to display that the recipients has no pubkeys, I think?

tomholub (commented):
Looks pretty good for me, with some comments to fix.

I'll wait for @sosnovsky to do a review as well.

FlowCrypt/Functionality/Mail Provider/Message Provider/MessageService.swift
ioanmo226: ok

FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+ActionHandling.swift
ioanmo226: Not sure what you mean. it won't return empty result as we fetch remote data above that line.

FlowCrypt/Functionality/Api/Remote Pub Key Apis/PubLookup.swift
```diff
@@ -67,9 +67,11 @@ class PubLookup: PubLookupType {
...
     }
 
-    func fetchRemoteUpdateLocal(with recipient: Recipient) async throws -> RecipientWithSortedPubKeys {
-        let recipient = try await self.lookup(recipient: recipient)
-        try localContactsProvider.updateKeys(for: recipient)
-        return recipient
+    func fetchRemoteUpdateLocal(with recipient: Recipient) async throws {
+        let remoteRecipient = try await self.lookup(recipient: recipient)
+        if remoteRecipient.pubKeys.isEmpty {
+            throw ContactsError.keyMissing
```
ioanmo226, line 73: Added this throw when remote returns empty keys.
And it's handled here.
Let me know your thoughts @tomholub 

image

FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+ActionHandling.swift
ioanmo226: See below comment

FlowCrypt/App/AppStartup.swift
```diff
@@ -40,6 +40,31 @@ struct AppStartup {
...
+    @MainActor
+    private func checkAndUpdateIsRevoked(context: AppContextWithUser) async throws {
+        let revokedUpdatedFlag = "IS_PUB_KEY_REVOKED_UPDATED"
+        let userDefaults = UserDefaults.standard
+        if userDefaults.bool(forKey: revokedUpdatedFlag) {
+           return
+        }
+        // Added storage access directly because this function logic should be removed in the near future
+        let storage = try context.encryptedStorage.storage
+        let pubKeyObjects = storage.objects(PubKeyRealmObject.self).unique()
```
sosnovsky, line 54: Do we need `.unique()` here?

FlowCrypt/Functionality/Api/Remote Pub Key Apis/PubLookup.swift
tomholub: No public keys for a recipient is not an exceptional state, it's pretty normal. I wouldn't be throwing an exception for that. It's odd to see an exception handler for that, too. Normally, no public keys would be represented with an empty array.

In the code that I was commenting on, the result was optional and I was just wondering what would happen if it came back nil.

sosnovsky (commented):
Looks good for me 👍

FlowCrypt/App/AppStartup.swift
ioanmo226: Yeah, in addition to get unique array, it converts Collection to Array.

FlowCrypt/Functionality/Api/Remote Pub Key Apis/PubLookup.swift
ioanmo226: Do you want me to revert this throw logic?

tomholub: Ok. Please revert that. Else looks ok, as long as this is true:

> If fetchRemoteUpdateLocal succeed, localContactsProvider.searchRecipient would not return nil

ioanmo226: ok

tomholub (approved):

PR #1650, opened 2022-05-26
#1559 EKM keys refresh: allow removing keys, except when revoked + improve revoked handing

### LINKED ISSUE #1559: EKM keys refresh: allow removing keys, except when revoked + improve revoked handing
- [x] if key was removed from EKM, then we also remove it from local storage, UNLESS we already have a revoked version of that key. revoked keys cannot be ever removed from our storage (the app needs to continue to be aware that the key was revoked, so it can reject all future. 
- [x] never replace our own revoked key with a newer version of the key from EKM whatsoever

That means, if we ever get a revoked key, no more updates are allowed to that key. The key cannot be removed or updated anymore.

ioanmo226:
This PR updated EKM keys refresh logic(allow removing keys, except when revoked + improve revoked handing)

close #1559 // if this PR closes an issue

### To be filled by reviewers


appium/tests/specs/mock/setup/CheckRefreshKeyFromEkmWithPassPhrasePrompt.spec.ts
```diff
@@ -59,19 +59,7 @@ describe('SETUP: ', () => {
       await KeysScreen.openScreenFromSideMenu();
       await KeysScreen.checkKeysScreen([ekmKeySamples.key0, ekmKeySamples.key1]);
 
-      // stage 4 - modified key gets updated, removed key does not get removed
```
ioanmo226, line 62: @tomholub I removed this test because this test doesn't seem to be valid now.

tomholub: You should instead change it, to test for the expectation that the key will get removed

ioanmo226: I already implemented it in https://github.com/FlowCrypt/flowcrypt-ios/blob/86d9c10f64129f3bd292aa3e8b0dd23448b7e4d7/appium/tests/specs/mock/setup/CheckRefreshKeyFromEkmKeyDeletionAndRevokedKey.spec.ts#L54 :)

tomholub (approved):

PR #1653, opened 2022-05-26
#1163 Add ui test for wrong error format

### LINKED ISSUE #1163: write appium test that returns FES mock error with wrong error format
currently the app gets confused

should still show that it was a 400 from FES url

sosnovsky:
This PR adds ui test for handling FES errors with wrong format.

close #1163

- Tests added or updated

### To be filled by reviewers


FlowCryptCommon/Extensions/ErrorExtensions.swift
```diff
@@ -14,7 +14,9 @@ public extension Error {
         case let self as CustomStringConvertible:
             return String(describing: self)
         default:
-            return localizedDescription
+            return localizedDescription.trimmingCharacters(
+                in: CharacterSet(charactersIn: "\"") // remove quotes from error message
+            )
```
tomholub, line 19: I imagine the quotes could sometimes be meaningful in error messages, like quoting some wrong input. Which may eg be empty and without quotes, that would be difficult to describe

appium/api-mocks/apis/fes/fes-endpoints.ts
```diff
@@ -70,7 +76,11 @@ export const getMockFesEndpoints = (
 
 const throwErrorIfConfigSaysSo = (config: FesConfig) => {
   if (config.returnError) {
-    throw new FesHttpErr(config.returnError.message, config.returnError.code);
+    if (config.returnError.format == "wrong-text") {
+      throw new FesPlainHttpErr(config.returnError.message, config.returnError.code);
+    } else {
+      throw new FesHttpErr(config.returnError.message, config.returnError.code);
+    }
```
tomholub, line 83: Here please make a if-else that == `wrong-json` and it would return a json of a wrong format for example `{"wrongFieldError": {"wrongFieldCode": 400, "wrongFieldMessage": "wrong format"}}`

That will be returned if you put `returnError: {  code: 400, message: "wrong format", format: "wrong-json" }` in config.

tomholub: Then write one more test (or extend the newly added test - eg you can change config and restart app) that also tests the wrong-json format.

tomholub (commented):
Looks good - there is also one more thing to test

FlowCryptCommon/Extensions/ErrorExtensions.swift
sosnovsky: It removes quotes only in the beginning and end of the string, so `"some client err"` becomes `some client err`.
I added this trimming because quotes were automatically added to the error message from `wrong-text` test.

There can be an issue if we have quotes only on 1 side, like `wrong format: "some text"`, it'll be transformed into `wrong format: "some text`.

I can leave as it was before - without trimming, or improve current logic to better handle such cases, what do you prefer?

tomholub: I think we can do without trimming. Thanks!

tomholub (approved):

PR #1654, opened 2022-05-27
#1099 Honor reply-to header

### LINKED ISSUE #1099: reply-to header not honored
on `flowcrypt.compatibility` account there is a test message `Honor reply-to address - encrypted`

When you open on iOS and hit reply, you will notice the `to:` shows `flowcrypt.compatibility` when it should show another email per the reply-to header. Compare with browser extension.

Copy the message to e2e account to also write a ui test for this, which will tap on the reply button and check the recipient field.

ioanmo226:
This PR honors reply-to header

close #1099 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


sosnovsky (approved):
Looks good 👍

PR #512, opened 2021-09-21
Issue #506 Bring in flowcrypt-mobile-core

IvanPizhenko:
This PR brings in contents of the https://github.com/FlowCrypt/flowcrypt-mobile-core into this repo.

close #506

----------------------------------

**Tests** _(delete all except exactly one)_:
- Tests added or updated

--------------------------------

### To be filled by reviewers

I have reviewed that this PR... _(tick whichever items you personally focused on during this review)_:
- [x] addresses the issue it closes (if any)
- [x] code is readable and understandable
- [x] is accompanied with tests, or tests are not needed
- [x] is free of vulnerabilities
- [ ] is documented clearly and usefully, or doesn't need documentation

Issue #506, opened 2021-09-21
bring flowcrypt-mobile-core repo contents into flowcrypt-ios repo


tomholub:
There is a repository https://github.com/FlowCrypt/flowcrypt-mobile-core where we had OpenPGP.js based tooling (called "Core") for both Android and iOS.

On Android we already transitioned to PGPainless based implementation, so that repository is now only used by iOS.

I'd like to bring the contents into this repository, to make it easier to maintain going forward. 

1) I propose a folder called `Core` ? Then inside of that, there would be most things you see here: https://github.com/FlowCrypt/flowcrypt-mobile-core
2) the tests that run here https://github.com/FlowCrypt/flowcrypt-mobile-core/blob/master/.semaphore/semaphore.yml should be added to semaphore file on this iOS repository
3) you don't have to update other `scripts` in package.json - I need to clean these up myself
4) the result of the build does not need to be copied into the iOS Swift project (to use the built file, for now we'll have to manually copy it from the built result into https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/Resources/flowcrypt-ios-prod.js.txt

tomholub:
After we bring it here, we'll also remove Android-specific parts of the code. We'll do that in a separate step, to make the transition cleaner.

PR #1664, opened 2022-05-31
#1533 Commit hook and lint staged

### LINKED ISSUE #1533: extend CI linters to appium directory + discuss commit hook
The way we solved this on other repos was with linters that run in CI. We did that here too, for TypeScript, but I suspect we didn't extend it to the /appium directory which we should.

With a commit hook, I'm worried it may become another thing that may break and need to be maintained, but maybe it's just my lack of experience. We could consider that for later, if it works well and if Roma agrees.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1526#issuecomment-1111169622_

ioanmo226:
This PR implemented commit hook and lint staged

close #1533 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -10,6 +10,7 @@ auto_cancel:
   running:
     when: branch != 'master'
 blocks:
+
   - name: Appium UI tests
     dependencies: []
     run:
@@ -28,7 +29,7 @@ blocks:
         commands:
           - checkout && cd ~/git/flowcrypt-ios/ && cache restore && make dependencies && cache store
           - mv ~/appium-env ~/git/flowcrypt-ios/appium/.env
-          - sem-version node 16 && cd ./appium && npm i && cd .. && cache store appium-npm appium/node_modules
+          - sem-version node 16 && cache restore appium-npm && cd ./appium && npm i && cd .. && cache store appium-npm appium/node_modules
       jobs:
         - name: Appium UI tests
           commands:
@@ -48,6 +49,7 @@ blocks:
           commands:
             - artifact push job ~/Library/Logs/DiagnosticReports/FlowCrypt_*
             - artifact push job ~/git/flowcrypt-ios/appium/video/*.mp4
+
   - name: TypeScript + Swift tests
     dependencies: []
     run:
@@ -64,7 +66,7 @@ blocks:
         commands:
           - checkout && cd ~/git/flowcrypt-ios/ && cache restore && make dependencies && cache store
           - sem-version node 16 && mkdir .custom-npm && cd .custom-npm && npm install npm@8.1.2 && cd ..
-          - ( cd Core && ../.custom-npm/node_modules/.bin/npm install && cache store core-npm node_modules )
+          - ( cd Core && cache restore core-npm && ../.custom-npm/node_modules/.bin/npm install && cache store core-npm node_modules )
       jobs:
         - name: TypeScript tests + code quality
           commands:
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -65,7 +65,7 @@ blocks:
       prologue:
         commands:
           - checkout && cd ~/git/flowcrypt-ios/ && cache restore && make dependencies && cache store
-          - sem-version node 16 && mkdir .custom-npm && cd .custom-npm && npm install npm@8.1.2 && cd ..
+          - 'wget https://github.com/FlowCrypt/flowcrypt-ios/files/8811545/custom-npm.zip -P ~/git/flowcrypt-ios/appium'
           - ( cd Core && cache restore core-npm && ../.custom-npm/node_modules/.bin/npm install && cache store core-npm node_modules )
       jobs:
         - name: TypeScript tests + code quality
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -75,8 +75,7 @@ blocks:
             - npm test
             - npm run tslint
             - npm run eslint
-            - cd ..
-            - cd appium
+            - cd ../appium
             - npm run lint
         - name: Swift tests
           commands:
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -0,0 +1,35 @@
...
+  "homepage": "https://github.com/FlowCrypt/flowcrypt-ios#readme",
+  "devDependencies": {
+    "husky": "^8.0.1",
+    "lint-staged": "^12.4.3"
+  },
+  "lint-staged": {
+    "Core/*.{js,ts}": [
+      "cd Core && npm test",
+      "cd Core && npm run tslint",
+      "cd Core && npm run eslint"
```
ioanmo226, line 28: It's not good idea to run `npm test` for every commits.
Here are options which I can suggest.

1. Do not run commit hook for `Core` and keep `Typescript tests` in semaphoreci.
2. Run `npm run tslint` and `npm run eslint` for staged files for every commits & run `npm test` for every pushes. And delete `Typescript tests` in semaphoreci. (I prefer this one)
3. Run `npm test`, `npm run tslint`, `npm run eslint` for every commits. (This seems bad idea)

Open to discussion @tomholub, @sosnovsky

package.json
```diff
@@ -0,0 +1,34 @@
...
+    ],
+    "appium/*.{js,ts}": [
+      "cd appium && npm run eslint"
```
ioanmo226, line 31: We can run `eslint --fix` before run `npm run eslint` for staged files to auto-fix eslint errors.
Do you want me to run `eslint --fix` before `npm run eslint`?

package.json
tomholub: > Run npm run tslint and npm run eslint for staged files for every commits & run npm test for every pushes. And delete Typescript tests in semaphoreci. (I prefer this one)

I guess this but why delete tests in semaphoreci? Let them run there too, it won't hurt.

package.json
tomholub: on a precommit hook, sure

on semaphore ci, run linters without any attempts to fix, else they would mask problems

ioanmo226: ok

package.json
ioanmo226: It won't hurt but it won't be necessary neither.

ioanmo226: All changes in git will be checked before every commits and Typescript tests on semaphoreci won't be necessary

ioanmo226:
[custom-npm.zip](https://github.com/FlowCrypt/flowcrypt-ios/files/8811545/custom-npm.zip)

.semaphore/semaphore.yml
```diff
@@ -65,15 +65,20 @@ blocks:
       prologue:
         commands:
           - checkout && cd ~/git/flowcrypt-ios/ && cache restore && make dependencies && cache store
-          - sem-version node 16 && mkdir .custom-npm && cd .custom-npm && npm install npm@8.1.2 && cd ..
+          - 'wget https://github.com/FlowCrypt/flowcrypt-ios/files/8811587/custom-npm.zip -P ~/git/flowcrypt-ios'
```
ioanmo226, line 68: Not sure why we used custom-npm (v8.1.2) before.
Anyway, now root project contains pacakage.json and previous method of `npm install npm@8.1.2` won't work (need to do npm init first).
To achieve this, I had to upload & use custom-npm folder which contains npm v8.1.2

tomholub: Use git blame to understand historical context. It was introduced here

https://github.com/FlowCrypt/flowcrypt-ios/blame/14b56af9f738f1b21e6118f3fc424fc461164908/.semaphore/semaphore.yml

https://github.com/FlowCrypt/flowcrypt-ios/issues/506

https://github.com/FlowCrypt/flowcrypt-ios/pull/512

Looks to be there for building the Core folder. It may have only built properly on some npm versions and not others.

image


Why not try the latest npm (a specific version) and since you are using it for everything, why not update the global one? The zipped version looks even more hacky than the previous setup.

tomholub:
Using global is good if it works but we still want to pin it to a particular version. Eventually they will upgrade / change the global and it may break the build without any code change on our end, which is annoying to investigate.

.semaphore/semaphore.yml
```diff
@@ -65,15 +65,19 @@ blocks:
       prologue:
         commands:
           - checkout && cd ~/git/flowcrypt-ios/ && cache restore && make dependencies && cache store
-          - sem-version node 16 && mkdir .custom-npm && cd .custom-npm && npm install npm@8.1.2 && cd ..
-          - ( cd Core && cache restore core-npm && ../.custom-npm/node_modules/.bin/npm install && cache store core-npm node_modules )
+          - npm install -g npm@8.11.0
```
sosnovsky, line 68: Why do we use such old node version here? I tried to build `Core` locally with `node@16` - works well, without any errors

sosnovsky: oops, misread it 🙂
then all good

sosnovsky:
@ioanmo226 to make it work I just need to run `npm install` in `Flowcrypt` directory or other setup steps needed?

ioanmo226:
> @ioanmo226 to make it work I just need to run `npm install` in `Flowcrypt` directory or other setup steps needed?

Just npm install and try to commit changes when you want to.

.semaphore/semaphore.yml
ioanmo226: 8.11.0 is latest npm version

.semaphore/semaphore.yml
```diff
@@ -65,15 +65,19 @@ blocks:
...
           commands:
             - cd Core
-            - ../.custom-npm/node_modules/.bin/npm test
-            - ../.custom-npm/node_modules/.bin/npm run tslint
-            - ../.custom-npm/node_modules/.bin/npm run eslint
+            - npm test
+            - npm run tslint
+            - npm run eslint
+            - cd ..
+            - cd appium
```
tomholub, line 79: can be one command as `cd ../appium` which will actually run faster on semaphore, especially when their servers are congested. There is overhead for each semaphore command

sosnovsky:
> Just npm install and try to commit changes when you want to.

Got it, then let's add `npm install` command to `Installation` section of `README.md`

sosnovsky (approved):

PR #1671, opened 2022-06-03
#1125 Replace try?, try! with try

### LINKED ISSUE #1125: avoid `try?` and `try!` usages, instead throw

ioanmo226:
This PR replaced try?, try! with try

close #1125 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Inbox/InboxViewController.swift
+++ b/FlowCrypt/Controllers/Inbox/InboxViewController.swift
@@ -370,13 +370,7 @@ extension InboxViewController {
 
         switch error {
         case GmailServiceError.invalidGrant:
-            do {
-                try appContext.globalRouter.renderMissingPermissionsView(
-                    appContext: appContext
-                )
-            } catch {
-                showAlert(message: error.errorMessage)
-            }
+            appContext.globalRouter.renderMissingPermissionsView(appContext: appContext)
         default:
             let appError = AppErr(error)
             switch appError {
--- a/FlowCrypt/Controllers/Inbox/InboxViewController.swift
+++ b/FlowCrypt/Controllers/Inbox/InboxViewController.swift
@@ -370,9 +370,13 @@ extension InboxViewController {
 
         switch error {
         case GmailServiceError.invalidGrant:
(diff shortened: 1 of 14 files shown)


FlowCrypt/Controllers/Inbox/InboxViewController.swift
```diff
@@ -370,9 +370,13 @@ extension InboxViewController {
...
         switch error {
         case GmailServiceError.invalidGrant:
-            appContext.globalRouter.renderMissingPermissionsView(
-                appContext: appContext
-            )
+            do {
+                try appContext.globalRouter.renderMissingPermissionsView(
+                    appContext: appContext
+                )
+            } catch {
```
sosnovsky, line 377: We can move this `do/catch` block inside `renderMissingPermissionsView` to remove code duplication. So it'll look like
```swift
do {
    let email = try appContext.encryptedStorage.activeUser?.email ?? ""
   ...
} catch {
    keyWindow.rootViewController?.showAlert(....
}
```

FlowCrypt/Models/Realm Models/RecipientRealmObject.swift
```diff
@@ -22,23 +22,23 @@ extension RecipientRealmObject {
...
+    ) throws {
         self.init()
         self.email = email
         self.name = name ?? ""
         self.lastUsed = lastUsed
 
-        let realmKeys = keys
-            .compactMap { try? PubKeyRealmObject($0) }
+        let realmKeys = try keys
+            .compactMap { try PubKeyRealmObject($0) }
```
sosnovsky, line 32: In this case recipient saving won't succeed even when some keys are valid. Previous code filtered `realmKeys` to include only keys with valid `primaryFingerprint`.

I think it'll be better to replace `try?` with optional init, so `PubKeyRealmObject` will return `nil` when `fingerprints` are empty.

FlowCrypt/Controllers/Inbox/InboxViewController.swift
ioanmo226: Sounds good 👍

sosnovsky (approved):

PR #1682, opened 2022-06-10
#1003 make sure Attester lookup failure doesn't affect WKD in task group

### LINKED ISSUE #1003: make sure Attester lookup failure doesn't affect WKD in task group
As implemented now, if Attester call fails (server down), I think it would throw even if WKD call was successfully returning a key. For now that's not bothering me since primary customer has Attester disabled using the `disallow_attester_search_for_domains=[*]` rule. But it's something that should be addressed later.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1002#issuecomment-966950340_

ioanmo226:
This PR makes sure Attester lookup failure doesn't affect WKD in task group

close #1003 // if this PR closes an issue

- Difficult to test (Hard to reproduce)

### To be filled by reviewers

### DIFF
--- a/appium/tests/specs/mock/setup/RespectsAllowAttesterSearchOnlyForDomains.spec.ts
+++ b/appium/tests/specs/mock/setup/RespectsAllowAttesterSearchOnlyForDomains.spec.ts
@@ -72,7 +72,7 @@ describe('SETUP: ', () => {
         }
       };
       await AppiumHelper.restartApp(processArgs);
-      await MailFolderScreen.clickCreateEmail();
+      await MailFolderScreen.clickCreateEmail(1000);
       await NewMessageScreen.setAddRecipient('empty-setting@enabled.test');
       // Checking added recipients color
       await NewMessageScreen.checkAddedRecipientColor('empty-setting@enabled.test', 0, 'gray');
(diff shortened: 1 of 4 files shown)


appium/tests/specs/mock/setup/RespectsAllowAttesterSearchOnlyForDomains.spec.ts
```diff
@@ -72,7 +72,7 @@ describe('SETUP: ', () => {
         }
       };
       await AppiumHelper.restartApp(processArgs);
-      await MailFolderScreen.clickCreateEmail();
+      await MailFolderScreen.clickCreateEmail(1000);
```
ioanmo226, line 75: @sosnovsky FYI, RespectsAllowAttesterSearchOnlyForDomains test failed in semaphoreci (i noticed this ui test failure 1 time before).
As I checked failure screenshot, MailFolderScreen was taken. That's why `aid-recipients-text-field-to` was not visibile.
To fix this, I increased delayMS.
So

sosnovsky (approved):
Looks good 👍

ioanmo226:
@sosnovsky Above fix try didn't work.
There seems to be issue/problem with Appium restart.
FYI I split that failure test into separate test.

sosnovsky (approved):

sosnovsky:
> There really seems to be issue/problem with Appium restart. Seems like no user interactions were enabled for 4th stage after Appium restart.

Yeah, something is wrong with Appium restarts on Semaphore, as locally they usually work well.
So it's better to split such complex tests into simpler ones.

Issue #1265, opened 2021-12-30
FES .upload `from` and `current account` should be passed separately for sendAs support


tomholub:
This is a subtle thing: `from` and `current account` may not be the same thing in the future - once we add support for different `sendAs` aliases, in which case you may send an email from `alias@example.com` while on account `account@example.com`. So this would really need an `account` and `from` separately, or in the future maybe `idToken` and `from` separately. For now though, it's ok. I'll file an issue.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1261#discussion_r776876297_

PR #1689, opened 2022-06-14
#1298 Send as alias

### LINKED ISSUE #1298: sendAs support of alias emails
every time the app is started, we should fetch https://developers.google.com/gmail/api/reference/rest/v1/users.settings.sendAs/list in the background and save list of sending emails for that account in Realm.

Then during compose, if the user has more than one sending address, we can give them choice which email to send from.

See settings at https://mail.google.com/mail/u/0/#settings/accounts where you can add a secondary sending email, which will then show up when that API is loaded.

You can try Gmail app, FlowCrypt Android app or FlowCrypt browser extension for existing implementations.

ioanmo226:
This PR implemented send as alias

close #1298 // if this PR closes an issue

- Tests added

### To be filled by reviewers


FlowCrypt/Controllers/Inbox/Container/InboxViewContainerController.swift
```diff
@@ -60,9 +62,16 @@ final class InboxViewContainerController: TableNodeViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
+        fetchSendAsList()
         fetchInboxFolder()
     }
 
+    private func fetchSendAsList() {
+        Task {
+            try await sendAsService.fetchList(isForceReload: true, for: appContext.user)
```
ioanmo226, line 71: @tomholub @sosnovsky, In which way do we have to display send as alias dropdown? As you know there's no such default dropdown feature.
Here are 3 options.

1. Use custom dropdown library such as this [one](https://github.com/AssistoLab/DropDown)
2. Make custom dropdown library.
3. Use iOS popover (Like recipient info popup when you tap recipient)
4. Use iOS action sheet (Like when you tap 3 dots in thread detail view.

Let me know which one do you prefer.

This is android screenshot
image

sosnovsky: I prefer option 4 - iOS action sheet, it's more familiar for iOS users and easy to implement, without need for custom elements and libraries.

ioanmo226: Sounds good. How about you @tomholub?

ioanmo226:
Regarding email encryption and sign, what private keys & public keys do i have to use?
Let's say I logged with `a@domain.test` and have `b@domain.test` as alias.
If I want to send email as `b@domain.test`, what private key & public key do i have to use for encryption & sign(I think device won't have keys for `b@domain.test`, Device will only have `a@domain.test` key info I think ) ?

FlowCrypt/Controllers/Inbox/Container/InboxViewContainerController.swift
ioanmo226: Gmail app displays actions sheet like @sosnovsky suggested.

ioanmo226:
@tomholub Please check above question

FlowCrypt/Controllers/Inbox/Container/InboxViewContainerController.swift
tomholub: Just tested gmail as well. Looked good to me.

tomholub:
For public keys (for encryption), `chooseSenderEncryptionKeys` already seems to have the functionality implemented, so:
 - continue to give it list of keys for `a@example.test` as before
 - `senderEmail` should be set to `b@example.test`

Then the code will do its work.

ioanmo226:
You mean configure an alias on the e2e user and ui test with that?

> You can configure an alias on the

ioanmo226:
Let me confirm.
To send email as alias, we don't only need `send as` email verified status in gmail but logged in account key details must have alias email user info?
I think most (90%) accounts (keys) won't have keys for aliases.

Please confirm

> check if filteredByUid is empty. If it is, throw Cannot sign message: none of your \(accountKeys.count) account keys can be used for sending address XXX so that an alert would be shown to the user.

tomholub:
> To send email as alias, we don't only need send as email verified status in gmail

Referring to https://developers.google.com/gmail/api/reference/rest/v1/users.settings.sendAs#SendAs

Since they have `verificationStatus`, it's reasonable to filter by it, to only choose `accepted`. We didn't do it on browser extension, and I'm not sure what we do on Android, but since you mention it, it makes sense to filter.

That means only verified sendAs addresses will be offered to user.

> but logged in account key details must also contain alias email user info?

For the sending of a signed message to be successful (and for the verification to later work), the message needs to be signed with some key that has the sender's email (as recorded in the from: header) encoded as one of its OpenPGP User IDs.

> I think most (90%) accounts (keys) won't have keys for aliases. Like my account.

Fair. We'll improve that, in the future, in the following ways:

 - (consumer) check sendAs during setup (key generation flow) and encode all sendAs aliases that were present during setup into the generated key. Browser extension does this.
 - (consumer) give user options on how to resolve the problem. The user could choose a) the app will edit their key to add the missing alias b) the app will generate a new key for that alias c) user will import a key that they may already have.
 - (enterprise) point the user to their admins, who can add a new key for that user on EKM to resolve the issue

> I added alias email in my flowcrypt gmail but my flowcrypt gmail account key doesn't contain alias email info.

That is expected, if it's a brand new alias that you've never used before. Eventually we'll also implement ways to help users resolve that scenario.

In your case, if that is a consumer account, you could generate a key outside of the app and import it in Settings -> Keys.

If it's an enterprise account, you can write me an email about which account and which new alias, and I can add a key to EKM for that alias, on your account.

ioanmo226:
Got it, But FYI in browser extension it doesn't check keys for aliases and I still can send email with alias.
It doesn't return any error. It's displaying like this (from recipient side) though.
 
image


> For the sending of a signed message to be successful (and for the verification to later work), the message needs to be signed with some key that has the sender's email (as recorded in the from: header) encoded as one of its OpenPGP User IDs.

tomholub:
Given that alias setup got more complicated then it used to be, we could merge this without automated tests (please test manually) and just based on the code review / manual tests.

After #1692 you will be able to make another PR where you could add a way to set gmail config:

```ts
apiMock.gmailConfig = {
  accounts: {
    'e2e...': {
       aliases: ['e2e...', 'alias@domain.test']
    }
  }
}
```

With that, you can design an automated test as well at that point.

If #1692 is ready before this merges, then you could do it in one PR.

tomholub:
But I guess you'll still need to test it manually. Ok, I'll have to set it up on e2e after all.

ioanmo226:
ok

tomholub:
done

image

image

image

ioanmo226:
Finished implementing this feature. I'm gonna implement ui test in separate issue (#1696).
By the way, Please check logic carefully. (i try my best to implement correctly but something might be wrong)

tomholub:
I suppose, you may be able to implement a ui test already here using live tests, given that e2e account has the alias now and the right keys associated.

Converting that to mock could be done after, and should be easy.

ioanmo226:
OK

FlowCrypt/Functionality/Pgp/KeyMethods.swift
```diff
@@ -68,7 +61,7 @@ final class KeyMethods: KeyMethodsType {
         if let byAnyUid = try filter(keys, usable, ({ $0.pgpUserEmailsLowercased.contains(senderEmail) })) {
             return byAnyUid // if any keys match by any uid, use them
         }
-        return try filter(keys, usable) ?? [] // use any usable keys
+        return []
```
tomholub, line 64: Choosing keys for signing and for encryption should work differently.

For encryption, this should be left as it was before.

For signing, need to follow the steps I explained, which this doesn't.

ioanmo226: You mean we should return any usable key even if we can't find keys for filtered condition?(which means logged in user's key contains senderEmail info)
> For encryption, this should be left as it was before.

What's different between you mentioned and current implementation?
> For signing, need to follow the steps I explained, which this doesn't.

tomholub: > You mean we should return any usable key even if we can't find keys for filtered condition?(which means logged in user's key contains senderEmail info)

I said encryption selection should be left unchanged except supplying the method with the senderEmail. But the following changed:

```diff
-        return try filter(keys, usable) ?? [] // use any usable keys
+        return []
```

Yes, for encryption, even when we cannot find key by the right uid, we can use the keys that don't match uid. It won't cause much trouble for encryption, but it does for signature verification, and that's why we're treating them differently. We can be more lenient for encryption.

> What's different between you mentioned and current implementation?

I've re-read the implementation for signing keys, and I think it will be ok. It's implemented differently, but I think it's the same. So just need to fix the selection for encryption, back to what it was.

------------------

You can include a variation of `// Yes, for encryption, even when we cannot find key by the right uid, we can use the keys that don't match uid. It won't cause much trouble for encryption, but it does for signature verification, and that's why we're treating them differently. We can be more lenient for encryption.` to make it clearer why we treat the two differently.

ioanmo226: Ok, got it

ioanmo226:
UI test is now added

appium/tests/specs/live/composeEmail/CheckSendAsAlias.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    await MailFolderScreen.clickCreateEmail();
+
+    await NewMessageScreen.composeEmail(recipientEmail, emailSubject, emailText);
+    await NewMessageScreen.checkRecipientLabel([recipientName]);
+    await NewMessageScreen.changeFromEmail(aliasEmail);
+
+    await NewMessageScreen.clickSendButton();
+    await MailFolderScreen.checkInboxScreen();
+
+    await MailFolderHelper.deleteSentEmail(emailSubject, emailText, aliasEmail);
```
tomholub, line 32: Until we can transfer this test to mock api, let's uncomment these three lines. To lower chances of messing up the state on email provider that we use for testing in the meantime.

tomholub (commented):
It's looking good (the key for signing / encryption selection, didn't review the rest).

When user has only one sendAs email, do we still show it on compose? I think it could be hidden in that scenario - at least web gmail does that (and browser extension, possibly android too).

tomholub:
> UI test is now added

Thank you. Sorry about the your code for signing key selection, it indeed did the same thing that I requested. I just got confused on it, because it looked different.

ioanmo226:
Currently, iOS app also indeed hides `from` field when there's only 1 sendAs email
> When user has only one sendAs email, do we still show it on compose? I think it could be hidden in that scenario

FlowCrypt/Functionality/Pgp/KeyMethods.swift
```diff
@@ -12,7 +12,12 @@ import Foundation
...
     func filterByPassPhraseMatch<T: ArmoredPrvWithIdentity>(keys: [T], passPhrase: String) async throws -> [T]
     func parseKeys(armored: [String]) async throws -> [KeyDetails]
-    func chooseSenderKeys(keys: [Keypair], senderEmail: String, forEncryptionFlag: Bool) async throws -> [Keypair]
+    func chooseSenderKeys(for type: ChooseSenderKeysType, keys: [Keypair], senderEmail: String) async throws -> [Keypair]
+}
+
+enum ChooseSenderKeysType {
+    case encryption
+    case signing
 }
```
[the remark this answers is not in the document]
ioanmo226: Yeah, it's not key type, I named it because it reprents type for `ChooseSenderKeys`.
Anyway, let me change as you suggested

tomholub:
There's a related issue https://github.com/FlowCrypt/flowcrypt-ios/issues/1265 which we could do in a follow-up.

tomholub:
@sosnovsky I've reviewed the key-related parts, please have a look at the PR overall. Thanks!

sosnovsky (commented):
Great job 👏

Just a small UI improvement - `From` is too close to sender email and it'll be nice to center them vertically, as currently sender is a bit lower than `From`:

Screenshot 2022-06-16 at 21 45 35

tomholub:
well noticed - also too much space be from & to. After adjusting it, please retake the above screenshot, and then one more screenshot with some recipient entered. Finally, a screenshot for an account that doesn't have alias.

ioanmo226:
With recipients entered.
image

Without recipients entered.
image

Without alias
image

image

sosnovsky (approved):
Looks good 👍

PR #1694, opened 2022-06-15
#1684 Add No result image + text in mail folder screen, added icons in side menu items

### LINKED ISSUE #1684: Add `No result` image + text in mail folder screen
Currently, when there are no emails/threads found, it displays nothing.
I think we should display `No result` image + text for no results.
Better if we display different `no result` images+texts for folder types like Gmail app does. (like for spam folder, we may display `Hooray no spam here` and no spam image)

ioanmo226:
This PR add no result image + text in mail folder screens, added icons in side menu items

close #1684 // if this PR closes an issue

- Not worth testing

### To be filled by reviewers


tomholub:
Please share screenshots of what the empty folders look like - thanks

FlowCrypt/Functionality/Services/Folders Services/RemoteFoldersProviderType/GmailService+folders.swift
```diff
@@ -79,7 +83,7 @@ private extension Folder {
         self.init(
             path: path,
             name: name,
-            image: nil
+            image: UIImage(systemName: path.mailFolderIcon)?.jpegData(compressionQuality: 1.0)
```
sosnovsky, line 86: I like how menu looks now with all these icons 👍

But I think we shouldn't store rendered iOS icons in database. Instead we can use `mailFolderIcon` if folder `image` field is empty:
```swift
image: object.image.flatMap(UIImage.init) ?? UIImage(systemName: object.name.mailFolderIcon),
```
(https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/Functionality/Services/Folders%20Services/Models/FolderViewModel.swift#L43)

@ioanmo226 what do you think?

ioanmo226: I absolutely agree. Thanks for your notice

sosnovsky (approved):
Nice 👍

FlowCrypt/Functionality/Services/Folders Services/RemoteFoldersProviderType/GmailService+folders.swift
ioanmo226: After Folder model is initialized, it will be saved into Realm (FolderRealmObject) database so that we can use it( without getting folder list again)  after initial fetch

tomholub: I see.

PR #1720, opened 2022-06-23
#1719 feat: added isHeadless true in CI mode

### LINKED ISSUE #1719: Set `isHeadless` to `true` in appium only in semaphoreci
I think it might speed up CI speed(Maybe 1.5x-2x) if we set `isHeadless` to `true` only in semaphoreci. (I think we still need to see ui results in local as that's useful for debugging).
Let me know your thoughts.

https://appium.io/docs/en/writing-running-appium/caps/#:~:text=true%20or%20false-,isHeadless,-Set%20this%20capability

image

ioanmo226:
This PR added isHeadless true in CI mode

close #1719 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/appium/config/wdio.live.conf.js
+++ b/appium/config/wdio.live.conf.js
@@ -37,7 +37,8 @@ config.capabilities = [
     wdaLaunchTimeout: 300000,
     wdaConnectionTimeout: 600000,
     wdaStartupRetries: 4,
-    wdaStartupRetryInterval: 120000
+    wdaStartupRetryInterval: 120000,
+    isHeadless: process.env.CI ? true : false
   },
 ];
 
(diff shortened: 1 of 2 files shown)


ioanmo226:
@sosnovsky `isHeadless: true` doesn't seem to speed up in semaphoreci.
In local, I saw good speed improvement and in semaphoreci, it takes same time as before(or a bit more).

Do you have any idea why?

sosnovsky:
@ioanmo226 maybe semaphore time is the same because many tests failed? As each test was retried, so it took 2x time

ioanmo226:
and also I wonder why all live tests and 2 mock test fail even though there are no code changes.
i tried on local and they failed too.

sosnovsky:
@ioanmo226 looks like something wrong with test account backups:

sosnovsky:
> Yes, you are right. FES returns nil ClientConfiguration. Do you know who is responsible for this?
> 
> image

Probably @tomholub should know how this issue can be fixed.

sosnovsky:
@ioanmo226 yeah, you're right - the same behavior happens on my personal `@flowcrypt.com` account too, so something wrong on FES side.

PR #1727, opened 2022-06-29
#1556 Handle situation when refreshing keys from EKM and have no local keys

### LINKED ISSUE #1556: handle situation when refreshing keys from EKM and have no local keys
I'll file another issue about this for the future. If no keys are present, then any pass phrase would be fine as long as it's strong enough. But then we need to also ask for the pass phrase second time to confirm it, like we do at initial setup, and we need to make sure the pass phrase is strong enough (same thing as setup).

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1510#discussion_r861973611_

ioanmo226:
This PR handled situation when refreshing keys from EKM and have no local keys

close #1556 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


sosnovsky (changes requested):

sosnovsky:
Strange, system icon should look sharp. Maybe it scales to fit cell size and this causes blurness?

sosnovsky:
> It looks blurry because I added opacity as 0.5?. Here's one without opacity

Probably know, I thought this line cause blur - https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-1556/FlowCryptUI/Cell%20Nodes/EmptyCellNode.swift#L41, but it's still blurred after I commented it.

Usually to change size of system icon such code is used, but it makes image blurry for `EmptyCellNode`.
```swift
let configuration = UIImage.SymbolConfiguration(pointSize: 100)
imageNode.image = UIImage(systemName: input.imageName ?? "", withConfiguration: configuration)
```

sosnovsky (approved):
Great! Looks nice now 👍

PR #1745, opened 2022-07-13
#1693 Migrate tests to mock APIs

### LINKED ISSUE #1693: bring all except one live gmail tests to mock gmail
after #1692 

we will leave one live gmail test there - something simple like "log in and read some message". All other tests will be mocked.

### LINKED ISSUE #1704: remove send message strategies - instead retrieve sent message from mockApi
We can remove `flowcrypt-ios/appium/api-mocks/apis/google/strategies` and instead test our expectations for sent email directly in the test spec, like this:

```ts
  it('user is able to reply to email', async () => {
    const mockApi = new MockApi();
    mockApi.fesConfig = { ... };
    mockApi.ekmConfig = {
      returnKeys: [ekmKeySamples.e2eValidKey.prv]
    }
    mockApi.googleConfig = {
      accounts: {
        'e2e.enterprise.test@flowcrypt.com': {
          messages: ['CC and BCC test'],
        }
      }
    }
    await mockApi.withMockedApis(async () => {
      await SplashScreen.mockLogin();
      await SetupKeyScreen.setPassPhrase();
      await MailFolderScreen.checkInboxScreen();
      await MailFolderScreen.clickOnEmailBySubject(emailSubject);

      await EmailScreen.checkOpenedEmail(senderName, emailSubject, emailText);
      await EmailScreen.clickReplyAll();
      await ComposeScreen.fillMsg(...);
      await ComposeScreen.clickSend();

      // now test the sent message - this replaces all of the "test strategy" classes
      expect(mockApi.google.sentMessages.length).to.equal(1);
      expect(mockApi.google.sentMessages[0].parsed.subject).to.equal(...);
      expect(mockApi.google.sentMessages[0].parsed.attachments.length).to.equal(0);
      const decryptedBody = decrypt(mockApi.google.sentMessages[0].parsed.body, ekmKeySamples.e2eValidKey.prv);
      expect(decryptedBody).to.equal(...);
      // ... end so on
    });
  });
});

### PR DESCRIPTION
**Pull request by sosnovsky**
This PR migrates all tests to mock (except `user should be able to check encrypted email, contacts after update app from old version`)

close #1693 
close #1704

- Tests added or updated

### To be filled by reviewers

### CODE REVIEW

File: appium/tests/specs/mock/setup/CheckRefreshKeyFromEkmKeyDeletionAndRevokedKey.spec.ts [hunk 1]
```diff
@@ -13,10 +13,16 @@ describe('SETUP: ', () => {
 
   it('will not update a revoked private key with valid one and delete local key if it\'s removed from EKM and not revoked one', async () => {
 
-    const mockApi = MockApi.e2eMock;
```
tomholub: Could call it `mockWithDefaultEnterpriseConfig`

sosnovsky: Ok, will rename it

FlowCrypt/Functionality/Mail Provider/Threads/MessagesThreadOperationsProvider.swift
```diff
@@ -41,7 +41,7 @@ extension GmailService: MessagesThreadOperationsProvider {
     }
 
     func moveThreadToTrash(thread: MessageThread) async throws {
-        try await update(thread: thread, labelsToAdd: [.trash], labelsToRemove: [.inbox])
+        try await update(thread: thread, labelsToAdd: [.trash], labelsToRemove: [.inbox, .sent])
     }
```
tomholub, line 45: I suppose this mimics Gmail behavior?

appium/api-mocks/apis/google/google-messages.ts
```diff
@@ -1,4 +1,4 @@
 /* ©️ 2016 - present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com */
 
-export type GoogleMockAccountEmail = 'e2e.enterprise.test@flowcrypt.com' | 'e2e.consumer.tests@outlook.com';
-export type GoogleMockMessage = 'CC and BCC test' | 'Your FlowCrypt Backup';
+export type GoogleMockAccountEmail = 'e2e.enterprise.test@flowcrypt.com' | 'flowcrypt.compatibility@gmail.com';
+export type GoogleMockMessage = 'CC and BCC test' | 'Test 1' | 'Signed and encrypted message' | 'Honor reply-to address - plain' | 'email with text attachment' | 'Message with cc and multiple recipients and text attachment' | 'new message for reply' | 'Signed only message' | 'Signed only message with detached signature' | 'Signed only message where the pubkey is not available' | 'Signed only message that was tempered during transit' | 'Partially signed only message' | 'encrypted - MDC hash mismatch - modification detected - should fail' | 'message encrypted for another public key (only one pubkey used)' | 'wrong checksum' | 'not integrity protected - should show a warning and not decrypt automatically' | 'key mismatch unexpectedly produces a modal' | 'test thread rendering' | 'Archived thread';
```
tomholub, line 4: Let's add some newlines :-) for each line to have reasonable length.

appium/api-mocks/mock.ts
```diff
@@ -72,5 +73,51 @@ export class MockApi {
     }
   };
 
+  static get e2eMock() {
```
tomholub, line 76: Certain things will repeat a lot, and it will make sense to have a default for them. For example, the FES and EKM config.

Other things are specific for each test, and should be defined for each test. For example, google config and attester config.

I have very bad experience trying to define the state globally for all tests in one place. It gets confusing as to which test is using which parts of the config. Changing the config by adding one more message or one more contact may break another test, and so on.

Therefore we should do something like the following, for each test definition:

```ts
    const mockApi = new MockApi();
    mockApi.fesConfig = MockApiConfig.defaultEnterpriseFesConfiguration;
    mockApi.ekmConfig = MockApiConfig.defaultEnterpriseEkmConfiguration;
    mockApi.googleConfig = {
      accounts: {
        'e2e.enterprise.test@flowcrypt.com': {
          contacts: [CommonData.contacts.someParticularContact],
          messages: ['CC and BCC test'],
        }
      }
    };
    mockApi.attesterConfig = { 
      servedPubkeys: { 'expired@flowcrypt.com': attesterPublicKeySamples.expiredFlowcrypt }
    };
    mockApi.wkdConfig = {}
    await mockApi.withMockedApis(async () => {
      // test goes here
    });
```

Sure, it will be wordier. But it will be an absolute bliss to write and maintain (when you return to a particular test 4 years later) because you can look at the top of the test and know exactly how it is configured and why.

If many tests are using the same config, and the config contains messages only used by some of the tests, and contacts or public keys only used by some of the tests.. it becomes harder to dis-entangle it when understanding a particular test. And it also becomes less clear what to do when writing a new test - should I update the global configuration and risk another test will break? Or should I copy everything just to adjust one thing for a new test?

What I'm proposing is a system where there is just one way to do a particular thing: each test has its own minimal precise state definition, even if it means more lines of code overall.

appium/api-mocks/apis/google/google-messages.ts
sosnovsky: Agree, looks bad for me too :)

appium/api-mocks/mock.ts
tomholub: 👍

tomholub: I'm ok to do it in a separate PR, if you'd like this one merged soon. Or it can be same PR - up to you.

sosnovsky: I'll do it in this PR, shouldn't take too much time.

appium/api-mocks/apis/wkd/wkd-endpoints.ts
```diff
@@ -30,136 +26,15 @@ export const getMockWkdEndpoints = (
...
+    '/.well-known/openpgpkey/hu/?': async () => {
+      return { "code": 404, "message": "Public key not found", "details": "" }
+    },
+    '/.well-known/openpgpkey/127.0.0.1:8001/policy': async () => {
+      return ''; // allow advanced for localhost
+    },
+    '/.well-known/openpgpkey/policy': async () => {
+      return ''; // allow direct for all
+    },
+  };
```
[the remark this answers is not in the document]
sosnovsky: Created https://github.com/FlowCrypt/flowcrypt-ios/issues/1747

appium/api-mocks/mock.ts
ioanmo226: How about we implement this way?

interface MockApiOptions {
  fesConfig?: FesConfig;
  googleConfig?: GoogleConfig;
  wkdConfig?: WkdConfig;
  ekmConfig?: EkmConfig;
  attesterConfig?: FesConfig;
}
  constructor(options: MockApiOptions) {
    this.fesConfig = options.fesConfig || MockApiConfig.defaultEnterpriseFesConfiguration;
    this.googleConfig = options.googleConfig; // here to use default google config if needed like above like ( || MockApiConfig.defaultGoogleConfig)
    this.wkdConfig = options.wkdConfig || {};
    this.ekmConfig = options.ekmConfig || MockApiConfig.defaultEnterpriseEkmConfiguration;
    this.attesterConfig = options.attesterConfig || {};
  }

ioanmo226: How about we implement this way?
This way we would above duplicating multiple lines of code in many places.

```ts
mockApi.fesConfig = MockApiConfig.defaultEnterpriseFesConfiguration; // this one
    mockApi.ekmConfig = MockApiConfig.defaultEnterpriseEkmConfiguration;  // this one
    mockApi.googleConfig = {
      accounts: {
        'e2e.enterprise.test@flowcrypt.com': {
          contacts: [CommonData.contacts.someParticularContact],
          messages: ['CC and BCC test'],
        }
      }
    };
    mockApi.attesterConfig = { 
      servedPubkeys: { 'expired@flowcrypt.com': attesterPublicKeySamples.expiredFlowcrypt }
    };
    mockApi.wkdConfig = {}  // this one

```
```ts
interface MockApiOptions {
  fesConfig?: FesConfig;
  googleConfig?: GoogleConfig;
  wkdConfig?: WkdConfig;
  ekmConfig?: EkmConfig;
  attesterConfig?: FesConfig;
}
  constructor(options: MockApiOptions) {
    this.fesConfig = options.fesConfig || MockApiConfig.defaultEnterpriseFesConfiguration;
    this.googleConfig = options.googleConfig; // here to use default google config if needed like above like ( || MockApiConfig.defaultGoogleConfig)
    this.wkdConfig = options.wkdConfig || {};
    this.ekmConfig = options.ekmConfig || MockApiConfig.defaultEnterpriseEkmConfiguration;
    this.attesterConfig = options.attesterConfig || {};
  }

const mockApi = new MockApi({
  attesterConfig: {
    servedPubkeys: { 'expired@flowcrypt.com': attesterPublicKeySamples.expiredFlowcrypt }
  }
});

mockApi.fesConfig = something; // if you want to change defaultFesConfig by constructor you can change it below init
```

ioanmo226: But imagine having this kind of duplicated codes in tens of places.
Also I don't think it's good to change class variable (better to make variable private and have public function for getter and setter).
With above suggestion, we can pass all required parameters in constructor and won't have above issue either

tomholub: > But imagine having this kind of duplicated codes in tens of places.

yes, but tests are different than implementation code. In tests, more explicit is better, to understand what is happening. More code is usually worse, but not always.

Also, many tests will be non-enterprise tests without FES or EKM. Maybe not as many as enterorise tests, but still. I much prefer to have it explicit.

> Also I don't think it's good to change class variable directly [like mockApi.ekmConfig = xx, mockApi.xx = xx...](better to make variable private and have public function for getter and setter).

We could have setters. Like

```ts
    mockApi.setFesConfig(MockApiConfig.defaultEnterpriseFesConfiguration);
    mockApi.setEkmConfig(MockApiConfig.defaultEnterpriseEkmConfiguration);
    mockApi.setGoogleConfig({
      accounts: {
        'e2e.enterprise.test@flowcrypt.com': {
          contacts: [CommonData.contacts.someParticularContact],
          messages: ['CC and BCC test'],
        }
      }
    });
    mockApi.setAttesterConfig({ 
      servedPubkeys: { 'expired@flowcrypt.com': attesterPublicKeySamples.expiredFlowcrypt }
    });
```

I guess we could have a little more compact way to set up google account:

```ts
    mockApi.setFesConfig(MockApiConfig.defaultEnterpriseFesConfiguration);
    mockApi.setEkmConfig(MockApiConfig.defaultEnterpriseEkmConfiguration);
    mockApi.setGoogleConfigWithOneAccount('e2e.enterprise.test@flowcrypt.com', {
      messages: ['CC and BCC test'],
      contacts: [CommonData.contacts.someParticularContact],
    });
    mockApi.setAttesterConfig({ 
      servedPubkeys: { 'expired@flowcrypt.com': attesterPublicKeySamples.expiredFlowcrypt }
    });
```

That would save some lines while keeping it explicit.

ioanmo226: But I still prefer to pass variables in constructor though rather than passing it via `setter` function or setting public variable.

sosnovsky: Okay, then I'll make these config variables private and add custom setters.

> But I still prefer to pass variables in constructor though rather than passing it via `setter` function or setting public variable.

In some tests we need to change these configs in the middle of the test execution. That's why we need ability to set configs not only in constructor.

sosnovsky: Let's leave current implementation for now, as it works well for tests and code will be the same for initialization and for updating mock config.

tomholub: I also think setters work pretty well for the consistency, whether you are creating a new mock api or updating it, the code is the same. I don't think the individual setters increase burden.

tomholub (approved):
👍

PR #1759, opened 2022-07-20
#1652 vscode workspace update

### LINKED ISSUE #1652: Update recommended packages in VS workspace
There are some outdated packages:
- `philhindle.errorlens` wasn't updated since 2019, currently there is another actively developed `error lens` extension - https://marketplace.visualstudio.com/items?itemName=usernamehw.errorlens
- `ms-vscode.vscode-typescript-tslint-plugin` documentation says 'TSLint has been deprecated in favor of ESLint and this extension has also been deprecated in favor of ESLint's tooling. Please look into [migrating your projects to ESLint](https://github.com/typescript-eslint/typescript-eslint#typescript-eslint).'

sosnovsky:
This PR:
- updates recommended packages for vscode workspace (not used packages are removed, `tslint` replaced with `eslint` etc)
- workspace code is separated into 3 groups - `appium`, `core` and `ios`
- `eslint` npm packages are moved to the root `package.json`, so they're not duplicated in `appium/node_modules` and `Core/node_modules`

close #1652

### To be filled by reviewers


appium/api-mocks/core/msg-block-parser.ts
```diff
@@ -112,22 +64,24 @@ export class MsgBlockParser {
...
+  // }
 
   public static stripFcTeplyToken = (decryptedContent: string) => {
     return decryptedContent.replace(/<div[^>]+class="cryptup_reply"[^>]+><\/div>/, '');
   }
 
   private static isFcAttachmentLinkData = (o: any): o is FcAttachmentLinkData => {
-    return o && typeof o === 'object' && typeof (o as FcAttachmentLinkData).name !== 'undefined' // tslint:disable-line:no-unsafe-any
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+    return o && typeof o === 'object' && typeof (o as FcAttachmentLinkData).name !== 'undefined'
```
ioanmo226, line 84: How about we do this way?
```ts
  private static isFcAttachmentLinkData = (o: FcAttachmentLinkData): o is FcAttachmentLinkData => {
    return o && typeof o === 'object' && typeof (o as FcAttachmentLinkData).name !== 'undefined'
      && typeof (o as FcAttachmentLinkData).size !== 'undefined' && typeof (o as FcAttachmentLinkData).type !== 'undefined';
  }
```

appium/api-mocks/core/msg-block-parser.ts
```diff
@@ -101,7 +52,8 @@ export class MsgBlockParser {
   }
 
   public static stripPublicKeys = (decryptedContent: string, foundPublicKeys: string[]) => {
-    const { blocks, normalized } = MsgBlockParser.detectBlocks(decryptedContent);
+    // eslint-disable-next-line prefer-const
+    let { blocks, normalized } = MsgBlockParser.detectBlocks(decryptedContent);
```
ioanmo226, line 56: Maybe we can use like this? I think it's good to avoid `disable eslint rules` if possible
```ts
const { blocks, normalized } = MsgBlockParser.detectBlocks(decryptedContent);
let changedNormalized = normalized;
```

Core/source/test/test-utils.ts
```diff
@@ -58,33 +58,34 @@ export const expectData = (_data: Uint8Array, type?: 'armoredMsg' | 'msgBlocks'
     expect(dataStr).to.contain('-----BEGIN PGP MESSAGE-----');
     expect(dataStr).to.contain('-----END PGP MESSAGE-----');
   } else if (type === 'msgBlocks') {
-    // tslint:disable-next-line:no-unsafe-any
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
ioanmo226, line 61: do we actually need this `disable rule`?

Core/source/test.ts
```diff
@@ -1086,6 +1092,7 @@ ava.default('decryptErr for not integrity protected message', async t => {
       [await getCompatAsset('mime-email-not-integrity-protected')]));
   expect(decryptJson.replyType).equals('plain');
   expect(decryptJson.subject).equals('not integrity protected - should show a warning and not decrypt automatically');
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
ioanmo226, line 1095: this one also seems unnecessary

Core/source/core/mime.ts
```diff
@@ -400,9 +398,8 @@ export class Mime {
   };
 
   private static getNodeContentAsUtfStr = (node: MimeParserNode): string => {
-    // tslint:disable-next-line:no-unsafe-any
     if (node.charset && Iso88592.labels.includes(node.charset)) {
-      return Iso88592.decode(node.rawContent!); // tslint:disable-line:no-unsafe-any
+      return Iso88592.decode(node.rawContent!); // eslint-disable-line @typescript-eslint/no-unsafe-return
```
ioanmo226, line 402: Maybe `return String(Iso88592.decode(node.rawContent!));`?

Core/source/core/pgp-key.ts
```diff
@@ -24,17 +24,17 @@ export type Contact = {
...
   longids: string[];
   keywords: string | null;
-  pending_lookup: number;
-  last_use: number | null;
-  pubkey_last_sig: number | null;
-  pubkey_last_check: number | null;
+  pending_lookup: number; // eslint-disable-line @typescript-eslint/naming-convention
+  last_use: number | null; // eslint-disable-line @typescript-eslint/naming-convention
+  pubkey_last_sig: number | null; // eslint-disable-line @typescript-eslint/naming-convention
+  pubkey_last_check: number | null; // eslint-disable-line @typescript-eslint/naming-convention
```
ioanmo226, line 37: Maybe better to disable eslint for block rather than disable eslint per line?

```ts
/* eslint-disable */
export type Contact {
...
}
/* eslint-enable */

```

Core/source/mobile-interface/validate-input.ts
```diff
@@ -207,26 +203,20 @@ const hasProp = (
       && value.filter((x: any) => typeof x === 'string').length === value.length;
   }
   if (type === 'PrvKeyInfo?') {
-    if (value === null) {
+    if (value === null) { // eslint-disable-line no-null/no-null
```
ioanmo226, line 206: I saw many `// eslint-disable-line no-null/no-null` lines.
I think we use `null` a lot in this project and I don't think `null` is unsafe to use.
Maybe we can change eslint setting to disable `no-null/no-null` rule?

ioanmo226 (commented):
This change looks great 👍 
Just a few notes below

appium/api-mocks/core/msg-block-parser.ts
sosnovsky: This code was used only in `extractFcAttachments`, which isn't used anywhere, so I just removed it.

Core/source/test/test-utils.ts
sosnovsky: Yes, next line has `JSON.parse(block)` with unknown structure which leads to `unsafe-return` error.

Core/source/test.ts
sosnovsky: The same error with `JSON.parse` here too

ioanmo226: Got it.
In this case, maybe we can change code like this?
```ts
const blocks = decryptData.toString().split('\n').map(block => JSON.parse(block) as MsgBlock);
  expect(blocks[1].decryptErr?.error.type).equals('no_mdc');
```

Core/source/test/test-utils.ts
ioanmo226: Ok, please check my below comment

ioanmo226:
By the way, @sosnovsky How about you handle this task (https://github.com/FlowCrypt/flowcrypt-browser/issues/4526)?
As you already did tslint migration here.

cc: @tomholub

Core/source/test.ts
ioanmo226: My pleasure :)

sosnovsky:
> By the way, @sosnovsky How about you handle this task ([FlowCrypt/flowcrypt-browser#4526](https://github.com/FlowCrypt/flowcrypt-browser/issues/4526)) when you are free? As you already did tslint migration here.

I'm good with it - currently have a couple tasks for `flowcrypt-ios`, but then can work on tslint migration for browser code (if it'll be priority task there)

ioanmo226:
tslint migration task is marked as `third priority` and i don't think it's urgent

ioanmo226 (approved):
👍

PR #1769, opened 2022-07-28
#1729 Fix: archive thread too aggressive

### LINKED ISSUE #1729: "archive thread" too aggressive
I've encountered this once or twice without being exactly sure what happened, but somehow the functionality archives more than I'd expect. I think the repro steps could be:

1) open a conversation with one or more emails in it
2) send one more email to that conversation from another device while the user has it open
3) archive conversation from iOS

It seems it will also archive the email that was received while user was reading the previous email in the conversation.

Expected: only archive emails that were present in the conversation the moment user opened the conversation, not the moment they decide to archive

I believe that will bring it closer to Gmail app behavior. Right after the archive + refresh of inbox, the conversation should still be there, with one unread email.

Good to have a mock test for this, too.

ioanmo226:
This PR fixed `archive thread` too aggressive issue

close #1729 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


appium/tests/specs/mock/inbox/CheckArchiveThreadTooAggressive.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    };
+
+    await mockApi.withMockedApis(async () => {
+      await SplashScreen.mockLogin();
+      await SetupKeyScreen.setPassPhrase();
+      await MailFolderScreen.checkInboxScreen();
+
+      await browser.pause(1000000);
+      await MailFolderScreen.clickOnEmailBySubject('Signed and encrypted message');
+      await EmailScreen.clickArchiveButton();
```
ioanmo226, line 35: @sosnovsky I need to mock new message arrive event here.
As you can see from this PR, test thread has 3 messages.
But I wanna gmail mock api to return thread with 2 messages first and then return 3 messages after user clicks `archive`.
This is to test `archive thread` too aggressive feature.

But not sure how to achieve this.
You have better understanding of how mock api works.
Do you have any suggestion on this?

FlowCrypt/Controllers/Threads/MessageThread.swift
```diff
@@ -34,7 +34,11 @@ struct MessageThread: Equatable {
     }
 
     var isArchived: Bool {
-        !labels.contains(.inbox) && !labels.contains(.sent)
+        guard let firstMessageLabels = messages.first?.labels else {
+            return false
+        }
+        // Thread is treaded as archived when first message label doesn't contain inbox & sent label
+        return !firstMessageLabels.contains(.inbox) && !firstMessageLabels.contains(.sent)
```
ioanmo226, line 41: I changed `isArchived` logic because it's not correct & same as gmail behavior.
As you can see from below screenshot, this thread is archived one and is shown as archived in gmail (but it wasn't on our app) so I had to update logic to match gmail logic.

image

appium/tests/specs/mock/inbox/CheckArchiveThreadTooAggressive.spec.ts
sosnovsky: @ioanmo226 You can change messages list by calling `mockApi.addGoogleAccount` with updated list:
```swift
mockApi.addGoogleAccount('e2e.enterprise.test@flowcrypt.com', {
   messages: [testMessage, testMessage2, testMessage3],
});
```

You can also rename `addGoogleAccount` to  `configGoogleAccount`, so it'll better describe functionality.

sosnovsky: But in this case changes for existing messages will be reset (`inbox` and `sent` labels will be added back for archived messages). It'll be better to add separate method for adding new messages, I'll work on it

FlowCrypt/Controllers/Threads/MessageThread.swift
sosnovsky: But then thread will still be shown as `archived` even when new message is received?

ioanmo226: Ohhh.. Good point!!!
I need to fix above code a bit

ioanmo226: Fixed

appium/tests/specs/mock/inbox/CheckArchiveThreadTooAggressive.spec.ts
ioanmo226: Got it. Thank you

ioanmo226: Please let me know once done. So that I can continue work on this test.

FlowCrypt/Controllers/Threads/MessageThread.swift
```diff
@@ -37,7 +37,7 @@ struct MessageThread: Equatable {
         guard let firstMessageLabels = messages.first?.labels else {
             return false
         }
-        // Thread is treaded as archived when first message label doesn't contain inbox & sent label
+        // Thread is treaded as archived when labels don't contain `inbox` and first message label doesn't contain sent label
```
sosnovsky, line 40: Looks like it fails in this situation:
1) receive message
2) archive it
3) go to 'All Mail' and send reply to this message
4) go to 'Sent' - this thread is visible (not archived, which is right), but `isArchived` is `true` (should be `false`)

appium/tests/specs/mock/inbox/CheckArchiveThreadTooAggressive.spec.ts
sosnovsky: @ioanmo226 I added `addGoogleMessage` mock method - https://github.com/FlowCrypt/flowcrypt-ios/pull/1769/commits/25157192684be8fe0e0aa212083c7a4016be5f71

You can use it like this
```js
mockApi.addGoogleMessage('e2e.enterprise.test@flowcrypt.com', addedMessageSubject);
```

FlowCrypt/Controllers/Threads/MessageThread.swift
ioanmo226: Well, `isArchived` doesn't seem to correct word but value should be `true` I think.
So that it can show `move to inbox` button for this thread and this thread moves to `inbox` when tapped.

appium/tests/specs/mock/inbox/CheckArchiveThreadTooAggressive.spec.ts
ioanmo226: Got it. Thanks

FlowCrypt/Controllers/Threads/MessageThread.swift
ioanmo226: Maybe field name should be `shouldShowMoveToInboxButton`?

ioanmo226: let me know your thoughts

sosnovsky: I checked Gmail app - for this message it doesn't show `move to inbox` and `move to archive` buttons, just `delete`.
It works the same in our `master` branch, but in this branch it shows `move to inbox`, which is wrong.

sosnovsky: Ah, Gmail app shows `Move to inbox` action only when you open this message in `All Mail` folder, but not in `Sent` folder

sosnovsky: > Maybe field name should be `shouldShowMoveToInboxButton`?

This field name sounds good for me, but we need to update it's logic to work the same as in Gmail app (show `move to inbox` folder depending on current folder - `sent` or `all mail`)

ioanmo226: Hmm.. For above case in my gmail(browser version), it shows `archive` button in both `sent` and `all mail` folder (not showing `move to inbox` button)
But `archive` button doesn't have any effect in gmail app and seems like a bug.

ioanmo226: For above case, I think we should show `move to inbox` button for this thread as currently implemented(as thread is archived).

ioanmo226: Are you testing in gmail browser version or mobile app version?

sosnovsky: I check Gmail iOS app

ioanmo226: I see. Let me show `move to inbox` button only for `all mail` folder.

sosnovsky: > I think there's nothing wrong to show `move to inbox` button for this thread (even for sent folder)
> What do you suggest?

Okay, let's leave `move to inbox` button on `sent` screen, so we just need to rename property to `shouldShowMoveToInboxButton`

ioanmo226: okay

appium/tests/specs/mock/inbox/CheckArchiveThreadTooAggressive.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    await mockApi.withMockedApis(async () => {
+      await SplashScreen.mockLogin();
+      await SetupKeyScreen.setPassPhrase();
+      await MailFolderScreen.checkInboxScreen();
+
+      await browser.pause(1000);
+      mockApi.addGoogleMessage('e2e.enterprise.test@flowcrypt.com', addedMessageSubject);
+      await MailFolderScreen.refreshMailList();
+      await browser.pause(10000000);
```
ioanmo226, line 37: @sosnovsky `addGoogleMessage` doesn't seem to work as expected (or am i doing something wrong?)
It is showing 2 threads if I call this method.
Could you check here?

image

sosnovsky: @ioanmo226 `message-export-18244f5e83a4a7aa.json` has different `threadId` than previous messages. It should be `18244f525d7e0023` instead of `18244f5e83a4a7aa`

ioanmo226: Aha, got it. Thanks

FlowCrypt/Functionality/Mail Provider/Threads/MessagesThreadOperationsProvider.swift
```diff
@@ -71,7 +71,14 @@ extension GmailService: MessagesThreadOperationsProvider {
     }
 
     func archive(thread: MessageThread, in folder: String) async throws {
-        try await update(thread: thread, labelsToRemove: [.inbox, .sent])
+        try await withThrowingTaskGroup(of: Void.self) { taskGroup in
+            for message in thread.messages {
```
sosnovsky, line 75: What's advantage of manually updating each message vs using `update(thread:...`?

I thought `GTLRGmail_ModifyThreadRequest` updates each message too (as documentation says - https://developers.google.com/gmail/api/reference/rest/v1/users.threads/modify)

FlowCrypt/Controllers/Threads/MessageThread.swift
```diff
@@ -33,8 +33,13 @@ struct MessageThread: Equatable {
...
 
-    var isArchived: Bool {
-        !labels.contains(.inbox) && !labels.contains(.sent)
+    var shouldShowMoveToInboxButton: Bool {
+        guard let firstMessageLabels = messages.first?.labels else {
+            return false
+        }
+        // Thread is treaded as archived when labels don't contain `inbox` and first message label doesn't contain sent label
+        // https://github.com/FlowCrypt/flowcrypt-ios/pull/1769#discussion_r931874353
+        return !isInbox && !firstMessageLabels.contains(.sent)
```
sosnovsky, line 42: I found another case when `move to archive` button isn't visible - send message from Flowcrypt, then open it:
- Gmail app shows 'move to archive` button
- Flowcrypt doesn't show it (because of `firstMessageLabels.contains(.sent)`)

FlowCrypt/Functionality/Mail Provider/Threads/MessagesThreadOperationsProvider.swift
ioanmo226: Well in this task, Tom mentioned archive thread feature is too aggressive.
If we just call `GTLRGmail_ModifyThreadRequest`, it would archive all emails in this thread (at the time user taps archive button).
But with updated logic, it would only archive messages which are currently displayed.


> It seems it will also archive the email that was received while user was reading the previous email in the conversation.
> 
> Expected: only archive emails that were present in the conversation the moment user opened the conversation, not the moment they decide to archive
> 
> I believe that will bring it closer to Gmail app behavior. Right after the archive + refresh of inbox, the conversation should still be there, with one unread email.
> 
>

FlowCrypt/Controllers/Threads/MessageThread.swift
ioanmo226: `Move to archive` button on gmail app for this case doesn't work at all i think.

ioanmo226: I think in gmail app, sent emails can't be archived

FlowCrypt/Functionality/Mail Provider/Threads/MessagesThreadOperationsProvider.swift
sosnovsky: Sounds good 👍

FlowCrypt/Controllers/Threads/MessageThread.swift
sosnovsky: Yeah, just re-checked it - `archive` button is visible just for some sent messages, but invisible for others. And message stays in `sent` folder even after tapping on this button.
Then everything is ok here 👍

sosnovsky (dismissed):

appium/tests/specs/mock/inbox/CheckArchiveThreadTooAggressive.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+      await MailFolderScreen.checkInboxScreen();
+
+      await MailFolderScreen.clickOnEmailBySubject(testMessage);
+      await EmailScreen.clickArchiveButton();
+
+      // Archived thread doesn't appear
+      await MailFolderScreen.checkEmailIsNotDisplayed(testMessage);
+
+      // Add message to archived thread (which simulates new incoming message from same sender)
+      mockApi.addGoogleMessage('e2e.enterprise.test@flowcrypt.com', addedMessageSubject);
```
sosnovsky, line 41: @ioanmo226 For reproducing `It seems it will also archive the email that was received while user was reading the previous email in the conversation.` we'll need to run `mockApi.addGoogleMessage` before tapping `archive` button.
So it should be moved to line 35

sosnovsky (changes requested):

appium/tests/specs/mock/inbox/CheckArchiveThreadTooAggressive.spec.ts
ioanmo226: Got it

FlowCrypt/Functionality/Mail Provider/Threads/MessagesThreadOperationsProvider.swift
```diff
@@ -71,7 +71,16 @@ extension GmailService: MessagesThreadOperationsProvider {
...
-        try await update(thread: thread, labelsToRemove: [.inbox, .sent])
+        // manually updated each message rather than using update(thread:...) method
+        // https://github.com/FlowCrypt/flowcrypt-ios/pull/1769#discussion_r932964129
+        try await withThrowingTaskGroup(of: Void.self) { taskGroup in
+            for message in thread.messages {
+                taskGroup.addTask {
+                    try await self.archiveMessage(message: message, folderPath: folder)
+                }
+            }
+            try await taskGroup.waitForAll()
```
sosnovsky, line 82: @ioanmo226 I found method for batch updating messages with Gmail API - https://developers.google.com/gmail/api/reference/rest/v1/users.messages/batchModify

Let's use it, so only 1 request will be needed for archiving any thread (even if it contains many messages)

sosnovsky (changes requested):

FlowCrypt/Functionality/Mail Provider/Threads/MessagesThreadOperationsProvider.swift
ioanmo226: Sounds good!
I also worried about thread (which contains tens of messages) and this seems good solution

appium/api-mocks/apis/google/google-endpoints.ts
```diff
@@ -241,6 +246,18 @@ export const getMockGoogleEndpoints = (
         return {}
       }
 
+      throw new HttpErr(`Method not implemented for ${req.url}: ${req.method}`);
+    },
+    '/gmail/v1/users/me/messages/batchModify': async (parsedReq, req) => {
```
sosnovsky, line 251: Well done 👍

sosnovsky (approved):

Issue #782, opened 2021-10-26
[June 2022] consider migration to SwiftUI


tomholub:
Ok to give SwiftUI a try in the future but can't afford the code churn right now - need to get the app stable first. 

Can evaluate it around June 2022, when our UI tests are comprehensive and we have spare dev cycles.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/766#issuecomment-950699871_

`InvalidStorageViewController` could be a good first candidate to test as it's small and not frequently encountered

Kharchevskyi:
https://github.com/FlowCrypt/flowcrypt-ios/pull/1126#discussion_r759672606
Please keep in mind that swift ui mostly use observing patterns to update it's ui

tomholub:
noted - thanks for mentioning it. 

We'll tackle that once we start working on SwiftUI.

PR #1776, opened 2022-08-02
#1702 Use prebuilt texture dependency

### LINKED ISSUE #1702: speed up project build in ci with prebuilt dependency - Texture
It take 28 minutes to build the project, and I seem to recall that Texture took a good chunk of that. Is that true? If yes, could the build times be sped up by making a prebuilt swift package?

image

ioanmo226:
This PR uses prebuilt texture dependency to speedup ci build

close #1702 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.0.0",
       "license": "ISC",
       "devDependencies": {
-        "@typescript-eslint/eslint-plugin": "^5.31.0",
+        "@typescript-eslint/eslint-plugin": "^5.32.0",
         "@typescript-eslint/parser": "^5.32.0",
         "eslint": "8.21.0",
         "eslint-plugin-header": "3.1.1",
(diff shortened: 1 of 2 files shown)


ioanmo226:
@sosnovsky As I checked with this prebuilt Texture framework, CI speed is about 2 mins faster(still build fails though. only unit tests pass)
Do you think it's worth we still go forward and fix? 
FYI I used [pod builder](https://github.com/Subito-it/PodBuilder) library and it has few maintainers.

sosnovsky:
@ioanmo226 I think we should check how much time takes building of each pod (with tool like https://github.com/MobileNativeFoundation/XCLogParser) to see if precompiling `Texture` will have a big effect on total build time.

Maybe it's possible to manually compile `Texture.xcframework` (without using additional gem like `podbuilder`) or add support for SPM (here is fork of original `Texture` with added SPM support)?

sosnovsky:
@ioanmo226 it looks like `Texture` won't have any new releases in the nearest future, as there is no new commits since December. So for now it'll be okay just manually compile `xcframework` without adding new libraries.

Maybe in the future we'll switch to SwiftUI and will be able to completely remove `Texture` - https://github.com/FlowCrypt/flowcrypt-ios/issues/782

ioanmo226:
Sounds good

sosnovsky:
@ioanmo226 I missed this link in my first comment - https://github.com/shimastripe/Texture/releases/tag/3.1.1 (it's a fork with SPM support)

ioanmo226:
Which way do you prefer(SPM vs manually compile framework file)?
I think manually compiling framework would be fine as we would move to `SwiftUI` soon.

sosnovsky:
@ioanmo226 I prefer SPM as including full `xcframework` will make `FlowCrypt` repository size much bigger.

It'll be better to create our own fork with SPM support (as for [mailcore](https://github.com/FlowCrypt/mailcore2) and [GMP](https://github.com/FlowCrypt/GMP)). @tomholub will need to create new repository (https://github.com/FlowCrypt/Texture), when he'll be available

PR #1786, opened 2022-08-10
#1379 Update Core build to support openpgpjs 5 and webpack 5

### LINKED ISSUE #1379: make OpenPGP.js v5 code work with iOS JavaScriptCore

sosnovsky:
This PR updates `Core` build process to support `openpgpjs 5` and `webpack 5`

close #1379

- Tests added or updated

### To be filled by reviewers


tomholub:
Another achievement of epic proportions :-)

sosnovsky:
> Another achievement of epic proportions :-)

Yeah, it took a lot of attempts to make it finally work :)

Core/tooling/fix-bundles.js
```diff
@@ -70,65 +69,48 @@ const replace = (libSrc, regex, replacement) => {
...
+/* disabled because it works faster without this change */
+// entrypointBareSrc = replace( // rsa verify on host
+//   entrypointBareSrc,
+//   /return publicKey\.rsa\.verify\(hashAlgo, data, s, n, e, hashed\)/, `
+//   // returns empty str if not supported: js fallback below
+//   const computed = await coreHost.modPow(s.toString(10), e.toString(10), n.toString(10));
+//   return computed
+//     ? new bn.default(computed, 10).toArrayLike(Uint8Array, 'be', n.byteLength())
+//     : await publicKey.rsa.verify(hashAlgo, data, s, n, e, hashed);`
+// );
```
sosnovsky, line 90: @tomholub I disabled these code changes for decrypting and verifying changes on host and `testRsaPerformance` test still passes (https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-1379-openpgpjs-5/FlowCryptAppTests/Core/FlowCryptCoreTests.swift#L391)

Probably `openpgp` had some performance improvements and we don't need to use iOS code for this functionality?

Core/source/core/pgp-msg.ts
```diff
@@ -173,19 +174,19 @@ export class PgpMsg {
         // .. which is not really an issue - an attacker that can append signatures
         // could have also just slightly changed the message, causing the same experience
         // .. so for now #wontfix unless a reasonable usecase surfaces
-        sig.match = (sig.match === true || sig.match === null) && await verifyRes.verified;
         if (!sig.signer) {
           // todo - currently only the first signer will be reported.
           // Should we be showing all signers? How common is that?
           sig.signer = await PgpKey.longid(verifyRes.keyID.bytes);
         }
+        sig.match = (sig.match === true || sig.match === null) && await verifyRes.verified;
```
sosnovsky, line 182: `verifyRes.verified` throws on invalid signature in `openpgpjs 5`. This lead to `sig.signer` to be `null`, so I moved `sig.match` below.

ioanmo226 (commented):
Great work @sosnovsky  👍 
Looks good on my side.
Better @tomholub finalize review as I'm not familiar with `Core` code and I might miss something.

sosnovsky:
Thanks @ioanmo226 

Yeah, let's wait when @tomholub will be available to check these changes.

Core/tooling/fix-bundles.js
tomholub: The test may have fairly benevolent maximum time allowed to not be too flaky. Could you see what are the individual operation times on v4 vs v5 as it is now?

Core/source/core/pgp-msg.ts
tomholub: And I suppose we had some test that expected `sig.signer` to be set even when it's a bad signature? Could you point me to it?

Core/tooling/build-final.js
```diff
@@ -7,22 +7,15 @@ const path = {
...
-  .replace(/require\(['"]encoding-japanese\.js['"]\)/g, 'dereq_encoding_japanese');
+};
 
 // bare
-const bareDepsSrc = fs.readFileSync(path.bareDepsBundle).toString()
-const bareEntrypointSrc = fs.readFileSync(path.bareEntrypointBundle).toString().replace("'[BUILD_REPLACEABLE_VERSION]'", 'APP_VERSION');
+const bareDepsSrc = fs.readFileSync(path.bareDepsBundle).toString();
+const bareEntrypointSrc = fs
+  .readFileSync(path.bareEntrypointBundle).toString()
+  .replace("'[BUILD_REPLACEABLE_VERSION]'", 'APP_VERSION');
```
tomholub, line 16: Maybe `  .replace("'[BUILD_REPLACEABLE_VERSION]'", 'APP_VERSION');` could be omitted, it doesn't seem to be doing anything useful, from a casual look. The actual version is set from `CoreHost` somewhere, if I remember. Let's leave it for another PR, this one is already looking good.

Core/tooling/fix-bundles.js
tomholub: Just to understand:
 - `rsa verify on host` was not replaced because it's faster when not replacing it - that sounds good. In that case, this chunk of code could be removed instead of just commented out. How much faster exactly?
 - `rsa decrypt on host` - could you also check and compare the time before & after? using rsa4096 key like in the performance test.

tomholub (approved):
This looks good, the other things could be settled, per comments.

Also now that the build process is fixed and modernized, we should copy the built result into resources folder each time the js code is built. It no longer has to be a separate step.

tomholub:
@sosnovsky this is better than I would have done it :+1:

Core/source/core/pgp-msg.ts
sosnovsky: Yes, we use `sig.signer` when detecting type of signature error - https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/Functionality/Mail%20Provider/Message%20Provider/MessageService.swift#L256

Previously openpgp didn't throw error when `sig.match = nil`, but in v5 this logic changed and now we need manually check error type for `.missingPubKey` error. (old code - https://github.com/FlowCrypt/flowcrypt-ios/blob/e89baece51bcca00b8ad4c58193541a8ab863d4f/FlowCrypt/Functionality/Mail%20Provider/Message%20Provider/MessageService.swift#L259)

This change lead to fail of `CheckAllEmailSignatureCases` test - https://github.com/FlowCrypt/flowcrypt-ios/blob/master/appium/tests/specs/mock/inbox/CheckAllEmailSignatureCases.spec.ts#L64

sosnovsky:
> @sosnovsky this is better than I would have done it 👍

thanks :)

Core/tooling/build-final.js
sosnovsky: We set `APP_VERSION` value in `CoreHost` code, so this change is needed for using value provided by `CoreHost`
https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/Core/Core.swift#L230

tomholub: Ah, I see.

PR #1840, opened 2022-09-29
#1805 configure pass phrase session length with client configuration

### LINKED ISSUE #1805: pass phrase session length should be configurable with client configuration
The returned JSON may contain `"in_memory_pass_phrase_session_length": 3600` for example. The number is an integer and to be interpreted as amount of seconds a pass phrase session will last. Currently hard-coded as 4 hours.

This mechanism applies only when browser extension is storing pass phrase in-memory. In other words, it does not affect the decision as to where the pass phrase will be stored, that was decided either by user or by `FORBID_STORING_PASS_PHRASE`. If the pass phrase is stored in storage, whatever the reason for it, this new client configuration will have no effect.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/4613#issuecomment-1228303813_

ioanmo226:
This PR configures pass phrase session length with client configuration

close #1805 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


FlowCryptAppTests/Functionality/Services/PassPhraseStorageTests/InMemoryPassPhraseStorageTest.swift
```diff
@@ -50,26 +49,28 @@ class InMemoryPassPhraseStorageTest: XCTestCase {
...
-        XCTAssertTrue(try sut.getPassPhrases(for: testPassPhraseAccount).isEmpty)
-
-        sut.save(passPhrase: testPassPhrase)
-        sleep(3)
-        XCTAssertTrue(try sut.getPassPhrases(for: testPassPhraseAccount).isEmpty)
+    func testExpiredPassPhrases() async throws {
+//        let passphrase = try await sut.getPassPhrases(for: testPassPhraseAccount)
+//        XCTAssertTrue(passphrase.isEmpty)
+//
+//        sut.save(passPhrase: testPassPhrase)
```
ioanmo226, line 70: @sosnovsky, @tomholub 
I currently have no idea how to implement `testExpiredPassPhrases` test with unit test.
I'm gonna implement it with appium ui test.
Could you tell me if you have better idea using unit test?

Before that, please check PR is implemented in correct way.

tomholub: It's complex behavior involving client configuration, app state etc, I think it would be quite appropriate to test with a ui test.

FlowCrypt/App/AppContext.swift
```diff
@@ -57,6 +60,7 @@ class AppContext {
             session: sessionType,
             userAccountService: try SessionService(
                 encryptedStorage: encryptedStorage,
+                passPhraseStorage: InMemoryPassPhraseStorage(encryptedStorage: encryptedStorage),
```
tomholub, line 63: naming mismatch - passPhraseStorage unclear if memory, stored, or general, with InMemoryPassPhraseStorage passed in

Also, more generally, it's odd that `InMemoryPassPhraseStorage` needs `encryptedStorage` passed into it. Not clear what for?

tomholub: Wouldn't it be more appropriate to use the `combinedPassPhraseStorage` already defined below?

FlowCrypt/Functionality/DataManager/SessionService.swift
```diff
@@ -52,7 +52,7 @@ final class SessionService {
 
     init(
         encryptedStorage: EncryptedStorageType,
-        passPhraseStorage: PassPhraseStorageType = InMemoryPassPhraseStorage(),
+        passPhraseStorage: PassPhraseStorageType,
```
tomholub, line 55: Same here, naming mismatch in original code

FlowCrypt/Functionality/Services/Local Private Key Services/CombinedPassPhraseStorage.swift
```diff
@@ -87,19 +87,19 @@ final class CombinedPassPhraseStorage: CombinedPassPhraseStorageType {
 
     init(
         encryptedStorage: PassPhraseStorageType,
-        inMemoryStorage: PassPhraseStorageType = InMemoryPassPhraseStorage()
+        inMemoryStorage: PassPhraseStorageType
```
tomholub, line 90: And here in original code

FlowCrypt/Functionality/Services/Local Private Key Services/InMemoryPassPhraseStorage.swift
```diff
@@ -38,7 +38,15 @@ final class InMemoryPassPhraseStorage: PassPhraseStorageType {
...
-    func getPassPhrases(for email: String) throws -> [PassPhrase] {
+    func getPassPhrases(for email: String) async throws -> [PassPhrase] {
+        let enterpriseServer = try EnterpriseServerApi(email: email)
+        let clientConfigurationService = ClientConfigurationService(
+            server: enterpriseServer,
+            local: LocalClientConfiguration(
+                encryptedStorage: self.encryptedStorage
+            )
+        )
+        let sessionLegnth = try await clientConfigurationService.configuration.passphraseSessionLengthInSeconds
```
tomholub, line 49: let's call it `sessionLengthInSeconds`

tomholub (commented):
Looks good, just some confusion about naming, already present in original code, plus concern about why in-memory class has encrypted storage as a dependency.

FlowCrypt/App/AppContext.swift
ioanmo226: I added `encryptedStorage` in `InMemoryPassPhraseStorage` because ClientConfigurationService required `LocalClientConfiguration` which requires `encryptedStorage`
I couldn't find better way to get `clientConfiguration.passphraseSessionLengthInSeconds` from `InMemoryPassPhraseStorage.getPassPhrases` function.

image

sosnovsky:
@ioanmo226 you'll also need to add `inMemoryPassPhraseSessionLength` property for [ClientConfigurationRealmObject.](https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-1805/FlowCrypt/Models/Realm%20Models/ClientConfigurationRealmObject.swift#L12)

As currently initializing `RawClientConfiguration` from [ClientConfigurationRealmObject](https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-1805/FlowCrypt/Models/Common/RawClientConfiguration.swift#L76) will ignore it.

FlowCrypt/App/AppContext.swift
ioanmo226: you mean to pass `ClientConfiguration` to `InMemoryPassPhraseStorage` constructor or `InMemoryPassPhraseStorage.getPassPhrases`?

tomholub: `EmailKeyManagerApi` doesn't get initialized in AppContext, meaning that we can use fresh client configuration when initializing it.

Not so with the pass phrase storage classes.

There is another option, instead of full-fledged `encryptedStorage` you could require `clientConfigurationStorage: ClientConfigurationStorage` interface. Then in AppContext you pass `clientConfigurationStorage: encryptedStorage`. `EncryptedStorage` would of course implement that interface. That would work for me: the intent would be clearer.

tomholub: Sometimes these kinds of issues aren't really with the code or with the implementation, but with properly communicating intent through the code to avoid confusion.

tomholub:
> @ioanmo226 you'll also need to add `inMemoryPassPhraseSessionLength` property for [ClientConfigurationRealmObject.](https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-1805/FlowCrypt/Models/Realm%20Models/ClientConfigurationRealmObject.swift#L12)
> 
> As currently initializing `RawClientConfiguration` from [ClientConfigurationRealmObject](https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-1805/FlowCrypt/Models/Common/RawClientConfiguration.swift#L76) will ignore it.

The UI test should target this too

FlowCrypt/Functionality/Services/Local Private Key Services/CombinedPassPhraseStorage.swift
ioanmo226: We can't keep original code because `InMemoryPassPhraseStorage` now requires `encryptedStorage`

ioanmo226: @tom

ioanmo226:
@sosnovsky Does UI test work on your local after you upgrade to Xcode 14?
I thought it was problem with iOS 16 but when appium doesn't even run app in iOS 15(with Xcode 14)

sosnovsky:
@ioanmo226 I installed beta version of `appium 2` with `xcuitest` driver and it works well in Xcode 14.
To use beta version I updated [package.json](https://github.com/FlowCrypt/flowcrypt-ios/blob/master/appium/package.json#L33):
```
"appium": "^2.0.0-beta.44",
"appium-xcuitest-driver": "^4.11.1",
```

sosnovsky:
@ioanmo226 I'll create new PR with fix for https://github.com/FlowCrypt/flowcrypt-ios/issues/1842 in the next hour, so we can merge it to `master` and use in other branches.

ioanmo226:
Sounds good. Thanks

FlowCrypt/App/AppContext.swift
ioanmo226: @tomholub I think that won't work that way.
If we try that way, swift compiler would throw this error.

image

sosnovsky: @ioanmo226 what about adding `var clientConfiguration: ClientConfiguration?` property to `CombinedPassPhraseStorage`? It will be set on [AppContextWithUser](https://github.com/FlowCrypt/flowcrypt-ios/blob/feature/issue-1805/FlowCrypt/App/AppContext.swift#L164) initialization like this:
```swift
self.clientConfigurationService = ClientConfigurationService(
    server: enterpriseServer,
    local: LocalClientConfiguration(
        encryptedStorage: encryptedStorage
    )
)

combinedPassPhraseStorage.configuration = clientConfigurationService.configuration
super.init(
    ...
)
```
And then we can pass `passphraseSessionLengthInSeconds` value directly to `getPassphrases` methods.

What do you think?

sosnovsky:
@ioanmo226 a couple tests are failing after updating to Xcode 14 and iOS 16, currently working on fix in https://github.com/FlowCrypt/flowcrypt-ios/tree/bugfix/issue-1842-ui-tests
You can check updated appium config there to run tests in your branch

FlowCrypt/App/AppContext.swift
ioanmo226: @sosnovsky
That sounds good to me and implemented this way.
Please check

FlowCrypt/App/AppContext.swift
```diff
@@ -154,20 +150,22 @@ class AppContextWithUser: AppContext {
...
-        self.enterpriseServer = try await EnterpriseServerApi(email: user.email)
+        self.enterpriseServer = try EnterpriseServerApi(email: user.email)
         self.clientConfigurationService = ClientConfigurationService(
             server: enterpriseServer,
             local: LocalClientConfiguration(
                 encryptedStorage: encryptedStorage
             )
         )
 
+        var updatedCombinedPassPhraseStorage = combinedPassPhraseStorage
```
sosnovsky, line 161: Let's call it `combinedPassPhraseStorageWithConfiguration`

FlowCrypt/App/AppContext.swift
sosnovsky: @ioanmo226 looks much more clear than before 👍

FlowCrypt/App/AppContext.swift
ioanmo226: Just changed

sosnovsky:
@ioanmo226 please fetch the latest changes from https://github.com/FlowCrypt/flowcrypt-ios/tree/bugfix/issue-1842-ui-tests

appium/tests/specs/mock/setup/RespectsPassPhraseSessionLength.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+
+describe('SETUP: ', () => {
+
+  it('respects in_memory_pass_phrase_session_length and passphrase expire in that seconds', async () => {
+
+    const mockApi = new MockApi();
+    const testMessageSubject = 'Message with cc and multiple recipients and text attachment';
+    const processArgs = CommonData.mockProcessArgs;
+
+    mockApi.fesConfig = {
```
sosnovsky, line 20: It can be just `mockApi.fesConfig = MockApiConfig.defaultEnterpriseFesConfiguration`

appium/tests/specs/mock/setup/RespectsPassPhraseSessionLength.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+        clientConfiguration: {
+          flags: ["NO_PRV_CREATE", "NO_PRV_BACKUP", "NO_ATTESTER_SUBMIT", "PRV_AUTOIMPORT_OR_AUTOGEN", "FORBID_STORING_PASS_PHRASE"],
+          key_manager_url: CommonData.keyManagerURL.mockServer,
+          in_memory_pass_phrase_session_length: 5
+        }
+      };
+      await AppiumHelper.restartApp(processArgs);
+      await browser.pause(5000);
+      await MailFolderScreen.clickOnEmailBySubject(testMessageSubject);
+      await BaseScreen.checkModalMessage('Please enter pass phrase');
```
sosnovsky, line 54: On app restart all passphrases stored in memory are removed, so current test logic doesn't test `in_memory_pass_phrase_session_length` config.

You should open message after app restart, enter passphrase, go back to inbox screen, wait for 5 seconds and open message again - it should ask for passphrase.

ioanmo226: Nice catch 👍

sosnovsky (approved):
👍

PR #1881, opened 2022-10-20
#1834 attester api methods naming

### LINKED ISSUE #1834: renaming of attester public key submit methods and welcome-message
The methods should be named as follows. Both are POST, one of them uses id_token and the other one not. Usage should reflect the comments.

```ts
  /**
   * Set or replace public key with idToken as an auth mechanism
   * Used during setup
   * Can only be used for primary email because idToken does not contain info about aliases
   */
  public submitPrimaryEmailPubkey = async (email: string, pubkey: string, idToken: string): Promise<void> => {
    if (!this.clientConfiguration.canSubmitPubToAttester()) {
      throw new Error('Cannot replace pubkey at attester because your organisation rules forbid it');
    }
    await this.pubCall(`pub/${email}`, 'POST', pubkey, { authorization: `Bearer ${idToken}` });
  };

  /**
   * Request to replace pubkey that will be verified by clicking email
   * Used when user manually chooses to replace key
   * Can also be used for aliases
   */
  public submitPubkeyWithConditionalEmailVerification = async (email: string, pubkey: string): Promise<string> => {
    if (!this.clientConfiguration.canSubmitPubToAttester()) {
      throw new Error('Cannot replace pubkey at attester because your organisation rules forbid it');
    }
    const r = await this.pubCall(`pub/${email}`, 'POST', pubkey);
    return r.responseText;
  };
```

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-android/issues/1984#issuecomment-1255147387_

sosnovsky:
This PR renames attester api methods

close #1834

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
+++ b/FlowCrypt/Controllers/Setup/SetupGenerateKeyViewController.swift
@@ -84,9 +84,12 @@ final class SetupGenerateKeyViewController: SetupCreatePassphraseAbstractViewCon
         }
 
         // sending welcome email is not crucial, so we don't handle errors
-        _ = try? await attester.testWelcome(
+        guard let idToken = try? await IdTokenUtils.getIdToken(user: appContext.user) else { return }
+
+        try? await attester.testWelcome(
             email: appContext.user.email,
-            pubkey: encryptedPrv.key.public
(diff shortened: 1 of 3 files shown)


ioanmo226 (approved):
👍

PR #1885, opened 2022-10-25
#1702 Build optimizations

### LINKED ISSUE #1702: speed up project build in ci with prebuilt dependency - Texture

sosnovsky:
In this PR:
- `cocoapods` dependency is removed (Swift Package Manager is used instead)
- `swiftlint` is now used through Xcode build plugin integration
- `swiftformat` config is unified for all targets
- optimized semaphore config for faster builds (now full test suite takes less than 1 hour)

close #1702

### To be filled by reviewers


ioanmo226 (approved):
Looks good 👍 
Speed is much faster now.

Issue #1909, opened 2022-11-09
Use native swift functions in WKWebView


sosnovsky:
With `JavaScriptCore` we used some native swift functions for increased performance (like `getSecureRandomByteNumberArray`, `decryptAesCfbNoPadding` etc).

Need to find way how to use them in `WKWebView` and check if they'll increase performance of `Core` functions.

tomholub:
If web view performance is this much better, we would no longer need the core Swift functions. So you can remove them.

sosnovsky:
`WKWebView` seems to be faster because it has native implementation for `crypto` functions and also `openpgp` can use some native optimisations for better performance there.

PR #1910, opened 2022-11-09
#1877 Improve decrypt performance

### LINKED ISSUE #1877: FlowCrypt ios app crash when trying to decrypt a very large message
Description:
I have recently tested the ios app regarding the security report we received and found out that the app crashed when trying to decrypt an encrypted message with 100Mb lengths of text. 

Such payload can be created through python: `python3 -c 'print("A"*100000000)' > 100_million_A.txt` and encrypt the plain text message using `gpg`

Steps to reproduce:

1. Create a file containing a large input. eg. python3 -c 'print("A"*100000000)' > 100_million_A.txt
2. Encrypt it with `gpg` and send it yourself using Gmail's built-in message composer.
3. Open the message on the FlowCrypt iOS and it will crash the app.

The android app has a limit of cutting 50Kb of decrypted messages to prevent this (though the android app is not working as expected yet since it cuts the encrypted content, not the decrypted content) - https://github.com/FlowCrypt/flowcrypt-android/blob/c5d8e4b841c784a0101f20a3854a84f406b89987/FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MessageDetailsFragment.kt#L1744

Please see https://github.com/FlowCrypt/flowcrypt-security/issues/215 for more details.

sosnovsky:
This PR increases `Core` performance by replacing `JavaScriptCore` with `WKWebView` (it's faster and also runs in separate process from app, which greatly reduces app memory and cpu usage)

Before migration decrypting 100mb message crashed app (because it was using several GBs of memory), and now it takes 10-15 seconds.

For now I haven't found how to run native swift functions through `WKWebView` (for increasing decrypting speed), but it currently doesn't affect overall app performance because of better `WKWebView` performance (created task for it - https://github.com/FlowCrypt/flowcrypt-ios/issues/1909)

I also changed `webpack` config to use `production` mode, which reduced generated js file size from 4.4mb to 2.9mb.

close #1877

- Tests added or updated

### To be filled by reviewers


Core/source/core/pgp-key.ts
```diff
@@ -383,6 +380,16 @@ export class PgpKey {
       lastModified,
       revoked: k.revocationSignatures.length > 0
     };
+
+    if (k.isPrivate()) {
+      Object.assign(keyDetails, {
+        private: k.armor(),
+        isFullyDecrypted: isFullyDecrypted(k),
+        isFullyEncrypted: isFullyEncrypted(k)
+      });
```
ioanmo226, line 389: Maybe better to return spread object(I don't prefer object assign though)
https://eslint.org/docs/latest/rules/prefer-object-spread
```ts
return {
    ...keyDetails,
    {
        private: k.armor(),
        isFullyDecrypted: isFullyDecrypted(k),
        isFullyEncrypted: isFullyEncrypted(k)
  }
};

```

sosnovsky: I like spread operator too for it's simplicity, but here we'll need to have 2 `return` to make it work:
```ts
if (k.isPrivate()) {
    return {
        ...keyDetails, 
        {
             private: k.armor(), 
             isFullyDecrypted: isFullyDecrypted(k),
             isFullyEncrypted: isFullyEncrypted(k)
        } 
    }
} else {
    return keyDetails
}

```

ioanmo226: Yeah, we indeed need it (we don't need else though).
But it's not bad for code performance and readability i think?

ioanmo226: I'm fine with your approach though

ioanmo226 (dismissed):
Changes look great 👍 
Both performance and code quality improved 👍

FlowCrypt/Core/Core.swift
```diff
@@ -27,71 +27,94 @@ protocol KeyParser {
...
-    private init() {}
+    private init() {
+        setupWebView()
+    }
+
+    // MARK: - Setup
+    func setupWebView() {
+        guard let jsFileSrc = getCoreJsFile() else { return }
+
+        let appVersion = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String ?? "1.0.0"
```
tomholub, line 63: the `??` should be `[unknown version]` or similar

ioanmo226:
@sosnovsky FYI I pushed small lint fixes to not disturb you.
By the way, did you try to login with latest changes? I got below error.

Next priority should be to fix `live test` I think. Because live test doesn't work now, semaphoreci won't notice us if something is broken in live.

image

tomholub (dismissed):
Is the webview process reused for more than one call? Because it probably takes a few hundred ms to initialize. I suppose the answer is yes.

sosnovsky:
> Is the webview process reused for more than one call? Because it probably takes a few hundred ms to initialize. I suppose the answer is yes.

@tomholub Yes, after initialization web view is reused for all calls. During testing I haven't noticed any issues with that - web view is initialized on `Core` init which happens before all calls.

sosnovsky:
> @sosnovsky FYI I pushed small lint fixes to not disturb you. 

Thanks!

> By the way, did you try to login with latest changes? I got below error.

Yes, I logged in several times today - worked well for me, and I think login code wasn't changed in this PR.

> Next priority should be to fix `live test` I think. Because live test doesn't work now, semaphoreci won't notice us if something is broken in live.

Yeah, I tried to re-enable it, but with no luck - gmail still asks for additional confirmation.

FlowCrypt/Core/Core.swift
sosnovsky: Done ✅

ioanmo226:
Hmm... not sure why it says `id token expired` in my device.
> Yes, I logged in several times today - worked well for me, and I think login code wasn't changed in this PR.

sosnovsky:
@ioanmo226 I just re-checked google login - no issues with it, does it still fail for you?

sosnovsky:
@ioanmo226 does it happen just after you tap `Continue with Gmail` or after you entered your login details? Did you try to login to another account?

sosnovsky:
@ioanmo226 I found such issue in `AppAuth` repo - https://github.com/openid/AppAuth-iOS/issues/443, maybe your device time is wrong too?

ioanmo226:
@sosnovsky You are right.
My device time was incorrect(I changed it because i had to test something)
I changed time to correct time (only had 2 hours difference though) and it's fine now.

Thank you.

ioanmo226 (approved):
👍

PR #1919, opened 2022-11-15
#1913 Fix trash button tap

### LINKED ISSUE #1913: tapping trash icon on an email does nothing
on version 1.1.1. It just shows loader for a second, but nothing changes.

sosnovsky:
This PR fixes bug when it was impossible to delete thread + adds error message alert for such cases (so it'll be easier to find reason for possible issues in the future)

close #1913

- Tests added or updated

### To be filled by reviewers


appium/api-mocks/apis/google/google-data.ts
```diff
@@ -389,14 +389,28 @@ export class GoogleData {
...
+        if (labelIds.length) {
+          const messageLabels = m.labelIds || [];
+          if (messageLabels.includes('TRASH')) {
+            return labelIds.includes('TRASH');
+          } else {
+            return messageLabels.some(l => labelIds.includes(l));
+          }
+        } else {
+          return true;
+        }
```
ioanmo226, line 403: we don't need `else` statements here. just `return` would do a job.
Not necessary to remove `else` though.

PR #1921, opened 2022-11-16
#1912 Check if attachment preview is available

### LINKED ISSUE #1912: app tries to render binary files that it cannot render
For example a kdbx file. It's just binary, and rendering it shows garbage. I can luckily still save the file and use it, but it would have been better to skip rendering it entirely.

Could we try to find a list of file extension that iOS can actually render, and use that as a whitelist for previewing? All others would be simply downloaded.

sosnovsky:
This PR adds check for attachments preview - if it's not available then app shows `download` screen instead of `preview`

close #1912

- Tests added or updated

### To be filled by reviewers


ioanmo226 (approved):
👍

PR #1925, opened 2022-11-17
#1863 Send plain messages

### LINKED ISSUE #1863: Allow to send plain messages on consumer release
On consumer release, we are currently still "forcing" the user to use password protected message when recipients don't have encryption.

Instead, in this situation, the user should get both options. That will make the iOS app usable as a general purpose email client, and will probably allow us to dodge some negative reviews from users who wish to use it that way. And also me :-) I'd like to be able to ditch the Gmail app and just use the FlowCrypt app.

No change of functionality on Enterprise app.

sosnovsky:
This PR adds ability to send non-encrypted messages to recipients without encryption set up.

close #1863

- Tests added or updated

### To be filled by reviewers


tomholub:
was the difference in functionality between enterprise and consumer build implemented and tested?

could you share a video of what it looks like?

sosnovsky:
@tomholub I just added check for enterprise users (+ ui test), so now we have 3 different alerts when recipient doesn't have public key:
- sender uses enterprise build - app asks to add message password:


- sender is using FES (so message passwords are supported) - app allows to choose between plain and password-protected message:


- sender doesn't use FES (no message password support) - app warns about non-encrypted message:

ioanmo226 (approved):
👍

PR #1941, opened 2022-11-29
#1830 hide message quote

### LINKED ISSUE #1830: Add ability to hide quote text in thread message
Currently each thread message shows quote of all previous messages in the end of text.
It makes each message cell very long in thread view and more scrolling is needed for finding specific message.

I propose to add `Show history` button instead of thread quote text, so all previous messages texts will be shown only after button tap.

sosnovsky:
This PR implements automatic hiding of quote text

close #1830

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Mail Provider/Message Provider/ProcessedMessage.swift
+++ b/FlowCrypt/Functionality/Mail Provider/Message Provider/ProcessedMessage.swift
@@ -135,6 +135,10 @@ extension ProcessedMessage {
         return (message, quote)
     }
 
+    var fullText: String {
+        [text, quote].compactMap { $0 }.joined(separator: "\n")
+    }
+
     var attributedMessage: NSAttributedString {
         String(text.prefix(maxLength)).attributed(color: type.textColor)
     }
--- a/FlowCrypt/Functionality/Mail Provider/Message Provider/ProcessedMessage.swift
+++ b/FlowCrypt/Functionality/Mail Provider/Message Provider/ProcessedMessage.swift
@@ -11,6 +11,17 @@ import UIKit
 struct ProcessedMessage {
     enum MessageType: Hashable {
         case error(DecryptErr.ErrorType), encrypted, plain
+
+        var textColor: UIColor {
+            switch self {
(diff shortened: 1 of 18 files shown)


appium/tests/helpers/ElementHelper.ts
```diff
@@ -56,6 +56,19 @@ class ElementHelper {
     await element.click();
   }
 
+  static deleteText = async (element: WebdriverIO.Element) => {
```
ioanmo226, line 59: I was confused about `deleteText` name. I thought this function also deletes text for buttons.
Maybe need to rename it to `deleteInputText` or `clearInput` or something like that?

ioanmo226 (commented):
👍 
Looks great. Just a few questions/suggestions.

appium/tests/helpers/ElementHelper.ts
sosnovsky: `clearInput` sounds good 👍

FlowCrypt/Functionality/Mail Provider/Message Provider/ProcessedMessage.swift
```diff
@@ -65,19 +76,75 @@ struct ProcessedMessage {
...
+                } else {
+                    lines.append(lastLine)
+                }
+                break
+            }
+        }
+        let hasQuote = quoteLines.contains { !$0.isEmpty }
+        let quote = hasQuote ? quoteLines.joined(separator: "\n") : nil
+        let message = lines.joined(separator: "\n").trimmingCharacters(in: .whitespacesAndNewlines)
+        return (message, quote)
```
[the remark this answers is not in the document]
sosnovsky: Yes, I checked browser logic before implementing iOS behavior, and also tried some other methods for parsing quotes (like trying to find start index of message quote).
Lines looping looks like the most stable solution (but not the most elegant :))

Regex for such parsing will be quite complicated (as some lines can have spaces before `>`, quote can include empty lines, `On ... wrote:` line can be not present etc)

ioanmo226 (approved):
👍

PR #1973, opened 2022-12-08
#1855 improve classes naming

### LINKED ISSUE #1855: Improve classes naming
`Gateway` is something that stuff passes through. Like an `EmailGateway` - you send an email, it passes through, ends up in another email server.

I guess we could call all of these classes just `...ApiClient` ? `DraftApiClient`, `MessagesApiClient`, `SendMessageApiClient` ? I think that would be the clearest. Can leave it for another PR.

In particular I don't want to call everything `...Service` unless it's some sort of background service that keeps running, like a daemon, which we don't have many off.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/pull/1851#discussion_r988961634_

sosnovsky:
This PR improves classes naming by replacing `Service` with `ApiClient`, `Provider` and `Manager` (depending on class functionality)

close #1855

### To be filled by reviewers


ioanmo226 (approved):
👍

PR #2022, opened 2022-12-28
#1993 feat: added currentSession in SessionManager

### LINKED ISSUE #1993: SessionManager should have currentSession property
`SessionManager` probably should have `currentSession` property, so we don't need to have separate `session` variable

https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/App/AppContext.swift#L17

ioanmo226:
This PR added currentSession in SessionManager

close #1993 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/DataManager/SessionManager.swift
+++ b/FlowCrypt/Functionality/DataManager/SessionManager.swift
@@ -36,7 +36,7 @@ protocol SessionManagerType {
     func switchActiveSessionFor(user: User) throws -> SessionType?
     func startActiveSessionForNextUser() throws -> SessionType?
     func cleanup() throws
-    var currentSession: SessionType? { get }
+    var currentSession: SessionType? { get throws }
 }
 
 final class SessionManager {
@@ -49,7 +49,19 @@ final class SessionManager {
 
     private lazy var logger = Logger.nested(Self.self)
 
-    var currentSession: SessionType?
+    var currentSession: SessionType? {
+        get throws {
+            guard let user = try encryptedStorage.activeUser, let authType = user.authType else {
+                return nil
+            }
+            switch authType {
+            case let .oAuthGmail(token):
+                return .google(user.email, name: user.name, token: token)
+            case .password:
+                return .session(user)
+            }
+        }
+    }
 
     init(
         encryptedStorage: EncryptedStorageType,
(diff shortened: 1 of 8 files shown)


FlowCrypt/Functionality/DataManager/SessionManager.swift
```diff
@@ -48,6 +49,20 @@ final class SessionManager {
 
     private lazy var logger = Logger.nested(Self.self)
 
+    var currentSession: SessionType? {
+        get throws {
```
sosnovsky, line 53: What about setting `currentSession` on `SessionManager` initialization (as we already pass `encryptedStorage` there)?
So we don't need to add `try/catch` for each `currentSession` getter

ioanmo226: Sounds good. Just changed 👍

ioanmo226: @sosnovsky I had to revert changes because if we don't use getter, app won't get current `currentSession` value.
Like when user clicks `sign in with google` button & after auth confirm, it doesn't get session for current logged in user.

sosnovsky: @ioanmo226 probably this can be fixed by updating `currentSession` property in `startSessionFor(session: SessionType)`, can you try it?

ioanmo226: You are right. 👍 
That way works fine.

FlowCrypt/Models/Common/User.swift
```diff
@@ -24,6 +24,18 @@ struct User: Codable, Equatable {
...
     }
+
+    var session: SessionType? {
+        switch authType {
+        case let .oAuthGmail(token):
+            return .google(email, name: name, token: token)
+        case .password:
+            return .session(self)
+        case .none:
+            print("authType is not defined in switchActiveSession")
```
sosnovsky, line 35: Probably this `print` not needed here?

FlowCrypt/Functionality/DataManager/SessionManager.swift
```diff
@@ -60,6 +63,9 @@ final class SessionManager {
         // maybe should instead get user
         self.imap = try imap ?? Imap(user: try encryptedStorage.activeUser ?? User.empty)
         self.encryptedStorage = encryptedStorage
+        if let user = try encryptedStorage.activeUser {
+            currentSession = user.session
+        }
```
sosnovsky, line 68: Let's simplify it a bit:
```swift
self.currentSession = try encryptedStorage.activeUser?.session
```

ioanmo226: Sounds good

sosnovsky (approved):
👍

PR #2028, opened 2022-12-30
#2013 fix: javascript exception after app was in background for long time

### LINKED ISSUE #2013: Javascript exception after app was in background for long time
reported by Tom:
```
happens after the app was running in background for a day
during that day, the app was on conversation view screen
happened when opening existing draft in that conversation
error shows again if i go back and reopen the draft
error goes away when i kill the app

likely related to the webview change. maybe something in the webview gets garbage collected or otherwise disposed to free memory after longer time
```

ioanmo226:
This PR fixed javascript exception after app was in background for long time

close #2013 // if this PR closes an issue

- Difficult to test (Hard to reproduce this issue in semaphoreci)

### To be filled by reviewers


sosnovsky (commented):
Were you able to check if this change fixes bug?
Did you try to use `WKUserScript` for injecting our `Core` js code?

ioanmo226:
Actually, No. But I think this will fix bug. Let me know your opinion.
I couldn't reproduce this bug though.
> Were you able to check if this change fixes bug?

Yes. but I think it's also fine to use `evaluateJavaScript` as long as we can re-setup whenever expired.
Did you try to use WKUserScript for injecting our Core js code?

sosnovsky:
Ok, I'll try to check if bug is fixed over the weekend (as it takes several hours for issue to appear)

tomholub:
>  I couldn't reproduce this bug though.

Try installing all TestFlight versions out there (enterprise / public). As long as you set up both, one of them is sure to fail sooner or later after some non-use.

It's hard to reproduce on a debug build that you work on daily, of course.

No need to do that now, more of an ask for the future, so that we can all get more exposed to these types of errors in production.

tomholub:
> Ok, I'll try to check if bug is fixed over the weekend (as it takes several hours for issue to appear)

Were you able to reproduce it / did you notice it previously?

sosnovsky:
> Were you able to reproduce it / did you notice it previously?

I usually use FlowCrypt from browser extension, so haven't noticed this issue before. Now I installed both (debug and testflight) builds, will wait until evening to re-open app and check if error appears.

And then I'll install new build from this PR to check if error is gone.

sosnovsky:
@ioanmo226 error still appears after app stays in background for several hours

tomholub:
Possible to catch that error, re-initialize the whole web view, and retry last action, all within the `Core` class transparently?

sosnovsky:
> Possible to catch that error, re-initialize the whole web view, and retry last action, all within the `Core` class transparently?

@ioanmo226 tried to implement this solution here: https://github.com/FlowCrypt/flowcrypt-ios/blob/067417f6673a772b68c8c669ce62a408553c9f36/FlowCrypt/Core/Core.swift#L296 but it didn't help. 

My proposal is to try using [WKUserScript](https://developer.apple.com/documentation/webkit/wkuserscript) for injecting `Core` code to `WKWebView`, instead of `evaluateJavascript`, maybe then it won't be unloaded from memory.

tomholub:
The solution there reused an already existing webview, by executing all of the setup code again.

What I was thinking was to destroy the webview that was already open when this error happens, and set up a brand new one.

tomholub:
Of course, your suggestion may work too. Just proposing alternatives.

sosnovsky:
> What I was thinking was to destroy the webview that was already open when this error happens, and set up a brand new one. Then it has no business failing again.

Ah, got it, this one looks simpler than my proposal, let's try it first

sosnovsky:
@ioanmo226 probably your fix failed for me because error code was `5`, not `4`

ioanmo226:
@sosnovsky Maybe re-evaluateJavascript for all WKErrorDomain errors?

sosnovsky:
> @sosnovsky Maybe re-evaluateJavascript for all WKErrorDomain errors?

Yes, let's try it this way, as web view shouldn't return any errors, so resetting it for each `WKErrorDomain` shouldn't affect `Core` functionality.

sosnovsky (approved):
Seems to work well now 👍

PR #2043, opened 2023-01-06
#2013 Fix: Javascript exception after app was in background for long time

### LINKED ISSUE #2013: Javascript exception after app was in background for long time

ioanmo226:
This PR fixed Javascript exception after app was in background for long time

close #2013 // if this PR closes an issue

- Difficult to test (Hard to reproduce in ci environment)

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Core/Core.swift
+++ b/FlowCrypt/Core/Core.swift
@@ -257,7 +257,7 @@ class Core: KeyDecrypter, KeyParser, CoreComposeMessageType {
 
     // MARK: Private calls
     @discardableResult
-    private func call(_ endpoint: String, params: [String: Any?] = [:], data: Data = Data()) async throws -> RawRes {
+    private func call(_ endpoint: String, params: [String: Any?] = [:], data: Data = Data(), retryAttempt: Int = 0) async throws -> RawRes {
         let paramsData = try JSONSerialization.data(withJSONObject: params).toStr()
         let requestData = [UInt8](data)
 
@@ -294,11 +294,12 @@ class Core: KeyDecrypter, KeyParser, CoreComposeMessageType {
 
             return RawRes(json: responseJson ?? Data(), data: responseData)
         } catch {
-            if error._domain == "WKErrorDomain" {
+            // Re-setup the webview only when the retry attempt is less than 3 to avoid entering an infinite loop
+            if error._domain == "WKErrorDomain", retryAttempt < 3 {
                 // Core js code injected using evaluateJavaScript result is removed when app is in background for long time
                 // Need to setup again. https://github.com/FlowCrypt/flowcrypt-ios/issues/2013
                 setupWebView()
-                return try await call(endpoint, params: params, data: data)
+                return try await call(endpoint, params: params, data: data, retryAttempt: retryAttempt + 1)
             }
             throw error
         }
--- a/FlowCrypt/Core/Core.swift
+++ b/FlowCrypt/Core/Core.swift
@@ -51,22 +51,23 @@ class Core: KeyDecrypter, KeyParser, CoreComposeMessageType {
     }
 
     private init() {
-        setupWebView()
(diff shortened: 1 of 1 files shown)


ioanmo226:
Please let me know if you still face issue with this version.
For me, I can't reproduce issue.

sosnovsky:
To reproduce this issue you need several hours to pass:
1) open FlowCrypt on real device
2) open some thread
3) go back to iOS home screen, so app stays in background
4) open app after several hours (it should still have thread screen opened)
5) try to open any thread - infinite spinner

For me it's reproduced when I open app in the morning and then the next time in the evening, so 8-10 hours should be enough.

ioanmo226:
Thank you. Got it

FlowCrypt/Core/Core.swift
```diff
@@ -62,8 +56,13 @@ class Core: KeyDecrypter, KeyParser, CoreComposeMessageType {
 
     // MARK: - Setup
     func setupWebView() {
-        guard let jsFileSrc = getCoreJsFile() else { return }
+        let userController = WKUserContentController()
```
sosnovsky, line 59: Here you updated only initial webView initialization, but we also need to use it when `WKErrorDomain` was caught.

ioanmo226: Well, when `WKErrorDomain` was caught this `setupWebView` function would be called and webview should be re-initialized.

```swift
if error._domain == "WKErrorDomain" {
    // Core js code injected using evaluateJavaScript result is removed when app is in background for long time
    // Need to setup again. https://github.com/FlowCrypt/flowcrypt-ios/issues/2013
    setupWebView()
    return try await call(endpoint, params: params, data: data)
}
```

sosnovsky: Ah, I missed it 👍
I'll install new build and check if bug is gone in the evening

sosnovsky:
@ioanmo226 I re-opened app in the evening and it's just crashed for me on step `5. try to open any thread`

ioanmo226:
It crashed in my side too. Let me dig further

FlowCrypt/Core/Core.swift
```diff
@@ -62,10 +56,17 @@ class Core: KeyDecrypter, KeyParser, CoreComposeMessageType {
 
     // MARK: - Setup
     func setupWebView() {
-        guard let jsFileSrc = getCoreJsFile() else { return }
-
-        let appVersion = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String ?? "[unknown version]"
-        webView.evaluateJavaScript("const APP_VERSION = 'iOS \(appVersion)';\(jsFileSrc)") { _, _ in }
+        DispatchQueue.main.async {
```
ioanmo226, line 59: @sosnovsky  I believe the app crashed because the webview was not set up on the main thread. When I tried to re-setup the webview in thread detail view, it threw an error stating that the webview setup code must be executed on the main thread.

ioanmo226: Waiting for my testflight build to throw error and new build to work fine.
Will let you know once I finish testing so that you can review

sosnovsky: Sounds good 👍
Maybe we can use `@MainActor` here?

ioanmo226: Sure. Maybe you need to build & run this PR in device and check after 8-10 hours if app crashes?

sosnovsky (approved):
Works well for me after this fix 👍

PR #2044, opened 2023-01-06
#2012 fix: Draft error handling

### LINKED ISSUE #2012: Problems with draft error handling
1) should not lose focus and keyboard  when error toast displays - disruptive to user workflow
2) what does that particular error mean? should think of how to address it in particular
3) should not keep trying forever for any of these errors. eg if it tried 20 times then could just drop it
4) use exponential backoff. currently keeps trying every 3 seconds
5) another issue. it suddenly showed many drafts after i sent the last one. when i go back to all conversations and open again, they are gone. so it’s just some rendering issue maybe, too many times and didn’t go away when actually sent?

image

ioanmo226:
This PR fixed draft error handlings.

close #2012 // if this PR closes an issue

- Difficult to test (Hard to reproduce in ci environment)

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Services/Compose Message Helper/ComposeMessageHelper.swift
+++ b/FlowCrypt/Functionality/Services/Compose Message Helper/ComposeMessageHelper.swift
@@ -254,7 +254,7 @@ final class ComposeMessageHelper {
         }
     }
 
-    func saveDraft(message: SendableMsg, threadId: String?, shouldEncrypt: Bool) async throws {
+    func saveDraft(message: SendableMsg, threadId: String?, shouldEncrypt: Bool, forceCreate: Bool = false) async throws {
         if let saveDraftTask {
             _ = try await saveDraftTask.value
         }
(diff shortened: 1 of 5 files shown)


ioanmo226:
@sosnovsky This one is ready for review

FlowCrypt/Functionality/Services/Compose Message Helper/ComposeMessageHelper.swift
```diff
@@ -268,16 +268,17 @@ final class ComposeMessageHelper {
                     fmt: shouldEncrypt ? .encryptInline : .plain
                 )
 
-                let threadId = self.messageIdentifier?.threadId?.stringId ?? threadId
+                let threadId = forceCreate ? nil : (self.messageIdentifier?.threadId?.stringId ?? threadId)
```
sosnovsky, line 271: I think `threadId` shouldn't be set to `nil` here, as in this case draft won't be added to current thread

ioanmo226: Well, forceCreate will be only called when `requested entity not found` error occurred.
When that thread/draft is deleted from Gmail side, currentThreadId should be invalid and need to create new one.

sosnovsky: I re-checked - if `threadId` is invalid then no error occurs and draft is saved to new thread.
app receives `requested entity not found` error only when draft was deleted, so `threadId` can be left as it is here

ioanmo226: Well, you can follow this way to check.
First create draft from iOS app and go back to draft screen
Open that draft again and delete this draft from browser(Gmail) while you are on compose screen.
Try to edit subject/text and api will now return `request entity not found error`
If we don't set threadId to nil in this case, this draft won't be saved and `request entity not found` error will occur permanentely.

sosnovsky: Ah, I see - I was testing with existing thread, but when creating new draft then this error occurs.
Let's leave it for now and see how it works, as it probably should be a quite rare case.

sosnovsky (approved):

FlowCrypt/Functionality/Services/Compose Message Helper/ComposeMessageHelper.swift
ioanmo226: ok

PR #2068, opened 2023-01-23
#2063 Fix: Double encrypt attachment issue when forwarding

### LINKED ISSUE #2063: Attachment is double-encrypted in forwarded message
While forwarding an email (with attachment) the attachment becomes "over" encrypted and has a .pgp.pgp extension, which makes it impossible to decrypt by recipient.

ioanmo226:
This PR fixed double encrypt attachment issue when forwarding

close #2063 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


appium/tests/specs/mock/composeEmail/CheckAttachmentAfterForward.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+  MailFolderScreen,
+  SetupKeyScreen,
+  SplashScreen
+} from '../../../screenobjects/all-screens';
+
+describe('COMPOSE EMAIL: ', () => {
+
+  it('check attachment after forward', async () => {
+    const mockApi = new MockApi();
+    const subject = 'Test attachment after forward';
```
sosnovsky, line 14: Let's use existing mock email with subject `email with text attachment`, as it has almost the same content, so we won't have several mock emails with similar content.

FlowCrypt/Controllers/Threads/ThreadDetailsViewController+TapActions.swift
```diff
@@ -153,6 +159,28 @@ extension ThreadDetailsViewController {
...
+                )
+            }
+            if updatedAttachment.isEncrypted {
+                handleFetchProgress(state: .decrypt)
+                return try await messageHelper.decrypt(
+                    attachment: updatedAttachment,
+                    userEmail: appContext.user.email
+                )
+            }
+            return updatedAttachment
```
sosnovsky, line 181: We have almost the same code in `getAttachment(at indexPath: IndexPath)`, may be it's possible to move it to separate method `fetchAndDecrypt(attachment: MessageAttachment)` to avoid duplication?

Then it'll be just 
```swift
let decryptedAttachments = try await attachments.asyncMap { fetchAndDecrypt(attachment: $0)
```

FlowCrypt/Controllers/Threads/ThreadDetailsViewController+TapActions.swift
```diff
@@ -153,6 +159,28 @@ extension ThreadDetailsViewController {
         let attachments = quoteType == .forward
             ? input.processedMessage?.attachments ?? []
             : []
+        showSpinner()
+        let decryptedAttachments = try await attachments.asyncMap { (attachment: MessageAttachment) -> MessageAttachment in
```
sosnovsky, line 163: Probably `MessageAttachment` declaration not needed here, as swift compiler will get it automatically

appium/tests/specs/mock/composeEmail/CheckAttachmentAfterForward.spec.ts
ioanmo226: Sounds good

PR #2270, opened 2023-07-07
#2269 Added prettier

### LINKED ISSUE #2269: Implement Prettier
Implement a prettier feature (like Flowcrypt browser extension) in `appium` and `core`.
Also, turn the autoSave setting to true, so that TypeScript code will be auto-formatted.

@sosnovsky What do you think?

ioanmo226:
This PR added prettier

close #2269 // if this PR closes an issue

### To be filled by reviewers


sosnovsky (commented):
Looks good, but let's also update compiled Core js file in app bundle, to check if everything works well.

It's done by running `npm run build` in `/Core` directory.

ioanmo226:
Sounds good!

sosnovsky (commented):
I missed it on the first review, but let's update `.prettierignore` to also include:
- `Core/build/` - no need to prettify built files
- `Core/source/lib` - these are 3rd-party libs, so let's leave their original code

And `Core/.jsbeautifyrc` file isn't needed anymore

ioanmo226:
I think there is no need to include `Core/build/` in the `.prettierignore` file because the `build` directory is already specified in the `.gitignore` file within the `Core` folder. Additionally, `prettier` prioritizes the rules specified in `.gitignore` before considering `.prettierignore`.

sosnovsky:
> I think there is no need to include `Core/build/` in the `.prettierignore` file because the `build` directory is already specified in the `.gitignore` file within the `Core` folder. Additionally, `prettier` prioritizes the rules specified in `.gitignore` before considering `.prettierignore`.

Sounds good 👍

sosnovsky (commented):
All good now, just need to copy `Core/lib` files from `master`, as here they were formatted without updated `.prettierignore`

sosnovsky (approved):
👍

PR #2291, opened 2023-07-21
#2099 fix: email showing as plain text issue

### LINKED ISSUE #2099: Some emails are showing just plain html instead of text
Currently app should strip html code and render just plain text for html emails. But it fails for some emails and showing just plain html, which makes it quite difficult to read message content.

ioanmo226:
This PR fixed email showing as plain text issue

close #2099 // if this PR closes an issue

image

- Difficult to test (Hard to test in CI env)

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Functionality/Mail Provider/Message Provider/Gmail+MessageExtension.swift
+++ b/FlowCrypt/Functionality/Mail Provider/Message Provider/Gmail+MessageExtension.swift
@@ -103,7 +103,7 @@ private extension GTLRGmail_Message {
 
     func parseMessageBody() -> MessageBody {
         let html = body(type: .textHtml)
-        let text = body(type: .textPlain) ?? html?.removingHtmlTags() ?? ""
+        let text = (body(type: .textPlain) ?? "").removingHtmlTags() ?? ""
         let bodyAttachment = text.isEmpty ? parseBodyAttachment() : nil
 
         return MessageBody(text: text, html: html, attachment: bodyAttachment)
--- a/FlowCrypt/Functionality/Mail Provider/Message Provider/Gmail+MessageExtension.swift
+++ b/FlowCrypt/Functionality/Mail Provider/Message Provider/Gmail+MessageExtension.swift
@@ -103,9 +103,9 @@ private extension GTLRGmail_Message {
 
     func parseMessageBody() -> MessageBody {
         let html = body(type: .textHtml)
-        let text = body(type: .textPlain) ?? html?.removingHtmlTags() ?? ""
+        var text = body(type: .textPlain) ?? html ?? ""
+        text = text.isHTMLString() ? (text.removingHtmlTags() ?? "") : text
         let bodyAttachment = text.isEmpty ? parseBodyAttachment() : nil
-
         return MessageBody(text: text, html: html, attachment: bodyAttachment)
     }
 
--- a/FlowCrypt/Functionality/Mail Provider/Message Provider/Gmail+MessageExtension.swift
+++ b/FlowCrypt/Functionality/Mail Provider/Message Provider/Gmail+MessageExtension.swift
@@ -104,7 +104,9 @@ private extension GTLRGmail_Message {
     func parseMessageBody() -> MessageBody {
         let html = body(type: .textHtml)
         var text = body(type: .textPlain) ?? html ?? ""
-        text = text.isHTMLString() ? (text.removingHtmlTags() ?? "") : text
+        if text.isHTMLString() {
+            text = text.removingHtmlTags() ?? ""
+        }
         let bodyAttachment = text.isEmpty ? parseBodyAttachment() : nil
         return MessageBody(text: text, html: html, attachment: bodyAttachment)
     }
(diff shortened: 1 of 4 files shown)


FlowCrypt/Functionality/Mail Provider/Message Provider/Gmail+MessageExtension.swift
```diff
@@ -103,7 +103,7 @@ private extension GTLRGmail_Message {
 
     func parseMessageBody() -> MessageBody {
         let html = body(type: .textHtml)
-        let text = body(type: .textPlain) ?? html?.removingHtmlTags() ?? ""
+        let text = (body(type: .textPlain) ?? "").removingHtmlTags() ?? ""
```
ioanmo226, line 106: @sosnovsky, Initially, I suspected that the HTML email was displaying as plain text because NSAttributedString initialization didn't recognize the xmlns attribute. 

However, after thorough evaluation, I discovered that my initial assumption was incorrect. The real issue is that `removingHtmlTags` function wasn't invoked for this particular instance.

What's your opinion on this fix?

sosnovsky: With this fix we'll probably lose message text if there is no plain text body as it'll be converted to `""`?

ioanmo226: Sorry, it was wrong fix. just pushed new fix

ioanmo226:
@sosnovsky Please re-review

FlowCrypt/Functionality/Mail Provider/Message Provider/ProcessedMessage.swift
```diff
@@ -96,6 +96,7 @@ extension ProcessedMessage {
     ) {
         self.message = message
         (self.text, self.quote) = Self.parseQuote(text: text)
+        self.text = self.text.removingHtmlTags() ?? self.text
```
sosnovsky, line 99: probably we can move `removingHtmlTags` to `parseQuote`, so it'll return text without tags and we won't need to remove them manually each time

ioanmo226: Sounds good!
I also removed html tags for `quote` too. What do you think?

sosnovsky: We don't need `?? ""` for quote text, as in this case quote `...` button becomes visible for all messages (even without quotes)

ioanmo226: Yep!. Missed it

sosnovsky (commented):
I re-checked how it works, and looks like it was a bad idea to run `removingHtmlTags` for all texts. It works well for example message from this issue, but when you open message with only plain text body (for example, encrypted message sent from FlowCrypt) - it'll wrap all message into 1 line, which is not correct behavior.

Probably source of initial issue is that FlowCrypt thinks that initial message has only plain text body (while it has only html text body), and that's why it doesn't run `removingHtmlTags` for it. So we should improve parsing of message body parts to correctly handle html-only messages. Then current code `let text = body(type: .textPlain) ?? html?.removingHtmlTags() ?? ""` should work well

sosnovsky:
`master` branch:

Screenshot 2023-07-21 at 21 57 51


this PR:

Screenshot 2023-07-21 at 21 58 37

ioanmo226:
Nice catch. Let me have a re-look early next week.

ioanmo226:
@sosnovsky Please re-check

sosnovsky (commented):
New approach looks good, but it doesn't work for another html message.
I sent this message to you, can you please check it, probably `isHtmlString` regex needs some correction

ioanmo226:
Ok, Let me check

ioanmo226:
Just fixed it.
The problem was I didn't handle the `DocType` when it had attributes. Thanks for your detailed review, by the way!

FlowCryptCommon/Extensions/StringExtensions.swift
```diff
@@ -95,6 +95,13 @@ public extension String {
         ).string
     }
 
+    func isHTMLString() -> Bool {
+        if let _ = self.range(of: "^\\s*(<!doctype html[\\s\\S]*>)?\\s*<[a-z][\\s\\S]*>\\s*$", options: [.regularExpression, .caseInsensitive]) {
+            return true
+        }
+        return false
```
sosnovsky, line 102: let's simplify it to be just
```swift
func isHTMLString() -> Bool {
  self.range(...) != nil
}
```

sosnovsky (commented):
Works great now, thanks for all fixes 👍

FlowCryptCommon/Extensions/StringExtensions.swift
ioanmo226: Sounds good!

sosnovsky (approved):

PR #2298, opened 2023-07-28
#418 Updated recipient selection color

### LINKED ISSUE #418: tapping delete key on keyboard to just added recipient wrongly turns it gray
Reproduce:

1) open message compose view
2) add a contact like `human@flowcrypt.com`, then space
3) it turns into a green bubble
4) tap the `del` key on keyboard above `return` key

expected: the last recipient gets removed

actual: it turns gray. It only gets removed when I tap `del` one more time

### LINKED ISSUE #2304: Disable user interaction when recipient is selected
Plus I think we should remove typing cursor when recipient is selected, as it's not common to have something selected and active cursor simultaneously (you can check how it works in iOS Messages or Mail app)

_Originally posted by @sosnovsky in https://github.com/FlowCrypt/flowcrypt-ios/issues/2298#issuecomment-1655512682_

ioanmo226:
This PR updated recipient selection color

close #418 // if this PR closes an issue
close #2304 // if this PR closes an issue

image
image

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+Nodes.swift
+++ b/FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+Nodes.swift
@@ -234,9 +234,20 @@ extension ComposeViewController {
             case let .imageTap(indexPath):
                 self.handleRecipientAction(with: indexPath, type: type)
             case let .select(indexPath, sender):
+                self.toggleRecipientTextField(isEnabled: false)
                 self.handleRecipientSelection(with: indexPath, type: type)
                 self.displayRecipientPopOver(with: indexPath, type: type, sender: sender)
             }
+        }
+    }
+
+    func toggleRecipientTextField(isEnabled: Bool) {
+        let types: [RecipientType] = [.to, .cc, .bcc]
+        for type in types {
+            self.recipientsTextField(type: type)?.isUserInteractionEnabled = isEnabled
+        }
+        if let type = selectedRecipientType, let textField = recipientsTextField(type: type) {
+            textField.becomeFirstResponder()
         }
     }
 
--- a/FlowCrypt/Controllers/Compose/ComposeRecipientPopupViewController.swift
+++ b/FlowCrypt/Controllers/Compose/ComposeRecipientPopupViewController.swift
@@ -13,6 +13,7 @@ import FlowCryptUI
 protocol ComposeRecipientPopupViewControllerProtocol {
     func removeRecipient(email: String, type: RecipientType)
     func editRecipient(email: String, type: RecipientType)
+    func enableRecipientEditing()
 }
 
 /**
@@ -97,6 +98,7 @@ extension ComposeRecipientPopupViewController: ASTableDelegate, ASTableDataSourc
         default:
             break
         }
+        self.delegate?.enableRecipientEditing()
     }
 }
 
--- a/FlowCrypt/Controllers/Compose/ComposeRecipientPopupViewController.swift
+++ b/FlowCrypt/Controllers/Compose/ComposeRecipientPopupViewController.swift
@@ -14,6 +14,7 @@ protocol ComposeRecipientPopupViewControllerProtocol {
     func removeRecipient(email: String, type: RecipientType)
     func editRecipient(email: String, type: RecipientType)
     func enableRecipientEditing()
+    func deselectRecipients(type: RecipientType)
 }
 
 /**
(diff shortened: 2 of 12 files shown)


ioanmo226:
Let me know if you have better suggestion.

sosnovsky:
It's still not very clear that recipient is selected - looks like just some another state, I would think that recipient wasn't selected and will try to select it by tapping.

What about just reducing opacity of selected recipient to 0.6:

Screenshot 2023-07-28 at 14 09 02

Then it'll still be possible to see current state of recipient (key found, no key etc), but reduced opacity will show that it is currently selected. Plus I think we should remove typing cursor when recipient is selected, as it's not common to have something selected and active cursor simultaneously (you can check how it works in iOS Messages or Mail app)

sosnovsky:
> Cool thought! But here's the thing, to me, it feels like 100% opacity gives off a "selected" vibe, while 50% opacity seems more "unselected", What do you think?

Yeah, I agree on this but haven't come up with something better yet :) Probably it'll still be more clear than just switching recipient color to blue.
Among other ideas there were making font bold or changing border color and width, but not sure if it work well too.

ioanmo226:
I agree. Let me proceed that way

ioanmo226:
Setting opacity to 50% for `keyNotFoundStateContext` doesn't seem to work well. (As it already has alpha value of 0.1, doesn't have much visual difference even we set it to 0.05)
Is this fine?

image

ioanmo226:
Ok

ioanmo226:
@sosnovsky Ready for review

FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+Nodes.swift
```diff
@@ -234,12 +234,23 @@ extension ComposeViewController {
...
 
+    func toggleRecipientTextField(isEnabled: Bool) {
+        let types: [RecipientType] = [.to, .cc, .bcc]
+        for type in types {
+            self.recipientsTextField(type: type)?.isUserInteractionEnabled = isEnabled
+        }
+        if let type = selectedRecipientType, let textField = recipientsTextField(type: type) {
+            textField.becomeFirstResponder()
+        }
+    }
```
sosnovsky, line 252: I tested and didn't notice any change of behavior - cursor is still there and I can type new recipient.

But I've just found that if we'll hide cursor, as described in https://github.com/FlowCrypt/flowcrypt-ios/issues/2304, then users won't be able to tap `backspace` for removing selected recipient. So let's leave previous behaviour, I think `toggleRecipientTextField(isEnabled: Bool)` isn't needed

ioanmo226: Could you try to clean & build. It might be because of cache.
If it still happens after clean, please let me know reproduce steps.
> I tested and didn't notice any change of behavior - cursor is still there and I can type new recipient.


I think it's good idea to prevent typing when user selected recipient. User can delete recipient by clicking `Remove` button. Gmail app also prevents typing when recipient is selected.

>  So let's leave previous behaviour

sosnovsky: I thought maybe cursor isn't hiding only in simulator, but re-tested on device - same behavior.

To reproduce:
1) open compose screen
2) enter recipient
3) wait until public key is found (or not found)
4) click backspace
5) recipient became selected and background color has changed (which is expected), but typing cursor is still visible
6) type another recipient (first recipient still remains selected)
7) after you've entered second recipient, both recipients will try to search for public keys:


8) second recipient will finish searching, while first one will be in searching state indefinitely:

sosnovsky: > I think it's good idea to prevent typing when user selected recipient. User can delete recipient by clicking Remove button. Gmail app also prevents typing when recipient is selected.

Ah, you probably mean case when there is popup with recipient actions, while I mean case when recipient is selected by tapping backspace (without popup).
Then here we should just deselect selected recipient when user starts typing, the same as Gmail app

ioanmo226: Aha. Got it

ioanmo226: I thought you were talking about state when there is a popup with recipient actions.
However fixed above issue. Please check

sosnovsky (approved):
Works well now 👍

PR #2313, opened 2023-08-07
#291 fix showspinner

### LINKED ISSUE #291: Fix showSpinner()
showSpinner() is an extension method to show MBProgressHUD on a view. 
Spinner dispatched on main queue with animation and without any completion on when animation ends. 
This can lead to a bug that HUD is not showed if any other operation scheduled

see `// TODO: - fix for spinner`

ioanmo226:
This PR fixed spinner and uses [ProgressHud](https://github.com/relatedcode/ProgressHUD/)

close #291 // if this PR closes an issue

### To be filled by reviewers


ioanmo226:
Yeah, I think it's expected behavior.
When we try to show side menu, we show spinner and then once load (fetching label list from api/cache)finishes we hide spinner.

ioanmo226:
Maybe we should only show&hide spinner when `force` load flag is true?

sosnovsky:
Ah, I just checked App Store build and it has spinner on side menu too, but because of quicker animation it's almost not visible. Can we change duration of animation of spinner appearance or maybe completely disable it for spinner in side menu?

> Maybe we should only show&hide spinner when force load flag is true?

Yes, let's show it only when `force` is true

sosnovsky (approved):
👍

PR #2349, opened 2023-09-06
#2348 ssl pinning improvements

### LINKED ISSUE #2348: SSL pinning improvements
> I think NSPinnedDomains is kinda rigid—lacks customization and is a not easy to debug.

For debugging you still need to use some 3rd-party tools, as you can't check SSL pinning implementation directly in Xcode. So I think it's not a big disadvantage.

I was able to test it on Mac using [Proxyman](https://proxyman.io) (guide about SSL Proxying - https://tanaschita.com/20220704-quick-guide-on-proxyman/). Current implementation works well, but error message is just `cancelled` instead of `SSL certificate mismatch`

Let's try to use `NSPinnedDomains` and fix this error message, maybe it'll be even easier with `NSPinnedDomains`

_Originally posted by @sosnovsky in https://github.com/FlowCrypt/flowcrypt-ios/issues/2336#issuecomment-1708367693_

ioanmo226:
This PR improved SSL pinning with NSPinnedDomains

close #2348 // if this PR closes an issue

### To be filled by reviewers


FlowCrypt.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
sosnovsky, line 1: please update packages to the latest versions, as they are on `master`

FlowCrypt/Controllers/Settings/Settings List/SettingsViewController.swift
```diff
@@ -116,14 +116,12 @@ extension SettingsViewController {
...
         switch setting {
         case .keys:
-            showSpinner()
             do {
                 viewController = try await KeySettingsViewController(appContext: appContext)
             } catch {
                 viewController = nil
                 showAlert(message: error.localizedDescription)
             }
-            hideSpinner()
```
sosnovsky, line 126: are these spinners not needed anymore?

FlowCrypt/Info.plist
```diff
@@ -79,5 +79,24 @@
...
 	</array>
+    <key>NSAppTransportSecurity</key>
+    <dict>
+        <key>NSPinnedDomains</key>
+        <dict>
+            <key>flowcrypt.com</key>
+            <dict>
+                <key>NSIncludesSubdomains</key>
+                <true/>
+                <key>NSPinnedLeafIdentities</key>
```
sosnovsky, line 90: I tried to send password-protected messages, but there is no such option on compose screen because `isUsingFes` is `false`.

I think it happens because app can't fetch correct configuration as here for root certificates we should use `NSPinnedCAIdentities` instead of `NSPinnedLeafIdentities`.

FlowCrypt/Controllers/Settings/Settings List/SettingsViewController.swift
ioanmo226: Well, spinner usually appears and disappears in 0.3-0.5 seconds and someone might think it's UI tweak issue.
Let me add spinner if you want so.

sosnovsky: Ah, in this case no need to add spinner here. I thought maybe it's related to some delays which appeared with SSL pinning

FlowCrypt/Info.plist
```diff
@@ -79,5 +79,23 @@
...
+		<dict>
+			<key>flowcrypt.com</key>
+			<dict>
+				<key>NSIncludesSubdomains</key>
+				<true/>
+				<key>NSPinnedCAIdentities</key>
+				<array>
+					<dict>
+						<key>SPKI-SHA256-BASE64</key>
+						<string>ivAzLVz0x7Rd0hhHf6M6pwLHZoPsEZrlqmw30tbbo5M=</string>
```
sosnovsky, line 94: I tried to send password-protected, but it still fails - recipient UI just stucks in this state:


But I found that it caused by wrong value in this config, how did you get it?

I generated this digest from ISRG Root X1 certificate (using commands from https://www.guardsquare.com/blog/leveraging-infoplist-based-certificate-pinning-ios-and-making-its-shortcomings guide) and got `C5+lpZ7tcVwmwQIMcRtPbsQtWLABXhQzejna0wHFr8M=`.
After setting this value app works well, I can send password-protected messages and get correct error when using SSL proxying:

ioanmo226: I used below command to get value.
But hmm.. with above value, app still shows red recipint in my side.
https://blogs.sap.com/2021/05/28/certificate-pinning-on-ios-14/#:~:text=on%20my%20Mac%3A-,openssl,-s_client%20%2Dshowcerts

image

sosnovsky: I tried to use command from your link https://blogs.sap.com/2021/05/28/certificate-pinning-on-ios-14/#:~:text=on%20my%20Mac%3A-,openssl,-s_client%20%2Dshowcerts, and it gives me the same `ivAzLVz0x7Rd0hhHf6M6pwLHZoPsEZrlqmw30tbbo5M=` value as we currently have in config (and it makes recipients red as on screenshot)

But if I go through guide at  https://www.guardsquare.com/blog/leveraging-infoplist-based-certificate-pinning-ios-and-making-its-shortcomings - export ISRG_Root_X1 from Keychain, extract public key and generate digest, then I get `C5+lpZ7tcVwmwQIMcRtPbsQtWLABXhQzejna0wHFr8M=` value, which fixes mentioned issue and now I can send password-protected messages. Can you try it on your side?

ioanmo226: I also get same value using above article you provided (C5+lpZ7tc....)

sosnovsky: It's strange, I tested email from your screenshot after setting `C5+lpZ7tcVwmwQIMcRtPbsQtWLABXhQzejna0wHFr8M=` value to config, and it works well:



I'll create new test build with updated `SPKI-SHA256-BASE64` value, so we can test it on real build and Mart too

sosnovsky: @martgil @ioanmo226 I uploaded new build 1.2.4 (2) to Testflight with updated SSL pinning, can you please check if you're able to send password-protected messages with it?

Also after enabling SSL proxying (using some 3rd-party tool, like Proxyman or HTTP Catcher) sending of password-protected message should fail with such error:

martgil: Indeed, that should be the expected outcome - I also get the same error on version 1.2.4 (2):

ioanmo226: @martgil I can see that from your screenshot, your test@gmail.com account is marked as red.
And other recipient is marked as gray(no pub key)?

martgil: Yes, @ioanmo226 - that is correct. I have sent you a short video demonstration of how it looks on my end via email. Please kindly check it and let me know of any other test I should perform.

sosnovsky: Ah, I think I found why recipient becomes gray for me and red for you - I enabled SSL proxying only for `*.flowcrypt.com` domains, while you have it enabled for all domains and app fails to load recipient info from `gmail.com`. Then we should check which error is received from `gmail.com` and handle it too

ioanmo226: Could you please explain briefly how we could handle such case for `gmail.com `domain (Also other domains like outlook.com, zoho.com, etc )?

sosnovsky: Seems like error happens at [fetchRemoteUpdateLocal](https://github.com/FlowCrypt/flowcrypt-ios/blob/12791b26905181867b6b3035aa05d3460c31cf11/FlowCrypt/Controllers/Compose/Extensions/ComposeViewController%2BActionHandling.swift#L48) - app fails to fetch public key from attester with error `An SSL error has occurred and a secure connection to the server cannot be made.\n\nGET https://flowcrypt.com/attester/pub/...`, but no error modal is shown.
Then we can just add error modal for these cases

ioanmo226: OK. Got it

ioanmo226: Fixed. Please check

sosnovsky: Works well now 👍

sosnovsky (approved):

PR #2359, opened 2023-09-15
#2350 Added icons to Inbox items

### LINKED ISSUE #2350: Add Encrypted & Attachment Icon  to Thread List View on iOS
The Android app currently displays an encrypted icon next to threads that are encrypted.
This is a nice UI feature that helps users quickly identify secure threads.
It would be beneficial to implement this feature on the iOS app as well.

Android app screenshot
Screenshot_2

ioanmo226:
This PR added encrypted, public key, attachment icons to Inbox items

close #2350 // if this PR closes an issue

image

### To be filled by reviewers


FlowCryptUI/Cell Nodes/InboxCellNode.swift
```diff
@@ -22,24 +22,38 @@ public final class InboxCellNode: CellNode {
...
             self.accessibilityidentifier = accessibilityidentifier
             self.badgeText = badgeText
+            self.isEncrypted = isEncrypted
+            self.hasAttachment = hasAttachment
+            self.hasPublicKey = hasPublicKey
         }
     }
 
+    private enum Constants {
+        static let iconSize: CGFloat = 25
```
sosnovsky, line 53: 25px seems too big, let's decrease size to 20px, should look a bit better

FlowCryptUI/Cell Nodes/InboxCellNode.swift
```diff
@@ -137,52 +177,62 @@ public final class InboxCellNode: CellNode {
...
-            nameLocationStack.children = [emailElement, messageStack]
-        } else {
-            nameLocationStack.children = [emailElement, messageNode]
-        }
+        let iconsSpec = ASStackLayoutSpec.horizontal()
+        iconsSpec.spacing = 3
+        iconsSpec.children = [
+            input.hasAttachment ? attachmentIcon : nil,
+            input.hasPublicKey ? publicKeyIcon : nil,
+            input.isEncrypted ? encryptedIcon : nil
```
sosnovsky, line 208: Screenshot 2023-09-18 at 12 48 18

Looks like there are too many icons when thread contains attachments and public key.

Let's show only `public key` icon when thread contains only public key, and only `attachment` icon when thread contains attachments and public keys

FlowCrypt/Functionality/Mail Provider/Threads/MessagesThreadApiClient.swift
```diff
@@ -60,7 +60,15 @@ extension GmailService: MessagesThreadApiClient {
 
     func fetchThread(identifier: String, path: String) async throws -> MessageThread {
         let query = GTLRGmailQuery_UsersThreadsGet.query(withUserId: .me, identifier: identifier)
-        query.format = kGTLRGmailFormatMetadata
+        query.format = kGTLRGmailFormatFull
```
sosnovsky, line 63: I think we used `metadata` here for improving performance, as fetching all messages for all threads can take too much time when there are several large threads with many messages.

Maybe there is some other way to check if thread messages contain attachments?

sosnovsky (commented):
Nice addition to UI 👍

FlowCrypt/Functionality/Mail Provider/Threads/MessagesThreadApiClient.swift
ioanmo226: Well, maybe...
But keep in mind that we also need to check if attachment is public key or not.

sosnovsky: We can simplify and not differentiate public keys and other attachments, if it'll improve performance

sosnovsky: By the way, how it's done on Android - does it fetch `full` format for threads?

ioanmo226: Need to check

ioanmo226: As far as i checked, in Android app it also fetches `full` format from Gmail API.
@DenBond7 Please correct me if I'm wrong.

DenBond7: > By the way, how it's done on Android - does it fetch full format for threads?

Just in case, Android doesn't support threads yet. 

> As far as i checked, in Android app it also fetches full format from Gmail API.
@DenBond7 Please correct me if I'm wrong.

Yes, we use `full` format to make some requests. But we don't load a full message. Please look at this code

https://github.com/FlowCrypt/flowcrypt-android/blob/1f653a0b24844ceeae0dfc2f6a4c273e200ff4bf/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/gmail/GmailApiHelper.kt#L115

ioanmo226: But gmail threads api only accept `format` & `metadataHeaders` params.

https://github.com/FlowCrypt/flowcrypt-android/blob/1f653a0b24844ceeae0dfc2f6a4c273e200ff4bf/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/gmail/GmailApiHelper.kt#L787C11-L787C11

image

ioanmo226: Does Android `GmailApiService` use another api or use some inner handling of fields param?

DenBond7: For now, Android uses `users.messages` and doesn't use `users.threads`

ioanmo226: Yep, you are right.
But from gmail api documentation, `users.messsages` api also doesn't accept `fields` api.

sosnovsky: Probably we can use data received with `metadata` format to determine message encryption - if it's `snippet` property starts with `-----BEGIN PGP MESSAGE-----` then message is encrypted.

DenBond7: > Yep, you are right. But from gmail api documentation, `users.messsages` api also doesn't accept `fields` api.

[this one ](https://developers.google.com/gmail/api/reference/rest/v1/users.messages/get) accepts

image

DenBond7: > Probably we can use data received with `metadata` format to determine message encryption - if it's `snippet` property starts with `-----BEGIN PGP MESSAGE-----` then message is encrypted.

to be honest, without loading the full message it's not possible to determine 100% if the message is encrypted or not. It was true at least for IMAP protocol. we have similar functionality on Android and it doesn't work well. From time to time we receive feedback from users that some message is not marked as encrypted.

`Gmail API` has `snippet`, but this info can be fetched only via `users.messages.get`. We can use [Batching Requests](https://developers.google.com/gmail/api/guides/batch) to improve the performance when we load the message list and fetch details about each message. But I'm still not sure that this will cover all cases to determine if the message is encrypted or not(need to investigate it).

sosnovsky: Yeah, it's more complicated on iOS as inbox screen can contain threads with a lot of messages, while Android loads a fixed number of messages. Documentation for batch requests says `Larger batch sizes are likely to trigger rate limiting. Sending batches larger than 50 requests is not recommended.`, so in some cases we won't be able to load messages for all displayed threads.

Current Ioan's implementation which uses `kGTLRGmailFormatFull` seems to be the simplest for this functionality, it'll be ok to use it, I just need to re-check how it works when user have large threads, as I thought we switched to `kGTLRGmailFormatMetadata` because of such issues.

sosnovsky: I tried to use it with some big messages and large threads - works well, didn't notice any lags.
Then let's leave it for now

ioanmo226: Ok,

FlowCryptUI/Cell Nodes/InboxCellNode.swift
ioanmo226: Well, what if email contains public key attachment and other attachment files?

ioanmo226: We need show both icons (public key icon and attachment icon) for only this case?

sosnovsky: No, let's show only `attachment` icon in this case. So `public key` icon will be shown when thread contains only public key attachment

ioanmo226: OK

FlowCrypt/Controllers/Inbox/InboxItem.swift
```diff
@@ -22,6 +22,10 @@ struct InboxItem: Equatable {
         return messages.contains(where: \.attachments.isNotEmpty)
     }
 
+    var onlyHavePublicKey: Bool {
+        return messages.count == 1 && messages.contains(where: \.hasPublicKeyAttachment)
```
sosnovsky, line 26: why `messages.count` should equal to 1?

for example, user sent me public key - app shows `public key attachment` icon, then I replied to him and now app shows `attachment` icon. I think it should show `public key attachment` icon for any number of messages

ioanmo226: > No, let's show only attachment icon in this case. So public key icon will be shown when thread contains only public key attachment

ioanmo226: I added that logic because of this comment. Maybe I understood something wrong?

sosnovsky: Yeah, I meant that we should show `public key attachment` icon only for cases when thread contains only public key attachment (or public key attachments), but if it has public key attachment and some other attachments - then show `attachment` icon, not depending on number of messages in thread.

ioanmo226: Ohh... I was confused because I thought I added logic to check if message contains only public key but now realized I only checked messages count. Let me fix it

sosnovsky (approved):
Works great now 👍

PR #2379, opened 2023-09-28
#2366 Reload inbox view when index was not found

### LINKED ISSUE #2366: archive doesn't remove item from inbox until refreshed - sometimes
swiping item in inbox to archive it works well

clicking the archive button on opened message sometimes doesn't actually remove the item

only happened once, couldn't reproduce again

very likely afailed nil check somewhere. in which case, instead of doing nothing, whole inbox could be refreshed

ioanmo226:
This PR reloads inbox view when index was not found.

close #2366 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Inbox/InboxViewController+TableView.swift
+++ b/FlowCrypt/Controllers/Inbox/InboxViewController+TableView.swift
@@ -402,6 +402,9 @@ extension InboxViewController {
         hideSpinner()
 
         guard let indexToUpdate = getUpdatedIndex(for: inboxItem) else {
+            // Just reload data when index was not found in rare cases.
+            // https://github.com/FlowCrypt/flowcrypt-ios/issues/2366
+            tableNode.reloadData()
             return
         }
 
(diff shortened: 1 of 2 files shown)


FlowCrypt/Controllers/Inbox/InboxViewController+TableView.swift
```diff
@@ -402,6 +402,9 @@ extension InboxViewController {
         hideSpinner()
 
         guard let indexToUpdate = getUpdatedIndex(for: inboxItem) else {
+            // Just reload data when index was not found in rare cases.
+            // https://github.com/FlowCrypt/flowcrypt-ios/issues/2366
+            tableNode.reloadData()
```
sosnovsky, line 407: Oops, I meant not just table reloading (it probably won't change anything, as it'll be the same data already used by table), but performing [refresh from remote mail server](https://github.com/FlowCrypt/flowcrypt-ios/blob/5fc8ba868fb755c02670b96463d64673780cbff3/FlowCrypt/Controllers/Inbox/InboxViewController.swift#L344)

ioanmo226: Yep! You are right. Just fixed it

sosnovsky (approved):
👍

PR #2380, opened 2023-09-29
#2367 Show `move to inbox` button for all cases when it's not inbox

### LINKED ISSUE #2367: mail in All Mail (archived) cannot be moved back to inbox
most can, but sometimes it can't, missing icon

example here

image

ioanmo226:
This PR shows `move to inbox` button for all cases when it's not inbox

close #2367 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Inbox/InboxViewController+TableView.swift
+++ b/FlowCrypt/Controllers/Inbox/InboxViewController+TableView.swift
@@ -417,7 +417,7 @@ extension InboxViewController {
         case .moveToTrash, .permanentlyDelete:
             removeMessage(at: indexToUpdate)
         case .archive, .moveToInbox:
-            if path.isEmpty { // no need to remove in 'All Mail' folder
+            if path.isEmpty || path == "SENT" { // no need to remove in 'All Mail' folder
                 updateMessage(
                     labelsToAdd: action == .moveToInbox ? [.inbox] : [],
                     labelsToRemove: action == .archive ? [.inbox] : [],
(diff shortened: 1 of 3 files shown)


sosnovsky (commented):
Looks good, but I noticed a small issue:
- compose and send new email
- go to Sent folder (or All Mail), open sent message
- toolbar shows `move to inbox` button, but should show `move to archive` button (as Gmail app does)

ioanmo226:
I believe Gmail's logic for handling sent emails is flawed. 
Archiving `Sent emails/threads` doesn't seem to have a purpose.(it actually does nothing)

Instead, offering a 'Move to Inbox' option for sent items would make more sense to me

ioanmo226:
Yep!. Nice catch. let me fix it

FlowCrypt/Controllers/Inbox/InboxViewController+TableView.swift
```diff
@@ -417,7 +417,7 @@ extension InboxViewController {
         case .moveToTrash, .permanentlyDelete:
             removeMessage(at: indexToUpdate)
         case .archive, .moveToInbox:
-            if path.isEmpty { // no need to remove in 'All Mail' folder
+            if path.isEmpty || path == "SENT" { // no need to remove in 'All Mail' folder
```
sosnovsky, line 420: it looks like we should remove messages only on `Inbox` screen, as moving to inbox or archiving doesn't remove messages from other folders.

For example, go to some label folder, archive message there - it should still stay in messages list of this label, but currently it's removed

ioanmo226: Yep! Correct

sosnovsky (approved):
Works well now 👍

PR #2413, opened 2023-10-26
#2396 Fix: choose correct keys for message encryption

### LINKED ISSUE #2396: Fix choosing correct key for message encryption
When WKD returns multiple keys for recipient, app should choose valid ~key~ **keys** for encrypting message.
Currently when there are multiple keys app seems to use just the first one.

Example:
Recipient has 2 public keys - one which can be used only for signing, and second one which can be used for encryption too. App fetches both these keys, but when trying to encrypt message it shows error `could not find valid encryption key`, so looks like it tries to use `only for signing` key to encrypt message, but it should choose key which can be used for encryption.

ioanmo226:
This PR fixed choosing correct key for message encryption

close #2396 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


FlowCrypt/Functionality/Services/Compose Message Helper/ComposeMessageHelper.swift
```diff
@@ -210,8 +210,15 @@ final class ComposeMessageHelper {
                 shouldUpdateLastUsed: true
             ).map(\.armored).joined(separator: "\n")
             let parsed = try await core.parseKeys(armoredOrBinary: armoredPubkeys.data())
+            // Fetch public keys which are usableForEncryption
+            // https://github.com/FlowCrypt/flowcrypt-ios/issues/2396
+            // Also fetch revoked or expired keys so that we can display correct error message
+            // UI Test title: sending message to user with expired/revoked public key produces modal
+            let keyDetails = parsed.keyDetails.filter {
+                $0.usableForEncryption || $0.revoked || !$0.isNotExpired
+            }
```
sosnovsky, line 219: probably we should also import sign-only keys to show appropriate error (in the case when there are no valid keys)?
as we already do for revoked and expired keys

appium/tests/specs/mock/composeEmail/CheckSignOnlyKey.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    const recipient = MockUserList.signOnlyKey;
+    const subject = 'sign only key subject';
+    const message = 'sign only key message';
+
+    mockApi.fesConfig = MockApiConfig.defaultEnterpriseFesConfiguration;
+    mockApi.ekmConfig = MockApiConfig.defaultEnterpriseEkmConfiguration;
+    mockApi.addGoogleAccount('e2e.enterprise.test@flowcrypt.com');
+    mockApi.attesterConfig = {
+      servedPubkeys: {
+        [MockUserList.signOnlyKey.email]: MockUserList.signOnlyKey.pub!,
```
sosnovsky, line 25: let's test original issue - user should have sign-only key and another one usable for encryption. and then we should check that app correctly chooses usable for encryption key

ioanmo226: variable name might be incorrect. but this key contains sign-only key plus normal key(usable for encryption)

FlowCrypt/Functionality/Services/Compose Message Helper/ComposeMessageHelper.swift
ioanmo226: I think we are seeking to only import correct key (usable for encryption) for message encryption in this issue?

sosnovsky: No, we should import all keys, but choose only valid ones when encrypting message, or show error message if there are no valid keys for recipient.

In current implementation if user manually imported sign-only key for recipient, then recipient email will be green when composing message, but when you tap `Send` button there will be error message `One or more of your recipients don't have encryption set up`. Which can be confusing for users, as key for recipient was imported and email was green when composing.

So it'll be better to validate keys before encrypting message and show appropriate error message.

appium/tests/specs/mock/composeEmail/CheckSignOnlyKey.spec.ts
sosnovsky: Got it, but having 2 different keys (not subkeys) should be more close to real situation.
We can use existing mock user (for example `robot@flowcrypt.com`) and add sign-only key as `pubSignOnly` property.

tomholub:
the title should say "correct keys" not "correct key". hopefully the implementation reflects that, didn't check

FlowCrypt/Functionality/Services/Compose Message Helper/ComposeMessageHelper.swift
ioanmo226: Could you please explain more about how we should proceed with this user which includes sign-only key?

Recipient status(maybe yellow or green?), what error message should I show when user choose to send message?, do we still have to allow user to send message in this case?

ioanmo226: Let's say recipient has 2 valid keys and 1 signing-only key.
Does above mean user won't send encrypted message in this case?

ioanmo226: Also does this mean we won't save draft in this case?

sosnovsky: > Let's say recipient has 2 valid keys and 1 signing-only key. Does above mean user won't send encrypted message in this case?

No, in this case we should use 2 valid keys and encrypt message with them. So message won't be send only in case when recipient doesn't have any valid key.

> Also does this mean we won't save draft in this case?

We should save draft in this case, but encrypt it only with current user private key (like it's saved when user composed draft but didn't set any recipient)

sosnovsky: > So we also need to show error message alert only when user doesn't have any valid key(like current sign-only key)

Correct. So if user has any valid key which can be used for encryption - we should use it (or if he has several valid keys, we should use all of them). If no valid key found - then appropriate error message should be shown.

sosnovsky:
I also think it'll be useful to have ability to send unencrypted message to recipient without valid keys, but it'll be for another issue.

For example, user tries to send message to recipient with expired key (or sign-only key) - currently we just show modal saying `recipient has expired key, please contact him`. But to message this recipient user will need to use another email client.

Here we can add button `Send unencrypted message` to error modal, tapping on it will open another compose screen with predefined text:
```
Hi,

I tried to compose encrypted message to you, but your public key is expired|revoked|sign-only.
Can you please share your valid public key.

Thanks
```

FlowCrypt/Functionality/Services/Compose Message Helper/ComposeMessageHelper.swift
ioanmo226: Hmm.. white text color doesn't look good on yellow background. Also yellow text color doesn't look good on white background.
So I'm thinking to use dark yellow like below(3rd and 4th image).

image
image
image
image

sosnovsky: Dark yellow is definitely better 👍

sosnovsky (commented):
Works great now 👍 Let's just rename some variables to be consistent

sosnovsky (approved):
👍

PR #2422, opened 2023-11-01
#2414 Show links to remote images

### LINKED ISSUE #2414: Show links to remote images
Was not sure where to post this - but on IOS Mobile Flowcrypt App there is no button for "Show Remote Images" - it just says in text "Remote content blocked for your privacy" and I cant find any settings to allow loading of remote images (which in case of some email I am getting is essential to approve security requests (they come as image with link and flowcrypt mobile does not load neither the image nor the link - while in gmail on desktop it loads images and links fine)

_Originally posted by @kubadoki in https://github.com/FlowCrypt/flowcrypt-browser/issues/4759#issuecomment-1780933675_

ioanmo226:
This PR implemented feature to show links to remote images

close #2414 // if this PR closes an issue
image

- Tests added or updated

### To be filled by reviewers


ioanmo226:
@sosnovsky Could you check current implementation if it's implemented correctly?
After I'll start to work on adding UI test.

sosnovsky:
> @sosnovsky Could you check current implementation if it's implemented correctly? After I'll start to work on adding UI test.

Implementation looks correct, but I noticed a small bug - some characters are rendered in wrong encoding:

it's message from github about your comment, in original message there is `—`
Screenshot 2023-11-01 at 23 17 16

in original message there is a space:
Screenshot 2023-11-01 at 23 20 22

And another issue is that previous thread replies are not hidden with 3-dots button, it causes `check thread rendering and archived message actions` ui test fail.

sosnovsky (commented):
I found another issue:
- send plain unencrypted message from Gmail web
- try to reply to it from FlowCrypt iOS app
- compose screen will show html code of original message, not plain text
Screenshot 2023-11-02 at 14 51 24

Also in unencrypted threads disappeared `>` symbols for previous replies (I'll send  comparison screenshots to your email)

ioanmo226:
Thank you. Nice find :-).
let me fix it

ioanmo226:
For this one do we have to display `text` content or html content which is translated into NSAttributedString?

> I found another issue:

sosnovsky:
> For this one do we have to display `text` content or html content which is translated into NSAttributedString?

let's show `text` content for reply compose, without remote images

ioanmo226:
OK

ioanmo226:
@sosnovsky, wrestling with HTML because we need to pluck out img tags from an HTML string and turn them into links. I’m hitting a snag with parsing quotes, though.

Here are a couple of solutions I'm considering:

- Whip up a regex to sniff out the gmail_quote class and split the quoted content from the main.
- Leverage a third-party library like SwiftSoup to navigate the HTML and isolate gmail_quote, separating the quoted and main content afterward.

What do you think? Got any slicker ideas?

https://github.com/FlowCrypt/flowcrypt-ios/blob/c8f746418c5bb2d166139912031317b413d0a2d6/FlowCrypt/Functionality/Mail%20Provider/Message%20Provider/ProcessedMessage.swift#L116

https://github.com/FlowCrypt/flowcrypt-ios/blob/c8f746418c5bb2d166139912031317b413d0a2d6/FlowCrypt/Functionality/Mail%20Provider/Message%20Provider/ProcessedMessage.swift#L128

image

sosnovsky:
I checked how it's implemented on Android - it just uses plain text from initial message, without image links:

Screenshot 2023-11-03 at 21 57 59

Probably it'll be easier to implement than trying to parse plain text from html?

ioanmo226:
I've checked and found that in the Android version, the app sanitizes the HTML and directly displays it in the EmailWebView, rather than using plain text, which contrasts with how things are handled in the iOS app.

In iOS app, we're trying to parse out the main content and quotes from the HTML to differentiate primary content from quoted messages, with the latter being accessible via a '3 dots' button. (As you know parsing main content and quote from plain text is already done in iOS. but as we want to display img links, we need to use html content.)

As of now, this parsing feature isn't present in the Android app, and it seems this could be linked to the absence of thread list functionality.

Do correct me if I've gotten anything wrong.

cc: @DenBond7

sosnovsky:
What if we'll additionally sanitize `processedMessage.fullText` and remove all html tags before sending it to compose screen here - https://github.com/FlowCrypt/flowcrypt-ios/blob/dae6c8edf128fe348c4d9b788c824ae906e74074/FlowCrypt/Controllers/Threads/ThreadDetailsViewController%2BTapActions.swift#L212

ioanmo226:
Sorry @sosnovsky .
I'm talking about thread detail screen(Not compose screen)
In thread detail page, we have to split main content and quoted content like below image. But current problem is we don't have way to parse quote & main content from html content.

image

sosnovsky:
> Sorry @sosnovsky . I'm talking about thread detail screen(Not compose screen) In thread detail page, we have to split main content and quoted content like below image. But current problem is we don't have way to parse quote & main content from html content.

What if we'll just use regex to extract quote content from `.gmail_quote` div?

ioanmo226:
Yep. It is same as my 1st suggested solution.
But one thing I worry is that  processing html with regex might produce another issues (like not displaying contents correctly).

> Whip up a regex to sniff out the gmail_quote class and split the quoted content from the main.
> Leverage a third-party library like SwiftSoup to navigate the HTML and isolate gmail_quote, separating the quoted and main content afterward.

sosnovsky:
> Yep. It is same as my 1st suggested solution.

Oops, missed it 👍

> But one thing I worry is that processing html with regex might produce another issues (like not displaying contents correctly).

I think it should be ok, as we won't make any complex transformations - just parse `div.gmail_quote` content, should work well. Also it'll be applied only to non-encrypted messages, as we don't have html in encrypted ones.

ioanmo226:
@sosnovsky Fixed above issues. Please check again.
Code looks a bit messy though :-(

sosnovsky (commented):
It works better now, but still there are issues with rendering quote content in some threads, as parsing html and removing tags sometimes leads to broken rendering (your assumption about issues with incorrect contents display was correct)

Some possible solutions:
- use plain text message body for quote content and html message body for message content
- add separate action `show html version` in 3-dots message menu
- try to render html content like it's currently implemented on Android - I haven't noticed issues with rendering quotes content there

What do you think about mentioned solutions?

ioanmo226:
What issues do you have while rendering quote content?

Also here are my thoughts for your solutions

1. So you mean you only have issues while rendering quote content? you do not notice issues while rendering main content?
2. Even though we implement `show html version`, we will have same issues when user clicks `show html version` button?
3. This way, we will lose `3 dots toggle` button which would toggle quote content view.

sosnovsky:
> So you mean you only have issues while rendering quote content? you do not notice issues while rendering main content?

Initially I noticed only quoted content issues, for example when using `text` body different quote levels are differentiated by different number of `>` in the beginning of the line, while in `html` body they indented with tabs which are removed when html is sanitized, so all previous quotes are on the same level.

But then I found that it affects also main content of messages too, as `html` body usually includes additional markup and just removing tags leads to broken message structure. Like this message from Google:

current PR


App Store version


Android


> Even though we implement show html version, we will have same issues when user clicks show html version button?

Original issue reported by user was that he wasn't able to approve security requests because image with link wasn't loaded in FlowCrypt app. Separate `show html version` button will allow to load full html version, which will include all images and links. And broken rendering of quoted content shouldn't be too critical there.

> This way, we will lose 3 dots toggle button which would toggle quote content view.

Yeah, that's why it's difficult to choose which solution will be the best one, all of them have pros and cons :)
I prefer this one, it should make rendered content look better, as even current implementation with `text` body has some flaws like unnecessary newlines (they're presented in the screenshot from App Store version).

So if we'll be able to separate quoted content from html body - it should give us the best result.

ioanmo226:
> I prefer this one, it should make rendered content look better

ioanmo226:
Sorry but which one do you mean?
> I prefer this one, it should make rendered content look better

sosnovsky:
> Sorry but which one do you mean?
> 
> > I prefer this one, it should make rendered content look better

This one - `try to render html content like it's currently implemented on Android`

ioanmo226:
You mean just display html content in WebView after sanitizing html? (Without `show quoted content` button`).
Also only for regular plain emails?

sosnovsky:
> You mean just display html content in WebView after sanitizing html? (Without `show quoted content` button`). Also only for regular plain emails?

Removing `show quoted content` button will make some threads too long for scrolling (as it was reason for adding such button). So it'll be great to find solution for splitting quoted content even in `html` body and then probably have 2 webviews - for message and for quoted content (which will be shown after `show quoted content` button tap)

ioanmo226:
How about my previous suggestion then?

> Leverage a third-party library like SwiftSoup to navigate the HTML and isolate gmail_quote, separating the quoted and main content afterward.

sosnovsky:
> How about my previous suggestion then?
> 
> > Leverage a third-party library like SwiftSoup to navigate the HTML and isolate gmail_quote, separating the quoted and main content afterward.

Yeah, it can be helpful in splitting email html, but to start with let's try to render `html` body in web view instead of current plain text. To check how it looks and how quoted content is rendered. For now, no need to split message and quoted content, let's render it as on Android.

ioanmo226:
@sosnovsky Please check this implementation

sosnovsky:
> @sosnovsky Please check this implementation

Looks promising - content looks better formatted comparing to html converted to plain text.
I think it's a correct direction, just need to some improvements:
- we strip too much tags, so `style="..."` is removed too and it breaks some styles
- sometimes web view height calculated incorrectly and there are a lot of empty space in the bottom
- most messages have a white background, which looks too bright when app is in dark mode, maybe there is some way to adapt email html to dark mode
- messages font is too small and difficult to read, comparing to other UI texts

FlowCryptUI/Cell Nodes/ThreadDetailWebNode.swift
```diff
@@ -0,0 +1,46 @@
...
+    private let input: ThreadDetailWebNode.Input
+
+    private let webViewNode = CustomWebViewNode()
+
+    public init(input: ThreadDetailWebNode.Input) {
+        self.input = input
+
+        super.init()
+        addLeftBorder(width: .threadLeftBorderWidth, color: UIColor(hex: "777777"))
+        webViewNode.setHtml("<style>* {font-size: 20pt;} </style>" + (input.message ?? ""))
```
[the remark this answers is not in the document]
ioanmo226: Yep. sounds good!

ioanmo226:
Do you want to leave `style` tag when stripping?
> we strip too much tags, so style="..." is removed too and it breaks some styles

sosnovsky:
> Do you want to leave `style` tag when stripping?

Yes, let's try to leave it, as it shouldn't affect app security, I think

ioanmo226:
I'm not expert at this security stuff but I think we might consider about CSS Injection via style tag?

https://www.getastra.com/blog/knowledge-base/guide-on-css-injection-prevention/

sosnovsky:
> I'm not expert at this security stuff but I think we might consider about CSS Injection via style tag?
> 
> https://www.getastra.com/blog/knowledge-base/guide-on-css-injection-prevention/

Seems like it applies only to opening such HTML in browser, where injection can fetch some data. While WKWebView is isolated web view, which just renders provided HTML so it shouldn't leak any user data.

It's also possible to completely disable JS in WKWebView - https://developer.apple.com/documentation/webkit/wkwebpagepreferences/3552422-allowscontentjavascript, so no malicious code will be executed.

ioanmo226:
Sounds good. Thank you

appium/tests/specs/mock/inbox/CheckRemoteImageRendering.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+      await SetupKeyScreen.setPassPhrase();
+      await MailFolderScreen.checkInboxScreen();
+
+      await MailFolderScreen.clickOnEmailBySubject(subject);
+      await EmailScreen.checkOpenedEmail(sender, subject, message);
+
+      // TODO: Check if WKWebView content contains remote content blocked image content. Couldn't seem to find a way to check WKWebView content
+      // https://discuss.appium.io/t/appium-and-wkwebview/4769/10
+      // await EmailScreen.checkEmailText(remoteContentBlockedMessage);
+      // await EmailScreen.checkEmailText('[img]');
```
ioanmo226, line 32: I left this as a `TODO` as I couldn't find way to check WKWebView content in appium.
https://discuss.appium.io/t/appium-and-wkwebview/4769/10.

@sosnovsky Do you know the way to check appium content?

sosnovsky:
> @sosnovsky As you can see, Core unit test fails because of `background` property not removed from style attribute. Like below leak html (which background css is in background). Are we fine to remove those checks?

No, let's leave these checks - it seems `html-sanitize` allows filtering of css styles (https://github.com/apostrophecms/sanitize-html?tab=readme-ov-file#allowed-css-styles), let's try to remove `background` properties which have `url` in value, so no remote content will be loaded

appium/tests/specs/mock/inbox/CheckRemoteImageRendering.spec.ts
sosnovsky: Probably we should enable web views debugging as described in Appium's docs - https://github.com/appium/appium-xcuitest-driver/blob/master/docs/device-preparation.md#enable-web-testing-availability

ioanmo226: Hmmm.. enabled web views debugging by setting `isInspectable` to true but still email body text becomes null.
Digging deeper now

sosnovsky: Probably we can just replace this test with Swift unit test, which will check that `Core.parseDecryptMsg` adds `remote content is blocked` text and replaces remote images with links

ioanmo226: Might be better to add webview content check logic for the future testing I think.

ioanmo226:
@sosnovsky Ready for review. Please check. Thank you

sosnovsky (commented):
Great job 👏 I like how messages look with these updates, but some improvements still needed:
- change font to iOS system font, it can be done by adding `<style>` tag after `<meta name="viewport" ...` line:
```html
<meta name="viewport" ... >
<style>* { font-family: -apple-system, "Helvetica Neue", sans-serif; }</style>
```
- support for dark mode (as currently messages always have white background) can be added in `<style>` tag too:
```css
:root { color-scheme: light dark; supported-color-schemes: light dark; }
```
And then we can set background in dark mode to be the same as app background using
```css
@media (prefers-color-scheme: dark) {
  :root {
    background-color: OUR_IOS_DARK_BACKGROUND;
    color: white;
  }
}
```
- web view has additional side paddings (around 8px, I think), let's remove it so web view text will be the same width as text in encrypted message. Maybe it's can be done through CSS too (like `html, body { padding: 0 !important; margin: 0 !important; }`) or by increasing width of web view.

ioanmo226:
Sounds good 👍 
> change font to iOS system font, it can be done by adding <style> tag after <meta name="viewport" ... line:

Oops. Missed it. Thank you for your notice
> support for dark mode (as currently messages always have white background) can be added in <style> tag too:

OKay
> web view has additional side paddings (around 8px, I think)

ioanmo226:
@sosnovsky It looks great with your suggestions. Thanks.
Please check.

FlowCryptUI/Cell Nodes/ThreadDetailWebNode.swift
```diff
@@ -0,0 +1,65 @@
...
+                <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0, user-scalable=no">
+                <style>
+                    * { font-family: -apple-system, "Helvetica Neue", sans-serif; }
+                    :root { color-scheme: light dark; supported-color-schemes: light dark; }
+                    @media (prefers-color-scheme: dark) {
+                      :root {
+                        background-color: #2D2C2E;
+                        color: white;
+                      }
+                    }
```
sosnovsky, line 38: looks great now, just need to update links color in dark mode as currently they're quite difficult to read:
Screenshot 2023-11-23 at 11 49 42

let's use the same color as for links in encrypted messages:
Screenshot 2023-11-23 at 15 13 40

FlowCryptUI/Cell Nodes/ThreadDetailWebNode.swift
```diff
@@ -0,0 +1,65 @@
...
+        """)
+        node.style.flexGrow = 1.0
+        return node
+    }()
+
+    public init(input: ThreadDetailWebNode.Input) {
+        self.input = input
+
+        super.init()
+        addLeftBorder(width: .threadLeftBorderWidth, color: UIColor(hex: "777777"))
```
sosnovsky, line 52: let's make a named color from `UIColor(hex: "777777")`, for example `plainTextBorder`, and use it here and in other places where `777777` color is used

appium/tests/screenobjects/email.screen.ts
```diff
@@ -1,6 +1,7 @@
 import BaseScreen from './base.screen';
 import { CommonData } from '../data';
 import ElementHelper from '../helpers/ElementHelper';
+import WebView from 'tests/helpers/WebView';
```
sosnovsky, line 4: can be replaced with `../helpers/WebView` for consistency

sosnovsky (commented):
Well done, just a couple small fixes left and then let's work on hiding/showing quote text in another PR

FlowCryptUI/Cell Nodes/ThreadDetailWebNode.swift
ioanmo226: Sounds good!

FlowCryptUI/Cell Nodes/ThreadDetailWebNode.swift
ioanmo226: Okay

sosnovsky (commented):
Oops, just noticed that issue from earlier reviews is still present (https://github.com/FlowCrypt/flowcrypt-ios/pull/2422#pullrequestreview-1710176791):
- send non-encrypted message from Gmail web interface
- open it in iOS app
- tap `reply` button
- on compose screen there will be html code
Screenshot 2023-11-23 at 16 34 41

Let's use plain text body on compose screen, as we don't support creating of html messages yet.

sosnovsky (approved):
Looks good now, well done 👍

PR #2429, opened 2023-11-08
#2421 feat: added ability to message recipient without valid public key

### LINKED ISSUE #2421: Add ability to message recipient without valid public key
I also think it'll be useful to have ability to send unencrypted message to recipient without valid keys, but it'll be for another issue.

For example, user tries to send message to recipient with expired key (or sign-only key) - currently we just show modal saying `recipient has expired key, please contact him`. But to message this recipient user will need to use another email client.

Here we can add button `Send unencrypted message` to error modal, tapping on it will open another compose screen with predefined text:
```
Hi,

I tried to compose encrypted message to you, but your public key is expired|revoked|sign-only.
Can you please share your valid public key.

Thanks
```

_Originally posted by @sosnovsky in https://github.com/FlowCrypt/flowcrypt-ios/issues/2413#issuecomment-1785075870_

ioanmo226:
This PR added ability to message recipient without valid public key

close #2421 // if this PR closes an issue

https://github.com/FlowCrypt/flowcrypt-ios/assets/77344191/ad6eacc7-b451-485b-9cb8-ef6a549d4d78

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+ErrorHandling.swift
+++ b/FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+ErrorHandling.swift
@@ -107,7 +107,9 @@ extension ComposeViewController {
             style: .cancel
         )
         cancelAction.accessibilityIdentifier = "aid-cancel-button"
-        alert.addAction(sendUnEncryptedAction)
+        if !Bundle.isEnterprise {
+            alert.addAction(sendUnEncryptedAction) // Disallow sending plain message for enterprise
+        }
         if isMessagePasswordSupported {
             alert.addAction(sendPasswordProtectedAction)
         }
--- a/FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+ErrorHandling.swift
+++ b/FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+ErrorHandling.swift
@@ -101,7 +101,7 @@ extension ComposeViewController {
                 self?.setMessagePassword()
             }
         )
-        sendUnEncryptedAction.accessibilityIdentifier = "aid-compose-send-message-password"
+        sendPasswordProtectedAction.accessibilityIdentifier = "aid-compose-send-message-password"
         let cancelAction = UIAlertAction(
             title: "cancel".localized,
             style: .cancel
--- a/FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+ErrorHandling.swift
+++ b/FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+ErrorHandling.swift
@@ -66,17 +66,11 @@ extension ComposeViewController {
         }
 
         switch error {
-        case MessageValidationError.noPubRecipients:
-            guard isMessagePasswordSupported else {
-                showPlainMessageConfirmationAlert()
-                return
-            }
-
-            if Bundle.isEnterprise {
-                setMessagePassword()
-            } else {
-                showPlainMessageAlert()
-            }
+        case MessageValidationError.noPubRecipients,
+             MessageValidationError.revokedKeyRecipients,
+             MessageValidationError.expiredKeyRecipients,
+             MessageValidationError.notUsableForEncryptionKeyRecipients:
+            processSendMessageWithNoValidKeys(error: error)
         case MessageValidationError.notUniquePassword,
              MessageValidationError.subjectContainsPassword,
              MessageValidationError.weakPassword:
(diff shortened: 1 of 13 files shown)


ioanmo226:
@sosnovsky I implemented like above one.
I think this one might be sufficient? (Allow user to send encrypted/password-protected message/cancel).
Prefixing text(like `Hi, i tried to send you encrypted message...`) is also a good idea (need to add extra logic though) but I think above implementation might be sufficient.
Let me know your thought.

sosnovsky:
> @sosnovsky I implemented like above one. I think this one might be sufficient? (Allow user to send encrypted/password-protected message/cancel). Prefixing text(like `Hi, i tried to send you encrypted message...`) is also a good idea (need to add extra logic though) but I think above implementation might be sufficient. Let me know your thought.

Yes, such implementation is simpler and good enough, allowing users to send messages for recipients even with invalid key.

I just noticed that `Web portal password added` section isn't visible when sending password message to recipient with invalid key:
Screenshot 2023-11-08 at 22 36 09

FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+TableView.swift
```diff
@@ -29,7 +29,7 @@ extension ComposeViewController: ASTableDelegate, ASTableDataSource {
         case (.main, .recipients(.cc)), (.main, .recipients(.bcc)):
             return !shouldShowEmailRecipientsLabel && shouldShowAllRecipientTypes ? 1 : 0
         case (.main, .password):
-            return isMessagePasswordSupported && contextToSend.hasRecipientsWithoutPubKey ? 1 : 0
+            return isMessagePasswordSupported && contextToSend.hasRecipientWithInvalidKey ? 1 : 0
```
ioanmo226, line 32: @sosnovsky After I changed this logic, UI test works strangely in local side(it works fine in semaphoreci side).
`setComposeSecurityMessage` function is not working properly and just append security message next to subject label. As a result, all UI tests which use `composeEmail` test fails.

Could you check in your local side and if this is also your case?

https://github.com/FlowCrypt/flowcrypt-ios/blob/c89812935ddf8fa856ef2456acde7a39b0081fe1/appium/tests/screenobjects/new-message.screen.ts#L199

sosnovsky: > setComposeSecurityMessage function is not working properly and just append security message next to subject label.

Same happens for me - I noticed that when all recipients are valid, then message is entered correctly. But if any recipient has invalid key and `set web portal password` cell is visible - message text is entered in subject text field.
Maybe we need some additional tap at message text view before entering message text, so cursor will be focused in the correct textfield.

ioanmo226: Yeah, sounds good.
Thank you for checking it

ioanmo226:
@sosnovsky Ready for review. Please check

FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+ErrorHandling.swift
```diff
@@ -86,26 +80,39 @@ extension ComposeViewController {
...
+                self?.setMessagePassword()
+            }
         )
+        sendPasswordProtectedAction.accessibilityIdentifier = "aid-compose-send-message-password"
+        let cancelAction = UIAlertAction(
+            title: "cancel".localized,
+            style: .cancel
+        )
+        cancelAction.accessibilityIdentifier = "aid-cancel-button"
+        alert.addAction(sendUnEncryptedAction)
```
sosnovsky, line 110: Previously we didn't allow to send unencrypted message for enterprise builds, let's add this restriction back - so users of enterprise build will be able to send only password-protected messages, but not unencrypted

appium/tests/specs/mock/composeEmail/SendEmailToRecipientWithoutPublicKey.spec.ts
```diff
@@ -99,20 +91,6 @@ describe('COMPOSE EMAIL: ', () => {
...
-      await MailFolderScreen.checkInboxScreen();
-      await MailFolderScreen.clickCreateEmail();
-      await NewMessageScreen.composeEmail(recipient, emailSubject, emailText);
-      await NewMessageScreen.checkFilledComposeEmailInfo({
-        recipients: [recipient],
-        subject: emailSubject,
-        message: emailText,
-      });
-      await NewMessageScreen.clickSendButton();
-      await BaseScreen.checkModalMessage(passwordModalMessage);
```
sosnovsky, line 115: this part of the test should be added back

sosnovsky (commented):
Works well, just need to add back restriction for enterprise builds

FlowCrypt/Controllers/Compose/Extensions/ComposeViewController+ErrorHandling.swift
ioanmo226: Aha, I see

sosnovsky (approved):
Works great now, thanks!

PR #2458, opened 2023-11-30
#2438 improve handling of invalid keys on message compose

### LINKED ISSUE #2438: Improve handling of invalid keys on message compose
When user tries to send encrypted message, but stored local keys are invalid (expired or revoked) then we just show `Your account keys are not usable for encryption` error message:

In this case, before showing error message, we should try to fetch updated keys from EKM and show error only if there are no valid keys on EKM.

ioanmo226:
This PR improved handling of invalid keys on message compose(Refresh EKM keys for invalid keys on message screen)

close #2438 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -78,7 +78,7 @@ blocks:
         commands:
           - checkout && cd ~/git/flowcrypt-ios/
           - mv ~/appium-env ~/git/flowcrypt-ios/appium/.env
-          - sem-version node 18 && cache restore appium-npm && cd ./appium && npm i && cd .. && cache store appium-npm appium/node_modules
+          - sem-version node 20 && cache restore appium-npm && cd ./appium && npm i && cd .. && cache store appium-npm appium/node_modules
           - cd appium
           - cache restore FlowCrypt-$SEMAPHORE_GIT_SHA.app
       epilogue:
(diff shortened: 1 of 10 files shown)


sosnovsky:
> @sosnovsky Could you check in your local if below semaphoreci issue also occurs in your local? It works fine in my local though. Only semaphoreci.

Works well locally for me too.
Semaphore issue can be caused by different `node.js` version there, maybe they changed it recently, so setting node version explicitly in `semaphore.yml` should fix this issue.

ioanmo226:
I think we set to correct node v18 in semaphoreci?

https://github.com/FlowCrypt/flowcrypt-ios/blob/6fd5696fd9841950006b69e7755924091eca715f/.semaphore/semaphore.yml#L81

sosnovsky:
> I think we set to correct node v18 in semaphoreci?

I checked your previous PR (where tests were passing) and Semaphore used node 18.18.2 there, while this PR uses node 18.19.0, so looks like there is some changes in this node update which broke tests code.

You can try to use node v20 on semaphore, to check if it fixes issue.

ioanmo226:
Yep!
Seems like `v18.19.0` is not stable and when I change it to `v20`, it works fine.

ioanmo226:
Thank you for your assistant

sosnovsky (approved):
Works great now, thanks!

PR #2626, opened 2024-09-27
#2548 Add warning for suspicious message

### LINKED ISSUE #2548: Add warning for suspicious messages
Suspicious messages has `Received-SPF: softfail` header, for such messages we need to show warning before message body:

```
Potentially suspicious message

It wasn't properly verified by the sender, so its authenticity can't be confirmed.
Be careful - avoid clicking links and downloading attachments, or sharing personal info.
```

ref: https://github.com/FlowCrypt/flowcrypt-security/issues/270
android: https://github.com/FlowCrypt/flowcrypt-android/issues/2670

ioanmo226:
This PR added warning for suspicious message

close #2548 // if this PR closes an issue

- Tests added or updated

image

### To be filled by reviewers


ioanmo226:
@sosnovsky XCode 16 image is not available in `a1-standard-4` machine so had to change it to `a2-standard-4`
But it has 2 concurrency running limit.
Do we have to increase this limit for `a2-standard-4` machine?

image

sosnovsky:
Probably this limit is set by Semaphore CI, as I don't see settings section for changing these machine limits.
But 2 machines running limit should be ok, as all jobs will still be executed, one-by-one.

Also I noticed that build time decreased from 30 minutes on previous machine type to 8 minutes on `a2-standard-4`, then we can just merge 4 test jobs into 2, if they'll run faster too.

And let's also add `fail_fast` configuration to `semaphore.yml` like at https://github.com/FlowCrypt/flowcrypt-ios/blob/79651582a739faa3e21cc233ae99c15126385bf3/.semaphore/semaphore.yml#L12-14 (it'll stop other jobs executions after first test failure):
```yml
fail_fast:
  stop:
    when: branch != 'master'
```

ioanmo226:
Will fix XCode 16 image issue in new issue at https://github.com/FlowCrypt/flowcrypt-ios/issues/2627

appium/tests/specs/mock/inbox/CheckSecurityWarningForSuspiciousEmail.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+    });
+
+    await mockApi.withMockedApis(async () => {
+      await SplashScreen.mockLogin();
+      await SetupKeyScreen.setPassPhrase();
+      await MailFolderScreen.checkInboxScreen();
+
+      await MailFolderScreen.clickOnEmailBySubject(subject);
+
+      await EmailScreen.checkSecurityWarningBlock();
```
sosnovsky, line 23: let's move this test to `CheckMessageProcessingErrors` test, which includes other message processing errors. having all such cases in 1 test should make test execution faster, than having a separate test.

sosnovsky (commented):
Works well 👍

FlowCryptUI/Cell Nodes/SecurityWarningNode.swift
```diff
@@ -0,0 +1,43 @@
...
+    private lazy var subjectNode: ASTextNode2 = {
+        let textNode = ASTextNode2()
+        textNode.attributedText = "message_security_warning_subject".localized.attributed(.bold(18), color: .black, alignment: .left)
+        textNode.accessibilityIdentifier = "aid-security-warning-subject-node"
+        return textNode
+    }()
+
+    private lazy var messageNode: ASTextNode2 = {
+        let textNode = ASTextNode2()
+        textNode.attributedText = "message_security_warning_message".localized.attributed(.thin(16), color: .black, alignment: .left)
```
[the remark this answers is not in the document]
ioanmo226: Sounds good!

appium/tests/specs/mock/inbox/CheckSecurityWarningForSuspiciousEmail.spec.ts
ioanmo226: Okay

sosnovsky (approved):
Well done 👍

PR #2628, opened 2024-10-01
#2627 Use Xcode16 in sempahoreci

### LINKED ISSUE #2627: Use XCode16 image in SemaphoreCI

ioanmo226:
This PR uses Xcode16 in sempahoreci

close #2627  // if this PR closes an issue

### To be filled by reviewers


ioanmo226:
Build still fails after changing fastfile & semaphoreci configuration.
I think it still fails because `appium-xcuitest-driver` doesn't support stable Xcode16 version?
@sosnovsky  What's your thought? 

https://github.com/appium/appium-xcuitest-driver/blob/master/docs/installation/requirements.md#:~:text=Xcode%2016%2Dbeta,7.24.15

sosnovsky:
It builds well, but error happens when appium tries to connect to `webdriverio`. I tried to upgrade `webdriverio` to v9 in https://github.com/FlowCrypt/flowcrypt-ios/tree/xcode-16-test, but it still fails, maybe needs some configuration changes, checking it too

ioanmo226:
I still think build fails because `appium-xcuitest-driver` doesn't support stable Xcode16 version?

sosnovsky:
I tried to run tests locally with Xcode 16 stable and they work well, can be just some CI-specific issue, which I hope can be fixed, as it would be great to have CI builds completed in 7 minutes instead of 30 :)

sosnovsky:
Hi @ioanmo226, I found reason for tests failure on Semaphore, but still can't find fix for it - Appium opens iOS simulator using `open -Fn /Applications/Xcode-16.0.0.app/Contents/Developer/Applications/Simulator.app` command (https://github.com/appium/appium-ios-simulator/blob/master/lib/simulator-xcode-10.js#L420), but it fails on Semaphore M1 machines with `Domain does not support specified action` error message.

The same happens on Semaphore M1 machines with Xcode 15, so it's not Xcode 16 specific issue.

Locally this command works well on M1, can be some Semaphore machine security settings restricting running apps from root `/Applications` folder. Maybe you have some tips on possible fix?

ioanmo226:
Aha, I see
Let me dig more

ioanmo226:
I think this might be the correct for our case?
SemaphoreCI mac machine is just running as a terminal and can't open interactive applications like Simulator/XCode?

https://forums.developer.apple.com/forums/thread/675543#:~:text=If%20you%27re%20not%20sure%20you%20can%20run%20the%20launchctl%20managername%20command%20to%20determine%20it.%20If%20it%20prints%20something%20else%20than%20Aqua%20it%27s%20most%20probably%20a%20non%2Dinteractive%20session%20and%20running%20graphical%20apps%20won%27t%20work.

sosnovsky:
Not sure about it, as Semaphore machines should be able to run Xcode and simulators for UI tests.
Also previous `a1-...` machines were able to open these apps, so I still believe that it's some misconfiguration, but let's wait for Semaphore support response - they should be better acknowledged about such issues.

sosnovsky:
Hi @ioanmo226, I think with current Semaphore configuration where `a2-standard-4` machines are not able to run UI tests, we can use `a2-standard-4` machines for building FlowCrypt app (which takes [4 minutes](https://flowcrypt.semaphoreci.com/workflows/360fa577-122d-4835-a2c9-b2c16ed38130?pipeline_id=ee2c549b-c000-4056-939d-dcb8459d9c6c) instead of [35 minutes](https://flowcrypt.semaphoreci.com/jobs/c5a3241c-a798-4558-96e7-051690676b55))  and WebDriverAgent, and then use `a1-standard-4` for running UI tests. I tried to use such configuration in [xcode-16-test](https://github.com/FlowCrypt/flowcrypt-ios/blob/xcode-16-test/.semaphore/semaphore.yml) branch

ioanmo226:
That sounds good.
As you can see from SemaphoreCI contact ticket, XCode16 image still doesn't work with `a1-standard-4` machine and they are working on to fix it. (which returns `npm not recognized error`)

sosnovsky:
Hi @ioanmo226, `application does not support the x86_64 Simulator architecture` error probably appears because app is built on `arm64` machine, while `a1-standard-4` has `x86_64` architecture.

It should be possible to build app for both architectures, I've tried to do it by adding [ARCHS='x86_64 arm64' VALID_ARCHS='x86_64 arm64'](https://github.com/FlowCrypt/flowcrypt-ios/blob/xcode-16-test/fastlane/Fastfile#L22) build args, but they didn't work, will need to check it further.

PR #2639, opened 2024-10-23
Fix unstable SemaphoreCI tests

### LINKED ISSUE #2638: Unstable SemaphoreCI tests
Tests are consistently failing in SemaphoreCI, making it difficult to achieve a passing result.
As a result, there are many PRs where tests pass locally but fail in SemaphoreCI.

We need to stabilize and fix these unstable tests in SemaphoreCI.

ioanmo226:
This PR fixed unstable SemaphoreCI tests.

close #2638 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


appium/tests/specs/mock/setup/CheckRefreshKeyFromEkmWithPassPhrasePrompt.spec.ts
```diff
@@ -33,13 +33,16 @@ describe('SETUP: ', () => {
...
-      await RefreshKeyScreen.fillPassPhrase('wrong passphrase');
-      await RefreshKeyScreen.clickOkButton();
-      await BaseScreen.checkModalMessage(CommonData.refreshingKeysFromEkm.wrongPassPhrase);
-      await RefreshKeyScreen.clickSystemOkButton();
-      await RefreshKeyScreen.cancelRefresh();
-      await KeysScreen.openScreenFromSideMenu();
-      await KeysScreen.checkKeysScreen([ekmKeySamples.key0]);
+      // TODO: Temporarily disable the logic below as the cancel button on the refresh screen
+      // is not functioning correctly in Semaphore CI for some reason.
+      // https://flowcrypt.semaphoreci.com/workflows/7310b5b7-d544-428c-82ce-7f7f060e7e60?pipeline_id=186f2ad1-5d22-44d3-ba07-76deaf3ea49f
```
ioanmo226, line 38: @sosnovsky Have any idea why this test fails (can't click button even though elements are visible on the screen)

ioanmo226:
@sosnovsky I’ve fixed most of the unstable tests, but in a few cases, they still occasionally fail on SemaphoreCI.

The main issue seems to be a 2-hour timeout limit we set in SemaphoreCI config, where tests get stuck on the home screen (with a list of apps in a closed state) or on the loading progress HUD.
Looks like we need to live with this until SemaphoreCI upgrades to M2 or M3 Macs.

Let me know your thoughts

appium/tests/specs/mock/setup/CheckRefreshKeyFromEkmWithPassPhrasePrompt.spec.ts
sosnovsky: Can be caused by some changes in iOS behavior which were adapted in the newer Appium version.
After upgrading to the latest Appium we can re-check if it works there.

sosnovsky (approved):
Looks good, thanks for checking out, hope it'll help to make tests passing in the new PRs 👍

sosnovsky:
> The main issue seems to be a 2-hour timeout limit we set in SemaphoreCI config, where tests get stuck on the home screen (with a list of apps in a closed state) or on the loading progress HUD.
> Looks like we need to live with this until SemaphoreCI upgrades to M2 or M3 Macs.

It should be ok for now, as previously it was almost impossible to make all tests pass.
Let's hope Semaphore will start using M3 machines soon, which should greatly improve tests performance.

PR #2653, opened 2025-01-17
#2651 Update to openpgp v6

### LINKED ISSUE #2651: Upgrade openpgp to v6
similar to https://github.com/FlowCrypt/flowcrypt-browser/pull/5866

ioanmo226:
This PR upgraded openpgp to v6

close #2651 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


ioanmo226:
Hi @sosnovsky,

I’ve completed migrating OpenPGP to v6. However, v6 heavily relies on the Web Crypto API, which doesn’t seem to be supported by the iOS WebView. This results in the following error caused by the line below:

https://github.com/openpgpjs/openpgpjs/blob/b2bd8a0fdd12902484d65baa4ae4eb7f146fcd32/src/util.js#L432

> "⚙️[Core] The WebCrypto API is not available\n -> js getWebCrypto@\n -> js 6382@\n -> js r@\n -> js 1341@\n -> js r@\n -> js 9545@\n -> js r@\n -> js 4010@\n -> js r@\n -> js 1592@\n -> js r@\n -> js 9033@\n -> js r@\n -> js @\n -> js @\n -> js global code@"

Some suggest using a polyfill, others recommend implementing an iOS native bridge to handle crypto operations, and some propose reverting back to OpenPGP v5—which we’d prefer to avoid.

Let me know your thoughts and suggestions.

sosnovsky:
Browser compatibility table shows that most of `WebCrypto API` functionality is supported in iOS WebView since iOS 11 - https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto#browser_compatibility
Maybe we need to change some configuration properties for `WKWebView` to enable `WebCrypto API`.
But as I understand, it should work without adding any polyfills. Can you please check it?

ioanmo226:
Sure. Let me check

ioanmo226:
From my research, it looks like we have two options(The Web Crypto API is only available in a secure context, so right now it’s undefined): 

1. **Run a local HTTPS server** for the JS file.  
2. **Register a custom webview scheme handler** so iOS treats the custom scheme as HTTPS.

I tried registering a custom scheme handler, but iOS still doesn’t recognize it as secure. It might be best to set up a local HTTPS server, though we’ll need to consider certificates (maybe a self-signed one). I’ll dig more into this.

Let me know your thoughts!

ioanmo226:
I tried both Telegraph and Flyingfox (which doesn’t support TLS) to serve **flowcrypt-ios-prod.js.txt** over HTTP/HTTPS, but I still got the same error.  
Digging deeper now.

ioanmo226:
I created a test application and loaded google.com into a WebView. Then, I inspected the WebView to check if the crypto.subtle API was available.

However, even with this simple WebView displaying google.com, the crypto.subtle API is not accessible.

Could it be that the WebCrypto API is not available in the WebView context? Or am I missing something?

Could you jump in and take a look at this issue with me?

sosnovsky:
ok, I think I'll be able to check it a bit later.
for now let's then work on browser extension tasks.
we got a minor ux issue report from the customer, should be a convenient improvement - https://github.com/FlowCrypt/flowcrypt-browser/issues/5926

ioanmo226:
Hmm.. strange let me check again

sosnovsky:
I also was able to make `openpgp` work by loading our core js file in `webView(_ webView: WKWebView, didFinish navigation: WKNavigation!)` method, after loading `https://flowcrypt.com` in `WKWebView`:
```swift
func setupWebView() {
        let userController = WKUserContentController()
        userController.add(self.coreMessageHandler, name: "coreHost")
        let configuration = WKWebViewConfiguration()
        configuration.userContentController = userController
        configuration.defaultWebpagePreferences.allowsContentJavaScript = true
        self.webView = WKWebView(frame: .zero, configuration: configuration)
        self.webView.navigationDelegate = self.coreMessageHandler
        if #available(iOS 16.4, *) {
            self.webView.isInspectable = true
        }

        let url = URL(string: "https://flowcrypt.com")!
        webView.load(URLRequest(url: url))
    }

// and then in CoreMessageHandler
func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
        guard let jsFileSrc = self.getCoreJsFile() else { return }
        let appVersion = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String ?? "[unknown version]"
        webView.evaluateJavaScript("const APP_VERSION = 'iOS \(appVersion)';\(jsFileSrc)") { _, _ in }
    }
```

But message decryption still fails with such error:

Screenshot 2025-02-10 at 15 07 27

I'll try to make it works with local https server, maybe using `webView(_ webView: WKWebView, didFinish navigation: WKNavigation!)` will make `openpgp` work correctly there too.

ioanmo226:
I tried with simple local http server and seems like openpgp code works fine in there too.

sosnovsky:
> I tried with simple local http server and seems like openpgp code works fine in there too.

Oh, I thought it'll work only with https server. Using http server should be much simpler

sosnovsky (approved):
Works great 👍

